PR #8, opened 2017-04-19
Adds Params handler specs

eliasjpr:
This is necessary to have tests against existing functionality.

### DIFF
--- a/spec/amber/pipe/params_spec.cr
+++ b/spec/amber/pipe/params_spec.cr
@@ -1,0 +1,92 @@
+require "../../spec_helper"
+
+module Amber
+  module Pipe
+    describe Params do
+      it "parses query params" do
+        request = HTTP::Request.new("GET", "/?test=test")
+        context = create_context(request)
+        params = Params.instance
+
+        params.call(context)
+
+        context.params["test"].should eq "test"
+      end
+
+      it "parses multiple query params" do
+        request = HTTP::Request.new("GET", "/?test=test&test2=test2")
+        context = create_context(request)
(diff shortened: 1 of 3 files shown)


spec/amber/pipe/params_spec.cr
```diff
@@ -0,0 +1,92 @@
...
+
+        params.call(context)
+
+        context.params["test"].should eq "test"
+      end
+
+      it "parses body params with charset" do
+        headers = HTTP::Headers.new
+        headers["Content-Type"] = "application/x-www-form-urlencoded; charset=UTF-8"
+        body = "test=test"
```
bew: put some utf-8 chars here?

eliasjpr: Can you give me more context about the utf-8 chars? Do you mean something like test=\x3D?

spec/amber/pipe/params_spec.cr
```diff
@@ -0,0 +1,92 @@
...
+        context = create_context(request)
+        params = Params.instance
+
+        params.call(context)
+
+        context.params["test"].should eq "test"
+      end
+
+      it "parses json array" do
+        headers = HTTP::Headers.new
```
bew: note: would be great to be able to do sth like:
```cr
headers = HTTP::Headers{"Content-Type": "application/json"}
```

spec/amber/pipe/params_spec.cr
bew: Yes! I mean some text that's not ascii (ex: `¿øœ`)

PR #14, opened 2017-04-21
Ep/controller params

eliasjpr:
### Description of the Change
This PR adds the possibility to define required params and validate them.

A malicious user can exploit this scenario by sending this extra parameter to our application, thereby making themselves an administrator.

We can easily fix this problem by filtering the allowed parameters that are permitted inside our application. Please always remember that params represent untrusted input.

* **Controller Actions Params Validation** Parameters are taken from the context and passed as an argument to #call. They are similar to a Ruby Hash, but they offer an expanded set of features.
    * Look into 
        * https://github.com/dry-rb/dry-logic/blob/master/lib/dry/logic/predicates.rb
        * https://github.com/neovintage/accord
    * Whitelisting anf Validations

### Benefits

* We can see validations as the set of rules for data correctness that we want for a specific use case. For us, a User can be persisted with or without a password, depending on the workflow and the route through which the User is persisted.

```crystal
params do
  required(:name).filled(:str?)
  required(:email).filled(:str?, format?: /@/).confirmation
  required(:password).filled(:str?).confirmation
  required(:terms_of_service).filled(:bool?)
  required(:age).filled(:int?, included_in?: 18..99)
  optional(:avatar).filled(size?: 1..(MEGABYTE * 3))
end
```
* The second important aspect is that we use validations to prevent invalid inputs to propagate in our system. In an MVC architecture, the model layer is the farthest from the input. It's expensive to check the data right before we create a record in the database.


spec/amber/support/params_spec.cr
```diff
@@ -0,0 +1,43 @@
...
+
+                    params = Params.new(http_params)
+                    result = params.required("name") { |v| v.str? & !v.empty? & v.email? }
+
+                    result.should eq false
+                end
+            end
+        end
+    end
+end
```
bew: this file indentation is 4 spaces instead of 2, run `crystal tool format` 😃

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
...
+                    range.includes? self
+                end
+
+                def gteq?(num)
+                    !lt?(num)
+                end
+            end
+        end
+    end
+end
```
bew: this file indentation is 4 spaces instead of 2, run `crystal tool format` 😃

spec/amber/support/params_spec.cr
eliasjpr: yes, I will reformat still in progress :) I just wanted to show you guys the Amber::Controller::Base file

src/amber/support/params.cr
```diff
@@ -0,0 +1,42 @@
...
+    end
+end
+
+# Monkey patch the
+class String
+    include Amber::Support::Predicates::String
+end
+
+abstract struct Number
+    include Amber::Support::Predicates::Number
```
bew: mixed indentation in this file (4 spaces here, 2 spaces there)

src/amber/support/params.cr
```diff
@@ -0,0 +1,42 @@
+require "./predicates"
+
+class Check
+  def self.it(match)
+    match.nil? ? false : true
+  end
+end
+
+# Monkey patch the
```
bew: the? 😄

src/amber/support/params.cr
```diff
@@ -0,0 +1,42 @@
...
+
+abstract struct Number
+  include Amber::Support::Predicates::Number
+end
+
+module Amber
+  module Support
+    class Params
+      property params
+      property valid : Bool = false
```
bew: you can just `property valid = false`, the compiler will guess from `false` that `valid` is a boolean

src/amber/support/params.cr
```diff
@@ -0,0 +1,42 @@
...
+
+      def initialize(params : HTTP::Params)
+        @params = params
```
bew: can be shortened with:
```cr
def initialize(@params : HTTP::Params)
end
```

src/amber/support/params.cr
```diff
@@ -0,0 +1,42 @@
...
+      end
+
+      def validate(&block)
+        result = with self yield
+        @valid = result
+      end
+
+      def valid?
```
bew: you can set on top of class:
```cr
property? valid = false
```
The `property?` will generate the `valid?` getter for you.

eliasjpr: Oh nice!!! Didn't knew about this one

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
+module Amber
+  module Support
+    module Predicates
+      module String
+        macro included
+                    def str?
```
bew: you'll have to do the indentation yourself here...

but fist, why put this in a macro hook instead of puting this in the module String directly?

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
...
+                    def time_string?
+                        Check.it self.match(/^(2[0-3]|[01]?[0-9]):([0-5]?[0-9]):([0-5]?[0-9])$/)
+                    end
+                end
+      end
+
+      module Number
+        # the positive number validation
+        def positive?
+          self.sign === 1
```
bew: what about `0` ?

src/amber/controllers/base.cr
```diff
@@ -0,0 +1,18 @@
+require "http"
+
+module Amber
+  module Controllers
```
fridgerator: seems like this should be singular `Controller`

src/amber/controllers/base.cr
```diff
@@ -0,0 +1,18 @@
...
+    class Base
+      property request : HTTP::Request | Nil = nil
```
fridgerator: this is just a matter of semantics / preference, but these could be shortened to `property request : HTTP::Request?`

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
...
+        end
+
+        # the negative number validation
+        def negative?
+          self.sign === -1
+        end
+
+        # the number equals zero validation
+        def zero?
+          self.sign === 0
```
bew: weird check, why not: `self === self.class.zero` ? (this works on any type of numbers)

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
...
+        end
+
+        # the number divisibility (n) validation
+        def div?(n)
+          self.divmod(n)[1] === 0
```
bew: simply `self % n === 0`

src/amber/support/params.cr
```diff
@@ -0,0 +1,37 @@
+require "./predicates"
+
+class Check
+  def self.it(match)
+    match.nil? ? false : true
```
bew: simplify to `!match.nil?` or `!!match`

src/amber/support/params.cr
```diff
@@ -0,0 +1,37 @@
+require "./predicates"
+
+class Check
```
bew: leave this class in top-level ??

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
+module Amber
+  module Support
+    module Predicates
+      module String
+        macro included
+            def str?
```
bew: still 4 spaces in the macro

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
...
+                self.class == String
```
bew: or `self.is_a? ::String`

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
...
+            end
+
+            # hexadecimal validation
+            def hex?
+                Check.it self.match(/^(0x)?[0-9A-F]+$/i)
+            end
+
+            # alpha characters validation
+            def alpha?(locale = "en-US")
+                matches = {
```
bew: This will allocate the hash each time the method is called => slow, you can put this outside the method (in a constant?) https://crystal-lang.org/docs/guides/performance.html#avoid-creating-temporary-objects-over-and-over

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
...
+            end
+
+            # upper case validation
+            def upper?
+                self.upcase === self
+            end
+
+            # affirmative validation
+            def affirmative?
+                Check.it self.match(/^(?:1|t(?:rue)?|y(?:es)?|ok(?:ay)?)$/)
```
bew: note: this works only in english

src/amber/support/params.cr
```diff
@@ -0,0 +1,37 @@
...
+  module Support
+    class Params
+      property params
+      property? valid = false
+
+      def initialize(@params : HTTP::Params); end
+
+      def validate(&block)
+        result = with self yield
+        @valid = result
```
bew: why not `@valid = with self yield` ?

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,236 @@
...
+                "sr-RS@latin" => /^[A-ZČĆŽŠĐ]+$/i,
+                "sr-RS"       => /^[А-ЯЂЈЉЊЋЏ]+$/i,
+                "tr-TR"       => /^[A-ZÇĞİıÖŞÜ]+$/i,
+                "uk-UA"       => /^[А-ЯЄIЇҐ]+$/i,
+                }
+                Check.it self.match(matches[matches.has_key?(locale) ? locale : "en-US"])
+            end
+
+            # numeric characters validation
+            def numeric?
```
bew: this is only Int validation

bew:
instead of merging master on this branch (which creates a merge cmmit) you could rebase this branch on master (`git rebase master`)

src/amber/support/locale_formats.cr
```diff
@@ -0,0 +1,76 @@
...
+        "pl-PL"       => /^[0-9A-ZĄĆĘŚŁŃÓŻŹ]+$/i,
+        "pt-PT"       => /^[0-9A-ZÃÁÀÂÇÉÊÍÕÓÔÚÜ]+$/i,
+        "ru-RU"       => /^[0-9А-ЯЁ]+$/i,
+        "sr-RS@latin" => /^[0-9A-ZČĆŽŠĐ]+$/i,
+        "sr-RS"       => /^[0-9А-ЯЂЈЉЊЋЏ]+$/i,
+        "tr-TR"       => /^[0-9A-ZÇĞİıÖŞÜ]+$/i,
+        "uk-UA"       => /^[0-9А-ЯЄIЇҐ]+$/i,
+      }
+
+      PHONE_FORMAT = matches = {
```
bew: remove `matches =` ?

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,165 @@
...
+      module String
+        macro included
+
+          def str?
+            self.is_a? String
+          end
+
+          # email validation
+          def email?
+            !!self.match(/^[_]*([a-z0-9]+(\.|_*)?)+@([a-z][a-z0-9-]+(\.|-*\.))+[a-z]{2,6}$/)
```
bew: why `!!` here but not for `domain?`, `url?`, etc.. ?

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,174 @@
...
+          end
+
+          # hexadecimal validation
+          def hex?
+            !!self.match(/^(0x)?[0-9A-F]+$/i)
+          end
+
+          # alpha characters validation
+          def alpha?(locale = "en-US")
+            !!self.match(ALPHA[matches.has_key?(locale) ? locale : "en-US"])
```
bew: from where does `matches` come from in `ALPHA[matches.has_key?(locale) ? locale : "en-US"]` ?

idea:
```cr
# somewhere outside:
DEFAULT_LOCALE = "en-US"

def alpha?(locale = DEFAULT_LOCALE)
  regex = ALPHA[locale]? || ALPHA[DEFAULT_LOCALE]
  !!self.match regex
end
```
I promise, this is the last comment 😛

eliasjpr: It's all good. I'm still working on it so. And yes the matches variable was removed so is not needed here

src/amber/controller/param_validator.cr
```diff
@@ -0,0 +1,43 @@
...
+        getter params = {} of String => String
+
+        def initialize(@valid, @params); end
+      end
+
+      macro included
+        property? valid = false
+        property validated_params = {} of String => String
+
+        def validate
```
elorest: Maybe we we should add these methods onto HTTP::Params.

```cr
class HTTP::Params
  def validate
     ....
  end

  private def required(key)  
   ....
  end
```

eliasjpr: @elorest I thought of extending the class instead of monkey patching. Now, what would be the design implications for extending or monkey patching this way? A couple of things that comes to mind is `params.validate {}` I guess this is similar to rails `params.required` the difference being is that we will be able to do the validations something that rails lack, the only framework that's doing this nicely is hanami. 

My personal preference is extending the **HTTP:: Params** class. Any thoughts?

elorest:
That would be my preference as well.

PR #16, opened 2017-04-22
Use full class name in routes

eliasjpr:
Issue: https://github.com/Amber-Crystal/amber/issues/15

* The current route definition `get "/", :hello, :index, :api` is a little
unreadable with 3 symbols in a row, and is not immediately obvious what
each symbol does.

* Updated the routes macros methods to use class names instead of
symbols. Now routes can be defined as follow:

```crystal 
get "/*", HelloController, :world, :static
get "/hello", HelloController, :world, :api
get "/hello/:planet", HelloController, :world, :api
```

* The routes are now more descriptive and easier to read and understand.


elorest:
What is the point of defining api, static, etc in the route? Shouldn't the controller be in charge of what content it's returning? It seems that ultimately it would anyway.

eliasjpr:
There is the concept of pipelines that you as developer can define, a pipeline will allow you to set a list of handlers to perform do the that type of request. For instance for static content you don't need to run certain handlers. See the example https://github.com/Amber-Crystal/amber/blob/master/examples/demo.cr

fridgerator:
Phoenix has the same concept and defined which pipes the routes will use at the routes level as well.  Instead they do it with blocks rather than on each individual route.  http://www.phoenixframework.org/docs/routing

eliasjpr:
It has been a small detail rather hard for me to implement with a block and implicitly inferring the the pipe based on the block definition. I would like to clean that up if possible.  Any idea how we can implement this cleanly

bew:
Wow, I ❤️❤️❤️ the routing capabilities of phoenix, maybe I'll try an implementation of it!

PR #17, opened 2017-04-23
Refactor routing & pipeline dsl

bew:
### Description of the Change

This PR moves the pipeline and routes DSLs inside Amber module, in the revelant classes.
`DSL` stands for Domain Specific Language. Here we have 2 distinct DSLs:
- Pipeline (with `plug`)
- Routing (with `get`, `post`, ...)

Before theses changes, the 2 DSLs where exposed in the top level and where only macros, which could potentially break every program that includes them (see https://github.com/crystal-lang/crystal/issues/236). 

### Benefits

- The DSLs are only available inside specific blocks (`plug` in `pipeline` for example).
- They apply changes to a specific router/pipeline (no singleton involved here!)


spec/amber/pipe/router_spec.cr
```diff
@@ -6,7 +6,7 @@ module Amber
       describe "#call" do
         it "raises exception when route not found" do
           router = Router.instance
-          request = HTTP::Request.new("GET", "/hello/world")
+          request = HTTP::Request.new("GET", "/bad/route")
```
eliasjpr: 👍

src/amber/pipe/pipeline.cr
```diff
@@ -28,7 +28,13 @@ module Amber
       def build(valve : Symbol, &block)
         @valve = valve
         @pipeline[@valve] = [] of HTTP::Handler unless @pipeline.key? @valve
-        with self yield
+        with PipelineDSL.new(self) yield
+      end
+
+      record PipelineDSL, pipeline : Pipeline do
```
eliasjpr: I would like to learn more about Records this is very nice and claen

src/amber/pipe/router.cr
```diff
@@ -61,6 +61,27 @@ module Amber
         trail = build_node(:HEAD, route.resource)
         @routes.add(trail, route)
       end
+
+      record RouterDSL, router : Router do
```
eliasjpr: This is so much cleaner

eliasjpr:
Maybe we should consider moving all the DSLs into **Amber::Support::DSL** module? to keep it organize?

bew:
Why not, what is **Amber::Support** exactly for you?

eliasjpr:
**Amber::Support** should host all of our custom stuff specific to the framework: DSL, Extensions, Modules etc

PR #20, opened 2017-04-24
Amber::Controller:Base

eliasjpr:
## Better convention for controllers

- Moves base controller class to controller folder. This PR comes in order to split the Validations PR into smaller ones.

### Description of the Change

The idea is to have a better convention of where things should go.

### Alternate Designs

- Maybe we should use a module instead - I am opent to ideas

Developers should be able to extend the base controller of the framework to add functionality.

### Benefits

This sets a good convention on where to find the base controller class


examples/demo.cr
```diff
@@ -9,7 +9,7 @@
 # The first line requires the framework library.
 require "../src/amber"
 
-class HelloController < Amber::Controller
+class HelloController < Amber::Controller::Base
```
elorest: Formatting looks wrong.

examples/demo.cr
```diff
@@ -9,19 +7,20 @@
 # The first line requires the framework library.
 require "../src/amber"
 
-class HelloController < Amber::Controller
```
elorest: Fixed some formatting.

PR #21, opened 2017-04-25
Ep/param validations

eliasjpr:
### Requirements

- Allow parameters whitelisting. 
- Allow additional and friendlier validation predicates.

### Description of the Change

- With this PR we introduce Validations for params 

### Alternate Designs

One caveat is that at the moment of defining the validation rules validations are ran for those rules immediately. It would be nice to run the rules every time `validate!` or `validate` is called.

A better design would be to have this more OOP. With **Validator**, **Rules** and **Errors** objects.

- A malicious user can exploit this scenario by sending this extra parameter to our application, thereby making themselves an administrator.
- Validations perform before hitting the model prevents additional processing. 

### Benefits

Validation rules can be defined as following

```crystal 
# setup validation rules here
# Note this runs the validation immediately
params.validation do
  required(:email) { |p| p.url? }
end
```

Returns validated params if valid otherwise raises error
```crystal
user = User.new params.validate!
```

Validations can be checked with 
```crystal
# Usage 2/ input may be invalid
unless params.valid?
  # return an error then quit this method
  response.puts {errors: params.errors}.to_json
  response.status_code 400
  return
end
```


src/amber/validations/params.cr
```diff
@@ -0,0 +1,108 @@
...
+
+    # Setups validation rules to be performed
+    #
+    # ```crystal
+    # params.validation do
+    #   required(:email) { |p| p.url? }
+    #   required(:age, UInt32)
+    # end
+    # ```
+    # returns Validator
```
bew: The return value should not be documented

src/amber/validations/params.cr
```diff
@@ -0,0 +1,108 @@
...
+    #   response.puts {errors: params.errors}.to_json
+    #   response.status_code 400
+    # end
+    # ```
+    # returns Boolean
+    def valid?
+      errors.empty?
+    end
+
+    # Captures and Prforms the validation rules block
```
bew: typo `Prforms`

spec/amber/validations/params_spec.cr
```diff
@@ -0,0 +1,87 @@
...
+      it "validates required field" do
+        http_params = HTTP::Params.parse("name=elias&last_name=perez&middle=j")
+        validator = Validators::Params.new(http_params)
+
+        result = validator.validation do
+          required("name") { |v| v.str? & !v.empty? }
+          required("last_name") { |v| v.str? & !v.empty? }
+        end
+
+        result.valid?.should eq true
```
bew: `result.valid?.should be_true`

src/amber/validations/params.cr
```diff
@@ -0,0 +1,104 @@
...
+
+    # Validates each field with a given set of predicates returns true if the
+    # field is valid otherwise returns false
+    #
+    # ```crystal
+    # required(:email) { |p| p.email? & p.size.between? 1..10 }
+    # ```
+    def required(key, msg : String? = nil)
+      unless raw_params.has_key?(key)
+        errors[key] = {"invalid", "Param [nonexisting] does not exist."}
```
bew: What is `"invalid"`?

bew: Also, the following error could be rewritten as `"Param [#{key}] is required."`

src/amber/support/predicates.cr
```diff
@@ -0,0 +1,174 @@
+module Amber
+  module Support
+    module Predicates
+      module String
+        macro included
```
bew: why do you need this macro?

eliasjpr: It should be a module my bad

eliasjpr: It should be a module my bad

eliasjpr: Actually no, the intention of the macro included is to extend to monkey path the string class with new predicte methods

src/amber/validations/params.cr
```diff
@@ -0,0 +1,104 @@
...
+    #   response.status_code 400
+    # end
+    # ```
+    def valid?
+      errors.empty?
+    end
+
+    # Captures and performs the validation rules block
+    #
+    # returns Nil
```
bew: Still a useless return value documentated

src/amber/validations/params.cr
```diff
@@ -0,0 +1,104 @@
...
+
+    # Validates each field with a given set of predicates returns true if the
+    # field is valid otherwise returns false
+    #
+    # ```crystal
+    # required(:email) { |p| p.email? & p.size.between? 1..10 }
+    # ```
+    def required(key, msg : String? = nil)
+      unless raw_params.has_key?(key)
+        errors[key] = {"", "Param [#{key}] does not exist."}
```
bew: You removed `invalid` but what is this `""` value? what does it represent?

src/amber/support/predicates.cr
eliasjpr: Macro included not needed here. Will be removed

src/amber/validations/params.cr
eliasjpr: this holds the value for the `param[key]` but in this case there is no key so there is no value

bew: You should put `nil` do better differenciate things?

eliasjpr: Will do

src/amber/validations/params.cr
```diff
@@ -0,0 +1,103 @@
+require "../support/predicates"
+
+# We are patching the String class and Number struct to extend the predicates
+# available this will allow to add friendlier methods for validation cases.
+class String
```
fridgerator: `support/predicates.cr` already patches these, does it need to be duplicated?

eliasjpr: Good catch @fridgerator. I did some cleanup on this.

eliasjpr:
Please keep in mind that I will circle back to this implementation to make it more OOP as @bew suggested. Right now this is a working proof of concept as we figure out design matters.

bew:
Good job! :smile:

PR #25, opened 2017-04-26
added render, render_template, render_both

elorest:
### Description of the Change

`render`, `render_both`, `render_template` helper methods to render slang or ecr templates.
* `render template.slang` renders a file in views folder named after controller, with default application layout.
* `render_both("path/to/template.ecr", "path/to/layout.slang")`: renders path and layout with explicit paths.
* `render_template("path/to/template.slang")`: renders single template without layouts.

It's useful to have the ability to render html templates in web frameworks.

### Benefits

We can render templates in a web framework! :)


.gitignore
```diff
@@ -4,7 +4,10 @@
 /.shards/
 /.vscode/
 /.github/
+.DS_Store
+lib/
+.shards/
 
 # Libraries don't need dependency lock
 # Dependencies will be locked in application that uses them
-/shard.lock
```
fridgerator: add tmp.cr too? (file listed below)

elorest:
Done

.gitignore
elorest: Done.

examples/shard.lock
```diff
@@ -0,0 +1,18 @@
+version: 1.0
```
bew: you should put this file in gitignore

src/amber/support/dsls.cr
```diff
@@ -23,4 +23,24 @@ module Amber::Support::DSL
 
     {% end %}
   end
+
+  module ControllerActions
+    macro render_both(filename, layout)
```
eliasjpr: I think we can overload the render method.

src/amber/support/dsls.cr
```diff
@@ -23,4 +23,24 @@ module Amber::Support::DSL
...
+      content = render_template("{{filename.id}}")
+      render_template("{{layout.id}}")
+    end
+
+    # helper to render a template.  The view name is relative to `src/views` directory.
+    macro render_template(filename, *args)
```
eliasjpr: A nice comment explaining the usage of this method would be helpful since it receives an x amount of arguments.

src/amber/support/dsls.cr
```diff
@@ -23,4 +23,24 @@ module Amber::Support::DSL
...
+      {% if filename.id.split("/").size > 2 %}
+        Kilt.render("{{filename.id}}", {{*args}})
+      {% else %}
+        Kilt.render("src/views/{{filename.id}}", {{*args}})
+      {% end %}
+    end
+
+    macro render(filename, layout = application.slang, folder = __FILE__)
```
bew: use `__DIR__` instead of `__FILE__`

src/amber/support/dsls.cr
bew: What is this macro for?

src/amber/support/dsls.cr
```diff
@@ -23,4 +23,24 @@ module Amber::Support::DSL
 
     {% end %}
   end
+
+  module ControllerActions
+    macro render_both(filename, layout)
+      content = render_template("{{filename.id}}")
+      render_template("{{layout.id}}")
```
bew: use `render_template({{layout}})` to allow passing a variable

src/amber/support/dsls.cr
elorest: Perhaps. Because of the need to pass in the current file on render it's a little bit hard to get the exact same functionality. I don't feel that render_both will be needed very much.

src/amber/support/dsls.cr
elorest: `__DIR__ would return the controllers directory. I need the name of the file.

tmp.cr
```diff
@@ -0,0 +1,9 @@
+macro get_file_name(file = __FILE__)
```
bew: what is this fiiiile??!!

src/amber/support/dsls.cr
elorest: To explicitly render templates and layouts in any location. I also call it from render to keep the functionality dry.

src/amber/support/dsls.cr
elorest: Without quotes you mean?

bew: yes, and without `.id`

src/amber/support/dsls.cr
```diff
@@ -23,4 +23,24 @@ module Amber::Support::DSL
...
+    end
+
+    # helper to render a template.  The view name is relative to `src/views` directory.
+    macro render_template(filename, *args)
+      {% if filename.id.split("/").size > 2 %}
```
bew: `{% if filename.includes?('/') %}`

src/amber/support/dsls.cr
bew: but what is this `content` variable?
why render the file, then the layout? (it seems that I'm missing sth here..)

eliasjpr:
@elorest rebased master on top of this branch

README.md
```diff
@@ -107,6 +126,7 @@ MyAwesomeApp.config do
     get "/", :hello, :world, :api
```
fridgerator: do all of these `:hello` need to be changed to `HelloController` ?

elorest:
Probably but I didn't want to step on toes so I just changed mine.

spec/amber/pipe/csrf_spec.cr
```diff
@@ -0,0 +1,92 @@
+require "../../../spec_helper"
```
eliasjpr: This file should be removed.

spec/support/render.cr
```diff
@@ -0,0 +1,31 @@
+require "../spec_helper"
```
eliasjpr: This should be in `spec/controller` instead of `support/render`

eliasjpr:
LGTM. Feel free to merge

jeromegn:
If I may offer a comment here: It may not be advisable to add Slang as a dependency. Kilt was built so that template engines could be required, as needed, by users based on their preferences.

By default, Kilt supports ECR. Unless Amber is opinionated about the templating language to use, then I'd suggest not adding this dependency.

I'm not involved in this project, so feel free to dismiss my opinion!

spec/amber/controller/render.cr
```diff
@@ -0,0 +1,31 @@
+require "../../spec_helper"
+
+class TestController < Amber::Controller::Base
```
eliasjpr: Can you move this controller to the factories.cr file. It should be just a copy and paste.

src/amber/dsl/router.cr
```diff
@@ -23,4 +23,24 @@ module Amber::DSL
 
     {% end %}
   end
+
+  module ControllerActions
```
eliasjpr: I'm cool with this all in all. I would really like for us to implement class views eg. For a given `HomeController` you will have a corresponding `HelloView` with this in place we can implement the decorator pattern which allows for greater separation of concerns. An example directory structure would be something like:

Controller
-->HomeController.cr
View
-->templates
-->layouts
-->HomeView.cr

src/amber/router/pipe/csrf.cr
```diff
@@ -19,7 +19,7 @@ module Amber
         @check_methods = CHECK_METHODS
       end
 
-      def call(context : HTTP::Server::Context)
```
eliasjpr: Is there any particular reason why this was removed?

elorest:
How are there like 20 merge conflicts again?

elorest:
I no longer have the option to reply to comments, in this PR which is making communication really hard. I removed that the explicit type because I had added it in my lazy loading params branch but it didn't solve anything. Not sure how it got in this branch, unless maybe you added it at the same time. I just got back from Railsconf and have a lot to catch up on at work. By the time I have time to  change those things and fix the merge conflicts it will probably be far too divergent from master again. I'm fine with view objects in some cases but don't like the idea of moving files from `views/controller_name/actionname.(ecr|slang)` to templates.

PR #28, opened 2017-04-27
Controller Redirection method

eliasjpr:
### Requirements

* Adds controller redirection, it enables developers to redirect from within an action to another location. 
* Adds a route.call method, this serves in preparation for before and after controllers filters

```crystal
# Supports redirection in the following signatures
# reditect_to location, status
redirect_to "/new"
redirect_to "/home/index"
redirect_to :back
redirect_to :show
redirect_to "http://www.ambercr.io"
```

### Description of the Change

Creates a Redirect modules that is included in the base controller. The module has a `redirect_to` method that accepts a location and a status code.

Makes context, request, and response private variables since developers should not be able to modify these variables directly.

Redirections within an action are necessary at times to change path based on a given condition.

### Benefits

Developers can perform redirections within actions

### DIFF
--- a/src/amber/controller/base.cr
+++ b/src/amber/controller/base.cr
@@ -1,18 +1,22 @@
 require "http"
+require "./*"
 
 module Amber::Controller
   class Base
-    include Amber::DSL::ControllerActions
+    include DSL::ControllerActions
+    include Redirect
 
-    property request : HTTP::Request?
-    property response : HTTP::Server::Response?
-    property raw_params : HTTP::Params?
-    property context : HTTP::Server::Context?
+    protected getter request = HTTP::Request.new("GET", "/")
+    protected getter response = HTTP::Server::Response.new(IO::Memory.new)
+    protected getter raw_params = HTTP::Params.parse("")
+    protected getter context : HTTP::Server::Context?
+    protected getter params : Amber::Validators::Params?
 
-    protected def set_context(@context : HTTP::Server::Context)
-      self.request = context.request
-      self.response = context.response
-      self.raw_params = context.params
+    def set_context(@context : HTTP::Server::Context)
+      @request = context.request
+      @response = context.response
+      @raw_params = context.params
+      @params = Amber::Validators::Params.new(@raw_params)
     end
   end
 end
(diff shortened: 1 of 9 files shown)


src/amber/controller/base.cr
```diff
@@ -1,16 +1,23 @@
...
-      self.response = context.response
-      self.raw_params = context.params
+    getter _context : HTTP::Server::Context?
+
+    getter _request : HTTP::Request = HTTP::Request.new("GET", "/")
+    getter _response : HTTP::Server::Response =  HTTP::Server::Response.new(IO::Memory.new)
+    getter _params : HTTP::Params = HTTP::Params.parse("")
+    getter params : Amber::Validators::Params?
+
+    def _set_context(@_context : HTTP::Server::Context)
```
elorest: Would it be possible to set all of this with `def initialize` do keep it more standard?

eliasjpr: I was thinking about this actually and there is some benefits to the current implementation, one is being that there is only one controller instance loaded for the entire app. We could change it to inititialize a new controller for every request, but will have to redactor quite a bit

eliasjpr:
@fridgerator ideally to handle the parameters would be done through an URL helper method

fridgerator:
@eliasjpr sounds good

src/amber/controller/base.cr
fridgerator: which would be the most performant?

eliasjpr: The only way to answer that question would be to test it. But from what I see, currently, there is no object initialization happening on a request just execution of the handler and passing the context to the controller. 

The other approach would be to initialize the requested controller right before handling the action.

elorest: Hmmm. Wouldn't it have to be another instance if we ever wanted it to be multithreaded? Mainly I just don't like that were initializing all of the instance variables to nil with a union to what they should be.

eliasjpr:
This is ready for another look. I ran the format tool and some other files got pulled into this PR.

eliasjpr:
:)

PR #30, opened 2017-05-01
Controller Before/After action filters

eliasjpr:
## Controller Before/After action filters

Issue: [Controller Filters [ch16]](https://app.clubhouse.io/amber/story/16/controller-filters)

### Description of the Change

* Allows developers to add before and after action controller filters.

### Benefits

Before/After filters are a great way to execute extra logic during the
request lifecycle. This allows to add extra configuration at the
developer discretion without cluttering the controller and maintaining
controllers DRY.

### Changes

* Adds DSL module that will add 2 macros before_action, aftert_action
* Adds a filter record
* Adds Callbacks class this class in responsible for running and
registering before/after filters.
* Adds FilterBuilder to allow block-like configuration for filters.

### Requirements

* All new code requires tests to ensure against regression

### Design

```crystal 
before_action do
    only :index { say_hello }
    only [:index, show] {  set_admin_layout }
end
```


eliasjpr:
This is ready for review

src/amber/controller/filters.cr
```diff
@@ -0,0 +1,83 @@
...
+      filters.add Filter.new(precedence, action, block)
+    end
+  end
+
+  class Filters
+    include Enumerable({Symbol, Array(Filter)})
+
+    property filters = {} of Symbol => Array(Filter)
+
+    def register(precedence : Symbol)
```
fridgerator: would specifying nil return type for this method allow us to remove the explicit `nil` return when calling this method [below](https://github.com/Amber-Crystal/amber/pull/30/files#diff-38ccdec39f6cb0c091906082b1f27cddR7), i.e. `def register(precendence: Symbol) : Nil`

fridgerator:
Overall, looks good to me

src/amber/dsl/callbacks.cr
```diff
@@ -0,0 +1,21 @@
+module Amber::DSL
+  module Callbacks
+    macro before_action
+      def before_filters
```
bew: Why is there a def in the macro?

eliasjpr: because I want to trap the body of the macro without running to execute it at a later time, if the `def` would be there the code will execute in the wrong context, in the current case at the class level and not the instance level.

bew: nice!

PR #36, opened 2017-05-04
Routing Resources

eliasjpr:
## Route Restful Resources

Issue:  [Restful Resources [ch43]](https://app.clubhouse.io/amber/story/43/routes-resources)

### Requirements

As a developer, I want to generate restful resources for a given controller

```
GET     /users          UserController  :index
GET     /users/:id/edit UserController  :edit
GET     /users/new      UserController  :new
GET     /users/:id      UserController  :show
POST    /users          UserController  :create
PATCH   /users/:id      UserController  :update
PUT     /users/:id      UserController  :update
DELETE  /users/:id      UserController  :delete
```

### Description of the Change

The router supports other macros besides those for HTTP verbs like **get, post, and put**. The most important among them is resources, which expands out to eight clauses of the match/3 function.

### Alternate Designs
```crystal
resources "/user", UserController, [:index, :show]
resources "/user", UserController, actions: [:index, :show]
```


examples/src/demo.cr
```diff
@@ -35,7 +35,7 @@ MY_APP_SERVER.config do |app|
 
   # All static content will run these transformations
   pipeline :static do
-    plug HTTP::StaticFileHandler.new "../examples/public", true
+    plug HTTP::StaticFileHandler.new "../examples/public", false
```
bew: what is `false` here?

eliasjpr: `false` stops from processing the next handler, `true` allows for the processing of the handler.

Doesn't really matter this is is just for demonstration.

src/amber/dsl/router.cr
```diff
@@ -15,12 +20,24 @@ module Amber::DSL
...
-    {% for verb in {:get, :post, :put, :delete, :options, :head, :trace, :connect} %}
+    {% for verb in RESOURCES %}
 
       macro {{verb.id}}(*args)
         route {{verb}}, \{{*args}}
       end
 
     {% end %}
+
+    # TODO Clean this up
```
eliasjpr: Would like to know if there is a better implementation for this. My macro skills are limited. I tried approaching it with a nice hash and do the lookup base on the items in the actions variable.

examples/src/demo.cr
bew: Ok!

examples/src/demo.cr
```diff
@@ -45,14 +45,30 @@ MY_APP_SERVER.config do |app|
   # (HTTP METHODS)[https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html]
   routes :static do
     # Each route is defined as follow
-    # verb, resources : String, controller : Symbol, action : Symbol,
-    # pipeline : Symbol
-    get "/*", StaticController, :index
+    # verb resource : String, controller : Symbol, action : Symbol
+    get "/index.html", StaticController, :index
```
bew: Idea-for-later: `get_fallback StaticController`

eliasjpr: What's the intention? 

Also, we should find a way to define a static method/macro that does not need a controller/action to be defined with it just a path.

src/amber/dsl/router.cr
bew: The only thing I've managed to do is: https://carc.in/#/r/1zb8

examples/src/demo.cr
```diff
@@ -45,14 +45,30 @@ MY_APP_SERVER.config do |app|
...
   end
 
+  # Routes accepts a pipeline name and a scope a pipeline represents a stach of
+  # http handlers that will process the current request
   routes :web, "/v2" do
-    get "/hello", HelloController, :index
-    get "/hello/:planet", HelloController, :world
+    # You can also define all resources at once with the resources macro.
+    # This will define the following routes
+    # resoources path, controller, actions
```
fridgerator: just commenting on spelling stuff:
* "resoources" -> "resources"
* should be `resources "/users"` I think (we're not doing any pluralizing from what I can tell) ?

eliasjpr: Short answer is no. I think that should be up to the developer. Implicitly pluralizing I am not a fan of. What is your though on that?

fridgerator: I 100% agree, I'm not  fan either.  Was just clarifying.

PR #38, opened 2017-05-06
initial websockets implementation

fridgerator:
### Requirements

Initial implementation of a websocket handler, manager and server.  

### Description of the Change

This PR allows for a websocket handler to be created within the application config block.  A user can also define their own `Socket` and `Channel` functionality (authentication, message handling, etc).  Example code below.

* Allows for websocket communication to be easily added to any application.

### Features still TODO

* Heartbeat - server: to kill dead connections, client: to ensure connection to server is still active
* PubSub implementation through redis / postgres - for application scalability
* Allow for different decoding methods - allow the user to use something other that JSON if they want.
* Message buffering?
* Long polling fallback?

### Example Code

1. Create a socket and channel:
```crystal
struct UserSocket < Amber::WebSockets::ClientSocket
  channel "chat_room:*", ChatChannel

  def on_connect
    # This runs when the socket connects.  Authentication should happen here.
    # If `on_connect` returns false, the socket will be closed.
    true
  end
end

class ChatChannel < Amber::WebSockets::Channel
  def handle_joined
    # Called when a user joins a channel
  end

  def handle_message(msg)
    # Called when a message is sent to the channel
    # use rebroadcast! to broadcast to the message to all users subscribed to the channel
    rebroadcast!(msg)
  end
end
```

2. In the application config block, define the handler endpoint
```crystal
MyApp.config do
  ...
  socket_endpoint "/chat", UserSocket
 ...
end
```

3. Javascript library usage:
```javascript
let socket = new Amber.Socket("/chat");
socket.connect();

let channel = socket.channel('chat_room:123')
channel.join()

channel.on('msg:new', (msg) => { console.log('new message : ', msg) } )

channel.push('msg:new', {message: 'hey guys'})
```


src/amber/router/pipe/router.cr
```diff
@@ -11,6 +11,7 @@ module Amber
 
       def initialize
         @routes = Radix::Tree(Route).new
+        @socket_routes = Array(NamedTuple(path: String, handler: WebSockets::Server::Handler)).new
```
fridgerator: Should this also be a radix tree?  I didn't use one because I don't see there being many socket_endpoints in an application.

eliasjpr:
@fridgerator still reviewing... Looking good so far

fridgerator:
Sorry, are we squashing commits?  Should I squash all of these and re-open?

fridgerator:
closing to clean up commit history

PR #40, opened 2017-05-07
fixed routing helper resources

elorest:
### Description of the Change
If I create resourceful routes like:
```cr
resources "posts", PostController, [:index, :show]
```

It errors on compile because all the other actions don't exist in my controller. This happens because the if in the following code happens at run time but the macro that creates the route happens at compile time. 

```cr
    macro resources(path, controller, actions = [:index, :edit, :new, :show, :create, :update, :put, :delete])
      get "{{path.id}}", {{controller}}, :index if {{actions}}.includes?(:index)
      get "{{path.id}}/:id/edit", {{controller}}, :edit  if {{actions}}.includes?(:edit)
      get "{{path.id}}/new", {{controller}}, :new if {{actions}}.includes?(:new)
      get "{{path.id}}/:id", {{controller}}, :show if {{actions}}.includes?(:show)
      post "{{path.id}}", {{controller}}, :create if {{actions}}.includes?(:create)
      patch "{{path.id}}/:id", {{controller}}, :update if {{actions}}.includes?(:update)
      put "{{path.id}}/:id", {{controller}}, :update if {{actions}}.includes?(:update)
      delete "{{path.id}}/:id", {{controller}}, :delete if {{actions}}.includes?(:delete)
    end
```

The fix is to check if actions include the right thing at compile time in the macro before creating the route. Like this:
```cr
{% if actions.includes?(:index) %}
get "{{path.id}}", {{controller}}, :index
{% end %}
```

Because route helpers are useful.

### Benefits

It works now for more cases.

### Applicable Issues

https://github.com/Amber-Crystal/amber/issues/39

### DIFF
--- a/src/amber/dsl/router.cr
+++ b/src/amber/dsl/router.cr
@@ -21,7 +21,6 @@ module Amber::DSL
     end
 
     {% for verb in RESOURCES %}
-
       macro {{verb.id}}(*args)
         route {{verb}}, \{{*args}}
       end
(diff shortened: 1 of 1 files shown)


bew:
@elorest can you indent like:
```cr
{% if actions.includes?(:index) %}
  get "{{path.id}}", {{controller}}, :index
{% end %}
```
please?

elorest:
@bew I didn't indent that on purpose because it makes the code the Marco creates indented too far. In this case maybe it doesn't matter. But in a lot of cases where you're  creating methods or other blocks inside of a macro it looks really ugly when you indent wrong and are trying to debug.

bew:
Hmm ok, let's go then!

PR #41, opened 2017-05-07
Compiling Error While including Amber::DSL::Callbacks

eliasjpr:
## Compiling Error While including Amber::DSL::Callbacks

Issue: https://app.clubhouse.io/amber/story/44/bug-filters-included-callback-dsl

Error while compiling no references to DSL::Callbacks

* Add require "../dsl/*" to the filter file

No more compiling errors.

### Requirements

Amber was not compiling due to error in Controller::Filters

### Description of the Change

This adds a missing require to the top of the Controller::Filter file

### Benefits 

No longer receive the compiling error message.


src/amber.cr
```diff
@@ -5,8 +5,7 @@ require "colorize"
 require "secure_random"
 require "kilt"
 require "kilt/slang"
-require "./amber/dsl/*"
-require "./amber/support/*"
+require "./amber/*"
```
bew: You don't need this line, there is already `require "./amber/**"` on the next line

PR #44, opened 2017-05-08
Changed controller call to a proc so it can be initialized with real values.

elorest:
### Description of the Change
Changed route controller call to a proc so it can be initialized with real values.

### Alternate Designs

I considered passing the controller constant into the into `Amber::Route` and saving it as an instance variable. Then `Amber::Route.call(context)` would create a new controller instance with context and call the action. I think it's more consistent to create pass the handler in as a proc and just call it from `Amber::Route.call`

It cleans up some of the existing functionality simplifies garbage collection. 

### Benefits

It allows context to exist in controller and views without being a union with nil, which requires using `.not_nil!` everywhere.


eliasjpr:
@elorest I think you should rebase of this branch https://github.com/Amber-Crystal/amber/pull/43 since it has updates to the routes.

eliasjpr:
This is a nice improvement!

PR #45, opened 2017-05-09
added verbs for put, patch and delete

elorest:
### Description of the Change

Adds verbs for put, patch and delete. Fairly small change that doesn't seem to break anything. Don't merge it yet but I wanted to post it so that you didn't waste time in the same direction tomorrow.

Routing all restful routes is useful for web frameworks.

### Benefits
All verbs now work. Params are parsed before the route is looked up. Also the looked up node is saved and used since it was always added to the chain by default. This means we don't have to look up the same route twice per request.

### Possible Drawbacks
Params have to always be parsed in order to route.

### Applicable Issues
If you include params in your pipeline now it will cause an issue because it's being used by default.


eliasjpr:
I am ok with merging this now as long we agree to revisit this implementation

PR #48, opened 2017-05-13
Refactor: Cleans up routing

eliasjpr:
## Request Lifecycle 

* This refactor cleans the request lifecycle and it removes the necessity of a router handler.

### Requirements

* Params are now inline withing the context and always processed
* Pipeline execution is now more straight forward
* It moves files, flash, params and session into modules

### Description of the Change

* Refactor the request lifecycle
* Removes unnecessary router handler
* This maintains backwards compatibility with `_method`


elorest:
@eliasjpr I like everything I've seen up this. I'm on my phone right now though on my way back from crystal code camp. I'll finished reviewing this in the morning.

src/amber/router/context.cr
```diff
@@ -13,8 +13,9 @@ class HTTP::Server::Context
 
   def initialize(@request : HTTP::Request, @response : HTTP::Server::Response)
     router = Amber::Router::Router.instance
-    @route = router.match_by_request(@request)
     parse_params
     upgrade_request_method!
```
eliasjpr: The order matters here since we need parse the params first in order to check if the request method can be upgraded based on the `_method`. Then it tries to find the route for the request.

PR #49, opened 2017-05-16
Controllers actions are expected to return a String not an Int32.

elorest:
### Description of the Change

The `redirect_to` method is currently returning an Int32 of 302 instead of a string. This breaks actions that use a redirect.

### Alternate Designs

An action could be allowed to return a Union(String | Int32) but there is not benefit to returning an Int32.

It's a bug fix for something already in the core.


eliasjpr:
This is fine. I have a PR that will land either tonight or tomorrow morning that fixes redirects and adds more options to defining redirect_to such as redirect_to :controller, :action, :query_params

PR #50, opened 2017-05-19
Better Redirect module for controller

eliasjpr:
### Requirements

This PR makes better use of redirect_to controller method by allowing to define fallback for redirect_back, define controller and action. 

### Description of the Change

When using redirect_back(fallback), if there is no referrer, a redirect error will be raised. You may specify some fallback target define below.

Redirects the browser to the target specified

**Targets:**

* Location this can be a path string, domain name, link.
* Controller and Action this is the target controller and the target action within the controller
* Action the target action of the current controller.

### DIFF
--- a/src/amber/controller/redirect.cr
+++ b/src/amber/controller/redirect.cr
@@ -1,40 +1,58 @@
 module Amber::Controller
-  module Redirect
-    # Supports redirection in the following signatures
-    #
-    # reditect_to location, status
-    # redirect_to "/new"
-    # redirect_to "/home/index"
-    # redirect_to :back
-    # redirect_to :show
-    # redirect_to "http://www.ambercr.io"
-    def redirect_to(location : String | Symbol, status = 302)
-      response.headers.add "Location", parse_redirect(location)
-      response.status_code = status
-      "redirecting..."
+  # This class writes the redirect URL to the response headers and parses params.
+  class LocationRedirect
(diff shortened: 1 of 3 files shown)


src/amber/controller/redirect.cr
```diff
@@ -1,40 +1,58 @@
...
+
+    def initialize(@location : String,
+                   @status = 302,
+                   @query_params : Hash(String, String)? = nil)
+      # Todo add flash as a parameter for redirect
+      # Todo add default route scope for given controller action
+      raise_redirect_error(location) if location.empty?
+    end
+
+    def redirect(for response)
```
elorest: What is `(for response)`

src/amber/controller/redirect.cr
```diff
@@ -1,40 +1,58 @@
...
+      raise_redirect_error(location)
+    end
+
+    private def set_location(response, location, status = 302, query_params : _ = nil)
+      url_path = encode_query_string(location, query_params)
+      response.headers.add "Location", url_path
       response.status_code = status
-      "redirecting..."
+
+      "Redirecting to #{url_path}"
```
elorest: I like this.

src/amber/controller/redirect.cr
eliasjpr: is a crystal feature for you can define aliases to method arguments

PR #75, opened 2017-05-19
Removes id from new resources

eliasjpr:
Resources new should not make use of the :id in the
path.

Remove :id from resource and updated specs

After this changes the new resource should not required the :id in the
url path.


src/amber/dsl/router.cr
```diff
@@ -50,7 +50,7 @@ module Amber::DSL
       {% elsif action == :show %}
         get "{{path.id}}/:id", {{controller}}, :show
       {% elsif action == :new %}
-        get "{{path.id}}/:id/new", {{controller}}, :new
```
elorest: Awesome thanks!

PR #86, opened 2017-05-29
Added csrf tag helper and framework for other helpers.

elorest:
### Description of the Change

Added a `helpers` folder at `src/amber/controller/helpers/`. Added `tag.cr` which currently only contains the `csrf_tag` helper.

Csrf pipe is part of the core and this makes it easier for the end user to use it.

### DIFF
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -33,6 +33,26 @@ module Amber::Controller
 
         TestController.new(context).render_with_layout.should eq html_output
       end
+
+      it "renders a form with a csrf tag" do
+        request = HTTP::Request.new("GET", "/?test=test")
+        context = create_context(request)
+        html_output = <<-HTML
+        <form action="/posts" method="post">
+          <input type="hidden" name="_csrf" value="#{Amber::Pipe::CSRF.instance.token(context)}" />
+          <div class="form-group">
+            <input class="form-control" type="text" name="title" placeholder="Title" value="hey you">
+          </div>
+          <div class="form-group">
+            <textarea class="form-control" rows="10" name="content" placeholder="Content">out there in the cold</textarea>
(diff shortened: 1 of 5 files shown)


spec/amber/controller/base_spec.cr
```diff
@@ -33,6 +33,26 @@ module Amber::Controller
 
         TestController.new(context).render_with_layout.should eq html_output
       end
+
+      it "renders a form with a csrf tag" do
+        request = HTTP::Request.new("GET", "/?test=test")
+        context = create_context(request)
+        html_output = <<-HTML
```
eliasjpr: you can move this into a seed file and read the file contents to make the spec more cleaner.

src/amber/controller/base.cr
```diff
@@ -1,11 +1,12 @@
 require "http"
-require "./*"
+require "./**"
```
eliasjpr: 👍

src/amber/controller/helpers/tags.cr
```diff
@@ -0,0 +1,7 @@
+module Amber::Controller::Helpers
```
eliasjpr: 👍

eliasjpr: I think the helper::tags should be in extensions or support since its adding functionality

spec/amber/controller/base_spec.cr
elorest: There are quite a few of these so a refactor of that should probably be a different PR.

src/amber/controller/helpers/tags.cr
elorest: Perhaps, although it is directly related to controllers and about as useful as redirect or render. I think I'll merge it for now and when we start adding form helpers and stuff we can revisit this issue.

PR #92, opened 2017-06-02
Don't start a new process if thread_count is one.

elorest:
### Description of the Change

Stop fun from creating a new process when thread count is 1. This saves a tiny bit of time and makes processes easier to handle so that `amber watch` still works.

### Benefits

Amber watch works. Doesn't spawn unnecessary thread.


elorest:
Yes but that shouldn't be an issue for development. 

Watch just includes spawns and process and then kills it in a similar way to what our multithread mode is doing. For some reason when it kills the parent process the processes that we created for multithread aren't going down.

PR #93, opened 2017-06-02
Removed the requirement for additional handlers

elorest:
### Requirements

Removed the requirement for additional handlers since we already have params parsing. A hello world app or even api doesn't require anything else. Amber has been like this for the last few weeks but the pipeline rebuild added it back last night.


eliasjpr:
Does this mean that I don't need to define a pipeline at all, since there is no need to add handlers?

PR #96, opened 2017-06-03
Fixed sessions and csrf. Also handled exceptions and passed messages through.

elorest:
### Description of the Change

Fixed sessions. CSRF now works and shows a 403 instead of 500 error.
Handled exceptions thrown in the pipeline with appropriate messages and codes.

Exception Handling is now awesome!

### DIFF
--- a/src/amber.cr
+++ b/src/amber.cr
@@ -11,11 +11,16 @@ module Amber
   class Server
     property port : Int32
     property name : String
+    setter   project_name : String?
     property env : String
     property log : Logger
     property secret : String
     property host : String = "0.0.0.0"
     property port_reuse : Bool = false
+
+    def project_name
+      @project_name ||= @name.gsub(/\W/, "_").downcase
+    end
 
     def self.instance
       @@instance ||= new
@@ -27,7 +32,7 @@ module Amber
 
     def initialize
       @app_path = __FILE__
-      @name = "My Awesome App"
+      @name = "amber_project"
       @port = 8080
       @env = "development".colorize(:yellow).to_s
       @log = ::Logger.new(STDOUT)
--- a/src/amber/exceptions/exceptions.cr
+++ b/src/amber/exceptions/exceptions.cr
@@ -1,25 +1,38 @@
 module Amber
   module Exceptions
-    class DuplicateRouteError < Exception
+    class Base < Exception
+      getter status_code : Int32 = 500
+ 
+      def set_response(response)
+        response.headers["Content-Type"] = "text/plain"
+        response.print message
+        puts status_code
+        response.status_code = status_code
+      end
+    end
+
+    class DuplicateRouteError < Base
       def initialize(route : Route)
         super("Route: #{route.verb} #{route.resource} is duplicated.")
       end
     end
 
-    class RouteNotFound < Exception
+    class RouteNotFound < Base
       def initialize(request)
+        @status_code = 404
         super("The request was not found. #{request.method} - #{request.path}")
       end
     end
 
-    class Forbidden < Exception
+    class Forbidden < Base
       def initialize(message : String?)
+        @status_code = 403
         super(message || "Action is Forbidden.")
       end
     end
 
     module Controller
-      class Redirect < Exception
+      class Redirect < Base
         def initialize(location)
           super("Cannot redirect to this location: #{location}")
         end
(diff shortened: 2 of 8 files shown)


src/amber/router/pipe/pipeline.cr
```diff
@@ -23,7 +23,20 @@ module Amber
         if context.websocket?
           context.process_websocket_request
         else
-          @drain[context.valve].call(context) if @drain[context.valve]
+          begin
+            @drain[context.valve].call(context) if @drain[context.valve]
+          rescue e
```
eliasjpr: We can do a catch all here and initialize an exception from the Amber exceptions. We can move the case statement to an instance method in Amber::Exceptions and do a catch all there. That will clean this handler and there is no need to create and exception handler. We already have a handler that renders 404 and 500 errors. That can be called after the catch all

src/amber/router/pipe/session.cr
```diff
@@ -16,7 +16,7 @@ module Amber::Pipe
     end
 
     def initialize
-      @key = "#{Server.settings.name.gsub(" ", "_")}.session"
+      @key = "#{Server.settings.name.gsub(" ", "_").downcase}.session"
```
eliasjpr: Lets pass the app name on initialize

src/amber/exceptions/exceptions.cr
```diff
@@ -1,39 +1,52 @@
 module Amber
   module Exceptions
-    class DuplicateRouteError < Exception
+    class Base < Exception
```
eliasjpr: Love this!

src/amber/exceptions/exceptions.cr
```diff
@@ -1,39 +1,52 @@
...
+      getter status_code : Int32 = 500
+ 
+      def set_response(response)
+        response.headers["Content-Type"] = "text/plain"
+        response.print message
+        puts status_code
```
eliasjpr: Remove puts or use the app logger if the purpose is to show it on the console.

src/amber/router/pipe/csrf.cr
```diff
@@ -16,20 +15,23 @@ module Amber
...
           call_next(context)
         else
           raise Amber::Exceptions::Forbidden.new("CSRF check failed.")
         end
       end
 
       def valid_token?(context)
-        context.params.fetch(param_key, nil) == token(context) ||
-          context.request.headers.fetch(header_key, nil) == token(context)
+        if context.params[param_key]? == token(context) || context.request.headers[header_key]? == token(context)
```
eliasjpr: No need for the if conditional statement the expression returns a bool

```crystal
context.params[param_key]? == token(context) || context.request.headers[header_key]? == token(context)
 -          context.request.headers.fetch(header_key, nil) == token(context)		 +          context.session.delete(session_key)
```

src/amber/router/pipe/pipeline.cr
```diff
@@ -18,15 +18,15 @@ module Amber
...
 
         if context.websocket?
           context.process_websocket_request
         else
           @drain[context.valve].call(context) if @drain[context.valve]
         end
-
-        context
+      rescue e : Amber::Exceptions::Base
+        e.set_response(context.response)
```
eliasjpr: 👍

src/amber.cr
```diff
@@ -11,12 +11,17 @@ module Amber
...
     property name : String
+    setter   project_name : String?
     property env : String
     property log : Logger
     property secret : String
     property host : String = "0.0.0.0"
     property port_reuse : Bool = false
 
+    def project_name
+      @project_name ||= @name.gsub(/\W/, "_").downcase
```

src/amber/exceptions/exceptions.cr
elorest: Good point. Debugging code.

src/amber/router/pipe/csrf.cr
elorest: I added that because we needed to clear the csrf token out after it was used once.

eliasjpr: If we ever had to write a unit test for this method it would be difficult since the method is doing 2 things. Also I would expect the method to change anything base on the name for those reasons I think you should move`context.session.delete(session_key)` to line 22. 

```crystal
context.session.delete(session_key)
call_next(context)
```

PR #97, opened 2017-06-04
Removes Singeleton Pattern from Handlers

eliasjpr:
### Description of the Change

Currently handlers are following the singleton pattern. Singleton pattern is not necessary for handlers since handlers are initialized only once per pipileline.

### Requirements

* All new code requires tests to ensure against regressions

### Benefits

- Removes uncessary code.
- Makes it easy to maintain 
- Less complexity 
- Less code.

### DIFF
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -39,7 +39,7 @@ module Amber::Controller
         context = create_context(request)
         html_output = <<-HTML
         <form action="/posts" method="post">
-          <input type="hidden" name="_csrf" value="#{Amber::Pipe::CSRF.instance.token(context)}" />
+          <input type="hidden" name="_csrf" value="#{Amber::Pipe::CSRF.new.token(context)}" />
           <div class="form-group">
             <input class="form-control" type="text" name="title" placeholder="Title" value="hey you">
           </div>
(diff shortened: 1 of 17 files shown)


drujensen:
Remove this old comment:

https://github.com/Amber-Crystal/Amber/blob/master/src/amber/router/pipe/base.cr#L5

drujensen:
I don't think `self.config` will work anymore since you no longer have a singleton.  Consider removing this.
https://github.com/Amber-Crystal/Amber/blob/master/src/amber/router/pipe/base.cr#L11

elorest:
@drujensen makes a good point. Having them as singletons allowed us to configure in initializers by yielding self of a singleton.

elorest:
There's no reason we couldn't have a mix of singletons and normal instances. Session will probably need to stay a singleton because when I add encryption to sessions we're going to need to let them them choose to have it signed or encrypted and perhaps set session_key.

eliasjpr:
@drujensen thank you for pointing out the comments and the config method. I did a quick lookup in the source code and  we are not making making use of the `config` method anywhere in code. 

We already are providing some config options on intiialize for the handlers that needs it.

```crystal
key = "blog.session"
secret = "Some awesome secret key"
signed = https? 

 pipeline :static do
    plug HTTP::StaticFileHandler.new "../examples/public", false
    plug Amber::Pipe::Session.new secret, key, singed
 end
```

@elorest I agree we can mix both but I see no need for singletons at this point.  We are not going to keep track of sessions in memory within the sessions handler, the session handler should only have responsibility of encoding and decoding session cookie strings. We need to create a `SessionAdapter or SessionStore` to provide an interface for tracking sessions in cookie (Current implementation), memory, db, or redis.

PR #98, opened 2017-06-04
Fixed flash to go away after viewing. 

elorest:
### Description of the Change

Fixed Flash to go away the refresh after it's viewed.
Consolidated flash and session.
Added hash with indifferent access concept for flash and session.
Added `flash` and `session` to controller so you don't need to call `context.flash`

### Alternate Designs

At some point both `FlashHash` and `SessionHash` should probably inherit from a similar base object.


src/amber/router/pipe/session.cr
```diff
@@ -3,48 +3,79 @@ require "base64"
...
-      sha1 = OpenSSL::HMAC.hexdigest(:sha1, @secret, data)
-      "#{sha1}--#{data}"
+      class Hash < Hash(String, String)
+        def []=(key : String | Symbol, value : V)
+          super(key.to_s, value)
+        end
+
+        def [](key : Symbol | String)
+          key = key.to_s
+          @read << key
```
drujensen: Should the session hash keep track of the `@read` as well?

drujensen: curious how this compiled.

elorest: No it should not. I'm surprised it's still there. I fixed that hours ago lol

elorest: I think the tests are using fetch so they don't break.

src/amber/router/pipe/flash.cr
```diff
@@ -29,17 +29,69 @@ module Amber
...
+
+        def read 
+          @read
+        end
+
+        def []=(key : String | Symbol, value : V)
+          super(key.to_s, value)
+        end
+
+        def [](key : Symbol | String)
```
drujensen: What is someone calls `fetch` on the Hash to retrieve the value?

src/amber/router/pipe/session.cr
```diff
@@ -3,48 +3,77 @@ require "base64"
...
+      # worth of data in the session to avoid slow pageload times.
+      def session
+        @session ||= Hash.new 
+      end
 
-    private def encode(session)
-      data = Base64.encode(session.to_yaml)
-      sha1 = OpenSSL::HMAC.hexdigest(:sha1, @secret, data)
-      "#{sha1}--#{data}"
+      class Hash < Hash(String, String)
```
eliasjpr: Rename this to SessionHash to stay consistent with FlashHash. Beware that this is overwriting the crystal lib Hash class.

What do you think of this approach? 

```crystal 
module Amber
  module Pipe
     module SessionHelper
        def clear_session
          @session = Hash.new 
        end
        ...
     end

     class Session < Base
     end
  end
end
```

Then in the Context  `include Pipe::Session::Helpers`. This is a style convention that we all must agree upon so we stay consistent.

src/amber/router/pipe/session.cr
```diff
@@ -3,48 +3,77 @@ require "base64"
...
+      private def encode(session)
+        data = Base64.encode(session.to_yaml)
+        sha1 = OpenSSL::HMAC.hexdigest(:sha1, @secret, data)
+        "#{sha1}--#{data}"
+      end
     end
+  end
+
+  module Router
+    module Session
```
eliasjpr: Maybe we can get rid of the `module Router` and just leave one module, for instance `module ContextSession`. I feel that we are nesting too many modules.

src/amber/controller/base.cr
```diff
@@ -13,11 +13,15 @@ module Amber::Controller
     protected getter raw_params : HTTP::Params
     protected getter context : HTTP::Server::Context
     protected getter params : Amber::Validators::Params
+    protected getter flash : Amber::Router::Flash::Hash
```
eliasjpr: I don't think we should be using a crystal lib class name it feels that we are overwriting the class behavior. What about `Amber::Pipe::FlashHash`?

src/amber/router/pipe/flash.cr
bew: To handle this case, the simplest way would be to override `fetch` as well:
```cr
def [](key : Symbol | String)
  fetch(key.to_s, nil)
end

def fetch(key)
  @read << key
  previous_def key
end
```

src/amber/router/pipe/flash.cr
```diff
@@ -29,17 +29,69 @@ module Amber
...
+          key = key.to_s
+          @read << key
+          fetch(key, nil)
+        end
+
+        def each
```
bew: This is usually a bad idea to use private `Hash` stuff, because the Hash implementation is not a stable API, and things might change without warnings.
This could be rewritten as:
```cr
def each
  super do |key, value|
    @read << key
    yield({key, value})
  end
  self
end
```

src/amber/router/pipe/flash.cr
```diff
@@ -29,17 +29,69 @@ module Amber
...
+            yield({current.key, current.value})
+            @read << current.key
+            current = current.fore
+          end
+          self
+        end
+
+        def unread
+          #TODO: unread marks them as read. Maybe fix this. It shouldn't actually matter since it get's reloaded next request.
+          reject { |key, _| @read.includes? key }
```
bew: Can be rewritten as `reject @read`, it'll delete every keys contained in `@read`

src/amber/router/pipe/flash.cr
bew: To have abetter control on what's available for the user and how he get/fetch data from the flash store, we should wrap the underlying `Hash` into a more restricted class (e.g. `FlashStore`).

Plus this will be a more long-term solution, as overriding hash is discouraged, in favor of composing (wrap the hash inside one of our class, but do not inherit).

elorest: I'm not in any way over riding `Hash`. Hash continues to exist everywhere in it's original state. I'm creating Session::Hash nothing operates within that namespace. So hash would never be overwritten.

src/amber/router/pipe/session.cr
elorest: I'm not in any way over riding Hash. Hash continues to exist everywhere in it's original state. I'm creating Session::Hash and nothing operates within that namespace. So hash would never be overwritten. `Amber::Router::Flash::Hash` is way cleaner than `Amber::Router::Flash::FlashHash`

src/amber/controller/base.cr
elorest: Hash continues to exist everywhere in it's original state. I'm creating Fash::Hash and nothing operates within that namespace. So hash would never be overwritten.

src/amber/router/pipe/flash.cr
elorest: Cool. Thanks.

src/amber/router/pipe/flash.cr
elorest: I didn't just write this. It has been in the core for all of time.

bew: Yeah I know, I'm sorry about that, didn't review the first time it was introduced. Feel free to leave it like this, it's just a comment, I'm not "requesting changes right now"!

elorest: Ok. I do need to rework that because `flash.unread` marks them as read. Which isn't really an issue in how it's used but it makes testing weird.

bew: About `Flash::Hash#unread` maybe the name `flush_read` would be more consistent?

src/amber/router/pipe/session.cr
```diff
@@ -3,47 +3,78 @@ require "base64"
...
-    end
+module Amber
+  module Pipe
+    # The session handler provides a cookie based session.  The handler will
+    # encode and decode the cookie and provide the hash in the context that can
+    # be used to maintain data across requests.
+    class Session < Base
+      property secret : String
+      property key : String
+      property store : Symbol
```
elorest: The different store will probably end up becoming a different handler since it will be parsing and saving sessions to redis, memcache, or some other sort of data storage.

PR #101, opened 2017-06-04
Rename thread count to process count

eliasjpr:
Issue: https://github.com/Amber-Crystal/Amber/issues/100

Amber server is creating new processes not threads

* Rename Env var


eliasjpr:
Closed with https://github.com/Amber-Crystal/Amber/pull/101

PR #102, opened 2017-06-07
Implements Cookie Management

elorest:
# Cookie management for the Amber

**In progress**
Cookies are read and written through **Amber::Base::Controller#cookies**.

The cookies being read are the ones received along with the request, the cookies being written will be sent out with the response. Reading a cookie does not get the cookie object itself back, just the value it holds.

### Examples of writing:
```crystal
# Sets a simple session cookie.
# This cookie will be deleted when the user's browser is closed.
cookies[:user_name] = "david"

# Cookie values are String based. Other data types need to be serialized.
cookies[:lat_lon] = JSON.generate([47.68, -122.37])

# Sets a cookie that expires in 1 hour.
cookies[:login] = { value: "XJ-122", expires: 1.hour.from_now }

# Sets an encrypted cookie value before sending it to the client which
# prevent users from reading and tampering with its value.
# The cookie is signed by your app's `secrets.secret_key_base` value.
# It can be read using the encrypted method `cookies.encrypted[:name]`
cookies.encrypted[:discount] = 45

# Sets a "permanent" cookie (which expires in 20 years from now).
cookies.permanent[:login] = "XJ-122"

# You can also chain these methods:
cookies.permanent.signed[:login] = "XJ-122"
```

### Examples of Reading
```cr
cookies[:user_name]           # => "david"
cookies.size                  # => 2
JSON.parse(cookies[:lat_lon]) # => [47.68, -122.37]
cookies.encrypted[:discount]  # => 45
```
Please note that if you specify a :domain when setting a cookie, you must also specify the domain when deleting the cookie:

```cr
cookies[:name] = {
  value: 'a yummy cookie',
  expires: 1.year.from_now,
  domain: 'domain.com'
}
cookies.delete(:name, domain: 'domain.com')
```
The option symbols for setting cookies are:
```cr
:value - The cookie's value.

:path - The path for which this cookie applies. Defaults to the root of the application.

:domain - The domain for which this cookie applies so you can restrict to the domain level. If you use a schema like www.example.com and want to share session with user.example.com set :domain to :all. Make sure to specify the :domain option with :all or Array again when deleting cookies.

domain: nil  # Does not set cookie domain. (default)
domain: :all # Allow the cookie for the top most level
             # domain and subdomains.
domain: %w(.example.com .example.org) # Allow the cookie
                                      # for concrete domain names.
:tld_length - When using :domain => :all, this option can be used to explicitly set the TLD length when using a short (<= 3 character) domain that is being interpreted as part of a TLD. For example, to share cookies between user1.lvh.me and user2.lvh.me, set :tld_length to 1.

:expires - The time at which this cookie expires, as a Time object.

:secure - Whether this cookie is only transmitted to HTTPS servers. Default is false.

:httponly - Whether this cookie is accessible via scripting or only HTTP. Defaults to false.
```


src/amber/router/cookies.cr
```diff
@@ -0,0 +1,217 @@
...
+        @delete_cookies.each { |name, cookie| cookies << cookie.to_set_cookie_header }
+        headers.add("Set-Cookie", cookies)
+      end
+
+      def write_cookie?(cookie)
+        @secure || !cookie.secure
+      end
+    end
+
+    class JsonSerializer # :nodoc:
```
elorest: With only 4096 chars using yaml instead of json saves a significant amount of space and is pretty much exactly the same code.

src/amber/support/message_encryptor.cr
```diff
@@ -0,0 +1,67 @@
...
+      cipher.key = @secret
+
+      # Rely on OpenSSL for the initialization vector
+      iv = cipher.random_iv
+
+      encrypted_data = MemoryIO.new
+      encrypted_data.write(cipher.update(value))
+      encrypted_data.write(cipher.final)
+
+      "#{::Base64.strict_encode encrypted_data.to_slice}--#{::Base64.strict_encode iv}"
```
elorest: Using Base64 takes up a lot more space in our limited cookie space. It's already encrypted so what are we gaining by encoding it to base64?

elorest:
This ticket is currently implemented in that sessions are being stored in cookies and signed.
It would be nice to be able to completely encrypt as well. And it sort of looks like your going in that direction here? 

Currently lines 21-48  of session.cr (sign, set), and (verify, parse) signed cookies.
https://github.com/Amber-Crystal/amber/blob/master/src/amber/router/pipe/session.cr#L21

I don't think we need to store cookies on context since cookies are where the data is stored between requests and parsed into sessions during a request. What were your thoughts on that. We don't want to store duplicate data all over the place as it increases our memory footprint.

src/amber/support/message_encryptor.cr
eliasjpr: Base64 encoding schemes are commonly used when there is a need to encode binary data that needs be stored and transferred over media that are designed to deal with textual data. This is to ensure that the data remains intact without modification during transport.

elorest:
APIs use base64 to send images and other files all the time. This isn't
binary data though. Our only concern is that someone could modify their
cookies in the browser before sending them back. We overcome that by either
signing or encrypting but not both.

eliasjpr:
The base64 encoding is to avoid character encoding problems. When cookie value gets encrypted it is generate with special and wired characters that can be drop by the http handshake or a compresión algorithm while the data is being transmitted so the bade64 encoding is this problem. 

Ithe base64 here is not being use for security it is used to escape special characters

elorest:
Good point. Some forms of encryption do use none text characters. I was
just trying to think of a way to best use the small amount of space in
cookies but it's really not a big deal as we shouldn't ever store that much
in our session. Is there any good reason to sign an encrypted payload? With
a none encrypted payload signing allows us to ignore the payload if it
doesn't math the signature. If it's encrypted though it will be corrupt if
they modify it. I looks like rails 5 still uses a signature. I can't think
of a good reason though. Seems like more processing time and a bigger
payload with nothing to gain.

Isaac Sloan - isaacsloan.com
Artisan Rails Architect
1391 S Slate Canyon Dr.
Provo, Ut 84606
Cell: (435)-669-5005

eliasjpr:
This is ready for review:

- Adds rails like `cookies` method to manage cookies.
- Available to the Context, this is useful in case setting encrypting cookies in a pipeline Eg. Flash, Session, Generating JWT tokens
- Available to the controller for easy access to read and write cookies.

eliasjpr:
Closing and reopening this PR since I cannot add members to review

PR #104, opened 2017-06-08
Get All App Routes

eliasjpr:
Currently there is no way to get all the routes define for the
application. Being able to find all the routes is beneficial for
developer reference.

This was added in order to get all the application routes and print them
in the console using the amber_cmd.

**Issue** https://github.com/Amber-Crystal/amber/issues/103

```bash
# User types the routes command
amber routes

# Example possible output generated by command
Available routes For Your App Name
+-----------+-------------------+--------------------+------------------------+
| HTTP Verb | Controller#Action | Helper (url, path) | Resource               |
+-----------+-------------------+--------------------+------------------------+
| GET       | photos#index      | photos_url/path    | /photos, /photos/index |
+-----------+-------------------+--------------------+------------------------+
| GET       | photos#new        | new_photo_url/path | /photos/new            |
+-----------+-------------------+--------------------+------------------------+
```


elorest:
Does this effect anything performance wise or add complexity to the amber app its contained in? 

My strategy for this was going to be adding a very small subsection of route definition to amber_cmd and using it to translate `config/routes.cr` in the current directory to route string output... 

Your way is probably as good or better though. Thoughts?

elorest:
Why are the commits all about cookies?

eliasjpr:
Ha! I picked the wrong commit hash I'll fix it

elorest:
Oh ok.

eliasjpr:
@elorest this does not affect performance since the **all** method does not get used outside of the amber command line. Also, the Radix tree has a linked list is just matter of traversing should be fast.

PR #105, opened 2017-06-09
Fixed example site readme.

skunkworker:
### Description of the Change
The readme for the example didn't work as written. This fixes it and adds the `shards` command to make sure shards are installed first.


elorest:
Looks good although I know some of the file names in here changed today so we might want to wait for that to be merged in.

https://github.com/Amber-Crystal/amber/pull/104

PR #112, opened 2017-06-11
Adds benchmark results

eliasjpr:


elorest:
Do you think we should specify that it's with all optional pipes disabled and just returning "hello world"?

eliasjpr:
I don't know how specific you would like to be you ran the benchmark, I think you should add anything that you consider important in regards to the benchmark

elorest:
Ok. Specifically just the 2 things I said but I can add those when I get
home.

eliasjpr:
Sounds good, I will merge this and then you can open a PR with the additional stuff

elorest:
:+1:

eliasjpr:
kk

PR #113, opened 2017-06-12
pubsub adapters

fridgerator:
### Description of the Change

This implements the default pubsub adapter (`MemoryAdapter` to be used only in development) and the `RedisAdpater` to be used in production.  The Redis pubsub adapter allow sockets connected on different processes or even different servers to communicate with each other.

Simple UML diagram of this implementation: https://drive.google.com/open?id=0B2kt7nNk4YTSOEx2UGwwckhTZDg

p.s. I am not very good at UML diagrams

Without a websocket pubsub adapter, websocket communication is limited to only one server limiting scalability.

### Usage

Inside of the server config block, you can either use the default `MemoryAdapter` (recommended for development only), or `RedisAdapter`

`app.pubsub_adapter = Amber::WebSockets::Adapters::MemoryAdapter` (default)

\- or -

`app.pubsub_adapter = Amber::WebSockets::Adapters::RedisAdapter`

### DIFF
--- a/src/amber/websockets/client_sockets.cr
+++ b/src/amber/websockets/client_sockets.cr
@@ -1,13 +1,13 @@
 module Amber
   module WebSockets
     #
-    # Manages the entire collection of ClientSocket's.  Manages periodic timers for socket connections (heartbeat).
+    # Manages the entire collection of ClientSockets.  Manages periodic timers for socket connections (heartbeat).
     #
     module ClientSockets
       extend self
       @@client_sockets = Hash(UInt64, ClientSocket).new
       @@heartbeat_started = false
-      BEAT_INTERVAL = 3.seconds
+      BEAT_INTERVAL = 1.minute
 
       def add_client_socket(client_socket)
         @@client_sockets[client_socket.id] = client_socket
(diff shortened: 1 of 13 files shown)


src/amber/websockets/client_sockets.cr
```diff
@@ -36,8 +36,8 @@ module Amber
       private def heartbeat
         spawn do
           @@heartbeat_started = true
-          @@client_sockets.each_value(&.beat)
           sleep BEAT_INTERVAL
+          @@client_sockets.each_value(&.beat)
           heartbeat
```
fridgerator: I'm wondering if this will exit the previous `spawn`, or if this is going to get stacked up

fridgerator: This needs to be converted to use a loop instead

fridgerator: the spawn is safe, but using a loop is better - done in [nick/websockets_heartbeat](https://github.com/Amber-Crystal/amber/compare/nick/websockets_pubsub...nick/websockets_heartbeat?expand=1) branch

fridgerator:
Ran format tool and updated.  Merging...

PR #114, opened 2017-06-13
Implements Cookie Management

eliasjpr:
# Cookie management for the Amber

Original PR  https://github.com/Amber-Crystal/amber/pull/102
Cookies are read and written through **Amber::Base::Controller#cookies**.

The cookies being read are the ones received along with the request, the cookies being written will be sent out with the response. Reading a cookie does not get the cookie object itself back, just the value it holds.

### Examples of writing:
```crystal
# Sets a simple session cookie.
# This cookie will be deleted when the user's browser is closed.
cookies[:user_name] = "david"

# Cookie values are String based. Other data types need to be serialized.
cookies[:lat_lon] = JSON.generate([47.68, -122.37])

# Sets a cookie that expires in 1 hour.
cookies[:login] = { value: "XJ-122", expires: 1.hour.from_now }

# Sets an encrypted cookie value before sending it to the client which
# prevent users from reading and tampering with its value.
# The cookie is signed by your app's `secrets.secret_key_base` value.
# It can be read using the encrypted method `cookies.encrypted[:name]`
cookies.encrypted[:discount] = 45

# Sets a "permanent" cookie (which expires in 20 years from now).
cookies.permanent[:login] = "XJ-122"

# You can also chain these methods:
cookies.permanent.signed[:login] = "XJ-122"
```

### Examples of Reading
```cr
cookies[:user_name]           # => "david"
cookies.size                  # => 2
JSON.parse(cookies[:lat_lon]) # => [47.68, -122.37]
cookies.encrypted[:discount]  # => 45
```
Please note that if you specify a :domain when setting a cookie, you must also specify the domain when deleting the cookie:

```cr
cookies[:name] = {
  value: 'a yummy cookie',
  expires: 1.year.from_now,
  domain: 'domain.com'
}
cookies.delete(:name, domain: 'domain.com')
```
The option symbols for setting cookies are:
```cr
:value - The cookie's value.

:path - The path for which this cookie applies. Defaults to the root of the application.

:domain - The domain for which this cookie applies so you can restrict to the domain level. If you use a schema like www.example.com and want to share session with user.example.com set :domain to :all. Make sure to specify the :domain option with :all or Array again when deleting cookies.

domain: nil  # Does not set cookie domain. (default)
domain: :all # Allow the cookie for the top most level
             # domain and subdomains.
domain: %w(.example.com .example.org) # Allow the cookie
                                      # for concrete domain names.
:tld_length - When using :domain => :all, this option can be used to explicitly set the TLD length when using a short (<= 3 character) domain that is being interpreted as part of a TLD. For example, to share cookies between user1.lvh.me and user2.lvh.me, set :tld_length to 1.

:expires - The time at which this cookie expires, as a Time object.

:secure - Whether this cookie is only transmitted to HTTPS servers. Default is false.

:httponly - Whether this cookie is accessible via scripting or only HTTP. Defaults to false.
```


src/amber/router/context.cr
```diff
@@ -11,9 +11,11 @@ class HTTP::Server::Context
 
   property route : Radix::Result(Amber::Route)
   getter router : Amber::Router::Router
+  getter cookies : Amber::Router::Cookies::Store
 
   def initialize(@request : HTTP::Request, @response : HTTP::Server::Response)
     @router = Amber::Router::Router.instance
+    @cookies = Amber::Router::Cookies::Store.build(request, Amber::Server.key_generator)
```
elorest: I don't like that this will initialize and probably decode cookies even when session, flash, or other middleware aren't being used. In an api for instance we won't want to have it initialize a cookies as it will slow us down and serve no purpose... Maybe allow this to be nil and let middleware set values if chosen?

src/amber.cr
```diff
@@ -34,16 +21,37 @@ module Amber
       instance
     end
 
+    def self.key_generator
+      instance.key_generator
+    end
+
```
eliasjpr: I just move the class methods to the top

eliasjpr:
If you look closely the encryptor has an instance of the decryptor

src/amber/router/context.cr
eliasjpr: I read on the rails middleware and what we can do to make it optional is to create a Pipe (Hanlder) for the cookies.

elorest:
Sounds good. And then just throw an exception if they try using session
without it. Do you want to do that in another pr?

eliasjpr:
Instead of creating another handler, I just added a flag to the server, so if they want to use sessions or flash just need to enable cookies, this is better since it explicitlty states when to use cookies.

src/amber.cr
elorest: Oh ok. I might be wrong on this but it seems like properties, readers, writers should always be before methods.

src/amber/router/context.cr
elorest: Cool. That's what I shared above was from rails middleware. Lets just make it nilable here and controller base and create a cookie pipe in another PR where we refactor session/flash to write to the same cookie like rails. This should speed up those pipes as well.
```cr
getter cookies : Amber::Router::Cookies::Store?
```

elorest:
Ok I just saw your message about adding a flag to the server. That's probably a good solution.

src/amber/router/context.cr
eliasjpr: It is nillable  and the `Amber::Router::Cookies::Store.build` only sets cookies if the Server.cookies_enabled? returns true

src/amber/router/request.cr
```diff
@@ -0,0 +1,20 @@
+require "./**"
+
+class HTTP::Request
+  # This enables optional cookies store processing
+  @cookies_store : Amber::Router::Cookies::Store?
+
+  # Only enable advance cookies store if the Amber::Server.cookie_enabled?
```
elorest: Why patch HTTP::Request. Couldn't we just do the same thing on context? Instead of setting it here and copying to context and controller?

src/amber/router/context.cr
elorest: Yeah if you read down I eventually saw your last comment on that.

eliasjpr: I meant to remove it from here anyways and is nullable on the controllers

src/amber/router/request.cr
eliasjpr: Actually I was thinking of moving everything out the context and move it to the request. Cookies live in the request not in the context such as the params and other information. I decided not to move it now because it would have break a lot of things. 

Also by moving it to the request means that we don't have to worry about the cookies until we actually use in a pipeline and/or controller

src/amber.cr
eliasjpr: Yest we should create a config class that we can set a hash of settings something more manageable

src/amber/controller/base.cr
```diff
@@ -15,13 +15,15 @@ module Amber::Controller
...
     protected getter session : Amber::Router::Session::Params
+    protected getter cookies : Amber::Router::Cookies::Store?
 
     def initialize(@context : HTTP::Server::Context)
       @request = context.request
       @response = context.response
       @raw_params = context.params
       @flash = context.flash
       @session = context.session
+      @cookies = context.cookies
```
elorest: Lets not copy cookies to controller. We already have session and flash. If they want to manipulate other cookies they can call `context.cookies`.

src/amber/router/context.cr
```diff
@@ -20,6 +22,13 @@ class HTTP::Server::Context
     merge_route_params
   end
 
+  # Only enable advance cookies store if the Amber::Server.cookie_enabled?
```
elorest: Nice! Thanks.

src/amber/router/cookies.cr
```diff
@@ -0,0 +1,228 @@
+require "http"
```
drujensen: This is awesome!

What do you think of submitting this to the Crystal standard library as an enhancement to the existing Cookies store? https://crystal-lang.org/api/0.22.0/HTTP/Cookies.html

Or a separate shard that can be leveraged by other frameworks like kemal?

src/amber/controller/base.cr
eliasjpr: Rails has a cookies method in the controller to easily read and write cookies. The entire purpose of writing this was to make it available to the controllers. 

http://api.rubyonrails.org/v5.1/classes/ActionDispatch/Cookies.html
http://api.rubyonrails.org/classes/ActionController/Base.html

I'm gonna move this to a method instead of the initializing it here.

```cr
def cookies
  context.cookies
end
```

src/amber/router/cookies.cr
eliasjpr: We can convert it to a shard

eliasjpr:
@elorest addressed your concern of initializing the cookies in controllers.

src/amber/controller/base.cr
```diff
@@ -24,5 +25,9 @@ module Amber::Controller
       @session = context.session
       @params = Amber::Validators::Params.new(@raw_params)
     end
+
+    def cookies
+      @cookies ||= context.cookies
```
elorest: Ok. Probably cleaner to just delegate to `context.cookies`. Looks good though.

src/amber/router/context.cr
elorest: Remove this comment.

PR #115, opened 2017-06-14
Add HEADER method override pipe Handler

eliasjpr:
# Add HEADER method override pipe Handler

Adds a Request Method override handler to the framework this allows for
developers to customize the logic for overriding HTTP Verbs and decide if they want
to include method override since some proxy and HTTP servers allows new
HTTP methods.

- Adds Methods handler
- Methods Handler Specs
- Support _method param for overriding
- Supports X-HTTP-Method-Override for overriding

This removes the upgrade request method from the Context class by doing
this we can save processing speed by only including method override in
the desired pipeline.

### Usage

```cr
pipeline :web do
  plug Method.new
end
```

### Benefits

- Less processing to the Context initialization
- Greater flexibility on overriding HTTP methods
- This is now optional and customizable 

### Possible Drawbacks

- Developers must include the method handler


eliasjpr:
This is PR is open to discuss if this is the right direction

spec/amber/router/pipe/method_spec.cr
```diff
@@ -0,0 +1,91 @@
...
+            request = HTTP::Request.new("GET", "/?test=test", header, "_method=#{method}")
+            context = create_context(request)
+
+            Method.new.call(context)
+
+            context.request.method.should eq method
+          end
+        end
+
+        it "does not override other than PUT, PATCH, DELETE" do
```
drujensen: why do we want this restriction?

src/amber/router/context.cr
```diff
@@ -15,7 +15,6 @@ class HTTP::Server::Context
   def initialize(@request : HTTP::Request, @response : HTTP::Server::Response)
     @router = Amber::Router::Router.instance
     parse_params
-    upgrade_request_method!
     @route = router.match_by_request(@request)
```
drujensen: Is the route going to match since the method override doesn't happen until in the pipe?

src/amber/router/pipe/method.cr
```diff
@@ -0,0 +1,21 @@
+module Amber::Pipe
```
drujensen: Although this is 10 times cleaner, I'm worried that you have a chicken/egg issue.  The pipeline doesn't happen until after the route is determined, but the route is dependent on the method. 

I'm afraid this isn't going to work unless you add the handler before the router.

src/amber/router/context.cr
```diff
@@ -15,7 +15,6 @@ class HTTP::Server::Context
   def initialize(@request : HTTP::Request, @response : HTTP::Server::Response)
     @router = Amber::Router::Router.instance
     parse_params
```
elorest: If we're already parsing params here, what do we gain via removing upgrade_request_method here? That takes no time at all.

src/amber/router/context.cr
elorest: No.

src/amber/router/pipe/method.cr
elorest: My thoughts as well. Originally I had just required params to be parsed before route when they were both handlers. But @eliasjpr refactored it to be like this which made sense to me since they were coupled together.

Now it's still parsing params first which is the part that takes a little bit of time. Once we have the params upgrading the method/verb is basically a single line.

elorest:
We could perhaps optimize the existing system by only parsing enough params to get the method starting from the least time to the most time. Probably shouldn't be a priority though since it works fine how it is and is pretty fast.

spec/amber/router/pipe/method_spec.cr
elorest: Kemal does it is why I set it up with that list on the original code. Doesn't really seem to matter though... Doesn't hurt either?

eliasjpr:
I have opened a new PR https://github.com/Amber-Crystal/amber/pull/115 where maintains the current implementation and adds support for X-HTTP-Method-Override header. 

Closing this PR since is not a viable solution for us.

PR #117, opened 2017-06-14
HTTP Header Override

eliasjpr:
### Description of the Change

Based on the feedback of this PR https://github.com/Amber-Crystal/amber/pull/115 where it attempts to add method override as an HTTP::Handler 

## Description 
When deploying REST API based web services, you may encounter access
limitations on both the server and client sides. On the server side, the
access to your REST based web application might be done via a proxy web
server that does not support certain HTTP operations such as DELETE or
PUT for security reasons.

Such requests will lead to HTTP 405 error:
Http Status Code: 405
Reason: Request method is not allowed by the server

This PR adds support to offer these operations via POST operations and
ask your clients to add specific X-HTTP method override field in the
header of their requests.

### Benefits
- Keeps the current implementation 
- Maintains support for _method param for overriding
- Adds support http header X-HTTP-Method-Override for overriding

### Requirements
- Add support for HTTP methods PUT PATCH DELETE

### DIFF
--- a/src/amber/router/params.cr
+++ b/src/amber/router/params.cr
@@ -13,7 +13,8 @@ module Amber::Router
     MULTIPART_FORM   = "multipart/form-data"
     APPLICATION_JSON = "application/json"
     METHOD           = "_method"
-    OVERRIDE_METHODS = %w(patch put delete)
+    OVERRIDE_HEADER  = "X-HTTP-Method-Override"
+    OVERRIDE_METHODS = %w(PATCH PUT DELETE)
 
     alias ParamTypes = Nil | String | Int64 | Float64 | Bool | Hash(String, JSON::Type) | Array(JSON::Type)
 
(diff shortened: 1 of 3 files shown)


src/amber/router/params.cr
```diff
@@ -31,11 +32,37 @@ module Amber::Router
...
+    #
+    # Read RFC 2616 - HTTP 1.1 section 9.1.1
+    # (https://tools.ietf.org/html/rfc2616#section-9.1)
+    #
+    # In other words, if you are tempted to use a GET to simulate a PUT or DELETE,
+    # don't do it. Use a POST instead.
+    def override_request_method!
+      # If the current request method is not GET or POST it means that it was
+      # already overrided
+      return unless %(GET POST).includes? request.method
```
drujensen: Thanks for the detailed comment on why the override should be restricted to POST.

I disagree with this argument since the HTML conventions regarding 'safe' actions has been violated in almost every app I have seen.  For example, anchor links are supposed to be 'safe' and only perform retrieval but almost every app I've seen overrides this functionality and performs an 'unsafe' action with links.

I recommend removing this restriction that has caused much pain for developers over the years for no real good reason other than HTML hasn't been updated to support other methods.

eliasjpr: The code on line 58 is not restricting but bypassing if the current request is something other than a GET or POST, meaning that the request was upgraded before reaching to the amber server either by a proxy or http server that supports upgrades. You can still make GET request with `_method` and/or `X-HTTP-Method-Override` 

What this line is trying to avoid is to have a request.method = `PUT` with and override to some other method.

See line for GET support https://github.com/Amber-Crystal/amber/pull/117/files#diff-707c9269bb28ed65da046758b2df5c48R18 and https://github.com/Amber-Crystal/amber/pull/117/files#diff-707c9269bb28ed65da046758b2df5c48R56

drujensen: @eliasjpr your right.  All good!  thanks.

elorest: As long as I can still have `<a href="posts/?_method='delete'"/>` I'm good with it. 

I don't like the warning against overriding get with put or delete though. Technically because of IIS everything but POST, GET is considered unsafe. In rails, phoenix, nginx, apache anything thought it's completely as safe as any other verb.

PR #121, opened 2017-06-21
Adds different storage mechanisms for session management

eliasjpr:
## Description

Your application has a session for each user in which you can store small amounts of data that will be persisted between requests. The session is only available in the controller and the view and can use one of a number of different storage mechanisms:
****
This PR adds the following storage mechanism:
- Adds Cookies Session using new cookies storage
- Adds Redis Session storage

```
Amber::Session::CookieStore
Amber::Session::RedisStore
# Todo 
Amber::Session::DBStore
Amber::Session::MemCacheStore
```

All session stores use a cookie to store a unique ID for each session (you must use a cookie, Amber will not allow you to pass the session ID in the URL as this is less secure).

For most stores, this ID is used to look up the session data on the server, e.g. in a database table. There is one exception, and that is the default and recommended session store - the CookieStore - which stores all session data in the cookie itself (the ID is still available to you if you need it). This has the advantage of being very lightweight and it requires zero setup in a new application in order to use the session. The cookie data is cryptographically signed to make it tamper-proof. And it is also encrypted so anyone with access to it can't read its contents. (Amber will not accept it if it has been edited).

**Impportant**

About Session Cookie Store
- You can only store about 4kb of data in a cookie.
- Cookies are sent along with every request you make.
- If you accidentally expose your secret_key_base, your users can change the data you’ve put inside your cookie.

About Redis Store
- You can store more than 4kb of data in the session, only the session id gets stored in the cookie.
- This means smaller request 

## Configuration

```cr
@session = {
        :key     => "#{project_name}.session",
        :store   => :cookie, # :redis, must provide redis URL
        :expires => 2 * 60 * 60, # 2 Hours
        :secret  => secret,
        :redis_url => "redis://localhost:6379"
      }
```
**Accessing Session**
Session values are stored using key/value pairs like a hash:
```cr
  # Finds the User with the ID stored in the session with the key
  # **:current_user_id** This is a common way to handle user login in
  # an Amber application; logging in sets the session value and
  # Logging out removes it.
  private def current_user
    @_current_user ||= session[:current_user_id] &&
      User.find_by(id: session[:current_user_id])
  end
end
```
To store something in the session, just assign it to the key like a hash:
```cr
class LoginsController < ApplicationController
  # "Create" a login, aka "log the user in"
  def create
    if user = User.authenticate(params[:username], params[:password])
      # Save the user ID in the session so it can be used in
      # subsequent requests
      session[:current_user_id] = user.id
      redirect_to root_url
    end
  end
end
```
To remove something from the session

```cr

class LoginsController < ApplicationController
  # "Delete" a login, aka "log the user out"
  def destroy
    # Remove the user id from the session
    session.delete[:current_user_id] 
    redirect_to root_url
  end
end
```

## Extending Creating your own session store

To create new session store must follow the **Amber::Router::Session::AbstractStore** interface.

### DIFF
--- a/src/amber/router/pipe/csrf.cr
+++ b/src/amber/router/pipe/csrf.cr
@@ -5,35 +5,41 @@ module Amber
     # The CSRF Handler adds support for Cross Site Request Forgery.
     class CSRF < Base
       CHECK_METHODS = %w(PUT POST PATCH DELETE)
-      property session_key, header_key, param_key, check_methods
-
-      def initialize
-        @session_key = "csrf.token"
-        @header_key = "HTTP_X_CSRF_TOKEN"
-        @param_key = "_csrf"
-      end
+      HEADER_KEY    = "X-CSRF-TOKEN"
+      PARAM_KEY     = "_csrf"
+      CSRF_KEY      = "csrf.token"
 
       def call(context : HTTP::Server::Context)
-        if !CHECK_METHODS.includes?(context.request.method) || valid_token?(context)
-          context.session.delete(session_key)
+        if valid_http_method?(context) || valid_token?(context)
+          context.session.delete(CSRF_KEY)
           call_next(context)
         else
           raise Amber::Exceptions::Forbidden.new("CSRF check failed.")
+          context.response.status_code = 403
+          context.response.content_type = "text/plain"
+          context.response.print("CSRF check failed.")
         end
       end
 
+      def valid_http_method?(context)
+        !CHECK_METHODS.includes?(context.request.method)
+      end
+
       def valid_token?(context)
-        (context.params[param_key]? || context.request.headers[header_key]?) == token(context)
-      rescue
-        false
+        (context.params[PARAM_KEY]? || context.request.headers[HEADER_KEY]?) == self.class.token(context)
       end
 
-      def token(context)
-        context.session[session_key] ||= SecureRandom.urlsafe_base64(32)
+      def self.token(context)
+        context.session[CSRF_KEY] ||= SecureRandom.urlsafe_base64(32)
+        context.session[CSRF_KEY]
       end
 
-      def tag(context)
-        %Q(<input type="hidden" name="#{param_key}" value="#{token(context)}" />)
+      def self.tag(context)
+        %Q(<input type="hidden" name="#{PARAM_KEY}" value="#{token(context)}" />)
+      end
+
+      def self.metatag(context)
+        %Q(<meta name="#{HEADER_KEY}" content="#{token(context)}" />)
       end
     end
   end
--- a/src/amber/router/session/cookie_store.cr
+++ b/src/amber/router/session/cookie_store.cr
@@ -1,0 +1,80 @@
+module Amber::Router::Session
+  # This is the default Cookie Store
+  class CookieStore < AbstractStore
+    property secret : String
+    property key : String
+    property expires : Int32
+    property store : Amber::Router::Cookies::Store
+    property session : Hash(String, String)
+
+    def initialize(@store, @key, @expires, @secret)
+      @session = current_session
+      @session[key] = SecureRandom.uuid
+    end
+
+    def id
+      @session[key]
+    end
+
+    def destroy
+      @session.clear
+    end
+
+    def [](key : String | Symbol)
+      @session.fetch(key.to_s)
+    end
+
+    def []?(key : String | Symbol)
+      fetch(key.to_s, nil)
+    end
+
+    def []=(key : String | Symbol, value)
+      @session[key.to_s] = value.to_s
+    end
+
+    def key?(key : String | Symbol)
+      @session.has_key?(key.to_s)
+    end
+
+    def keys
+      @session.keys
+    end
+
+    def values
+      @session.values
+    end
+
+    def to_h
+      @session
+    end
+
+    def update(hash : Hash(String | Symbol, String))
+      hash.each do |key, value|
+        @session[key.to_s] = value
+      end
+      @session
+    end
+
+    def delete(key : String | Symbol)
+      @session.delete(key.to_s) if key?(key.to_s)
+    end
+
+    def fetch(key : String | Symbol, default = nil)
+      @session.fetch(key.to_s, default)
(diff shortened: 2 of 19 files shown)


src/amber/router/pipe/csrf.cr
```diff
@@ -5,35 +5,41 @@ module Amber
...
 
-      def token(context)
-        context.session[session_key] ||= SecureRandom.urlsafe_base64(32)
+      def self.tag(context)
+        %Q(<input type="hidden" name="#{PARAM_KEY}" value="#{token(context)}" />)
       end
 
-      def tag(context)
-        %Q(<input type="hidden" name="#{param_key}" value="#{token(context)}" />)
+      def self.metatag(context)
```
elorest: metatag is more vague than tag. Should probably be csrf_tag. Although CSRF.tag is more readable.

src/amber/router/pipe/csrf.cr
```diff
@@ -5,35 +5,41 @@ module Amber
...
+        %Q(<meta name="#{name}" content="#{token(context)}" />)
```
elorest: Why use a metatag?

src/amber/router/session/cookie_store.cr
```diff
@@ -0,0 +1,80 @@
...
+
+    def id
+      @session[key]
+    end
+
+    def destroy
+      @session.clear
+    end
+
+    def [](key)
```
elorest: The current session hash allows symbols or strings which saves typing and will be familiar to people coming from rails. Lets discuss the current SessionHash further before removing that.

src/amber/router/pipe/csrf.cr
eliasjpr: These are used to generate the dynamic forms that implement non-remote links with :method. For AJAX requests other than GETs, extract the “csrf-token” from the meta-tag and send as the “X-CSRF-Token” HTTP header.

src/amber/router/pipe/csrf.cr
eliasjpr: Metatag is different from the input tag, metatag is used to assign a crsf header for the Ajax requests

src/amber/router/session/cookie_store.cr
eliasjpr: True, this is one detail about rails that I dislike, the inconsistency, I have seen people ran into issues because they think **session[:key]** is the same as **session["key"]** those 2 have different pointers. Same for params, I think we should keep it simple and remove the ambiguity.

drujensen: @eliasjpr I prefer symbols over strings for cleaner looking code and would like :key and "key" to point to the same value.

eliasjpr: I like symbols as well, in the case above since for cookie store the entire session is turned into a JSON when is it gets parse back to a hash it turns the keys into a string. Maybe there is a way in crystal to do this have string and symbol point to the same value. See also https://stackoverflow.com/questions/9712706/ruby-symbols-vs-strings-in-hashes

elorest: @eliasjpr In Rails `session[:key]` is the same as `session["key"]`. It extends HashWithIndifferentAccess. It super easy to do that here, but just storing it as a string but converting symbols to strings before looking it up. Like we're currently doing here: https://github.com/Amber-Crystal/amber/blob/master/src/amber/router/pipe/session.cr#L74

src/amber/router/pipe/csrf.cr
elorest: So do we still actually have a hidden field as well? I'm familiar with passing csrf in the header or the params.

src/amber/router/session/cookie_store.cr
elorest: It would probably only be a couple lines of code to move that over into `session/cookie_store` here.

src/amber/router/session/redis_store.cr
```diff
@@ -0,0 +1,81 @@
...
+      @store.hmset(session_id, hash)
+    end
+
+    def delete(key)
+      @store.hdel(session_id, key) if key?(key)
+    end
+
+    def fetch(key, default = nil)
+      return self[key] if key?(key)
+      (self[key] = default)
```
elorest: If a key doesn't exist why are we setting it with a value of default? Shouldn't it continue to not exist and just return default? Like with in the stdlib hash: https://github.com/crystal-lang/crystal/blob/d1f8c42f82d3e6491cb7b3b536984c19a51b94ba/src/hash.cr#L137

src/amber/router/pipe/csrf.cr
eliasjpr: Yes, hidden fields for forms, meta tag for ajax, user can make them both available when needed, they can both coexist as well

elorest: Ok. That's good then.

elorest:
I tried using the new session and got an error.
```sh
    session["test"] = "hello"
    ^~~~~~~

in lib/amber/src/amber/controller/base.cr:32: Can't infer the type of instance variable '@session' of HomeController

The type of a instance variable, if not declared explicitly with
`@session : Type`, is inferred from assignments to it across
the whole program.

The assignments must look like this:

  1. `@session = 1` (or other literals), inferred to the literal's type
  2. `@session = Type.new`, type is inferred to be Type
  3. `@session = Type.method`, where `method` has a return type
     annotation, type is inferred from it
  4. `@session = arg`, with 'arg' being a method argument with a
     type restriction 'Type', type is inferred to be Type
  5. `@session = arg`, with 'arg' being a method argument with a
     default value, type is inferred using rules 1, 2 and 3 from it
  6. `@session = uninitialized Type`, type is inferred to be Type
  7. `@session = LibSome.func`, and `LibSome` is a `lib`, type
     is inferred from that fun.
  8. `LibSome.func(out @session)`, and `LibSome` is a `lib`, type
     is inferred from that fun argument.

Other assignments have no effect on its type.

Can't infer the type of instance variable '@session' of HomeController

      @session ||= context.session
      ^~~~~~~~
```

PR #122, opened 2017-06-26
Session only writes to cookies if it has changed. 

elorest:
### Description of the Change

Session doesn’t need to serialize, sign and write to cookies if it hasn’t changed. This saves a ton of time.

### Benefits

Results in 3x (10,000req/s to 33,000req/s MBP singlecore) speed increase when session isn't being set and no noticeable difference when it is.

### Possible Drawbacks

This code will all be replaced soon.

### DIFF
--- a/src/amber/router/pipe/session.cr
+++ b/src/amber/router/pipe/session.cr
@@ -1,6 +1,6 @@
 require "http/cookie"
 require "base64"
-require "yaml"
+require "json"
 require "openssl/hmac"
 
 module Amber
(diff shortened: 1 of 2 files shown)


src/amber/router/pipe/session.cr
```diff
@@ -63,8 +65,12 @@ module Amber
       end
 
       class Params < Hash(String, String)
+        property changed = false
```
drujensen: :+1:

src/amber/router/pipe/session.cr
```diff
@@ -63,8 +65,12 @@ module Amber
...
+
         def []=(key : String | Symbol, value : V)
```
drujensen: Is there other ways of adding or changing values in a hash?  What about `merge!`?

src/amber/router/pipe/session.cr
```diff
@@ -22,25 +22,27 @@ module Amber
...
+          cookies << HTTP::Cookie.new(@key, value)
+          cookies.add_response_headers(context.response.headers)
+        end
       end
 
       private def decode(session, data)
         sha1, data = data.split("--", 2)
         if sha1 == OpenSSL::HMAC.hexdigest(:sha1, @secret, data)
-          values = YAML.parse(Base64.decode_string(data))
+          values = JSON.parse(Base64.decode_string(data))
```
drujensen: :+1:

drujensen:
very clean and simple solution with such a huge benefit!  nice job.

skunkworker:
So instead of YAML we are using JSON inside of the cookies?

elorest:
Very little difference between yaml and json. I figured I'd just go back to json since other parts of the repo had it.

src/amber/router/pipe/session.cr
elorest: store, an replace! aren't available to hash. `merge!` will work fine since it calls `[]=`. `merge` is fine since it doesn't actually modify the existing hash.

drujensen: excellent!

bew:
Also JSON parser is faster than YAML parser, but I'm not sure if it's relevant here because a cookie is usually small in data.

elorest:
@bew In a few benchmarks I found that for single depth key value pairs (what we're using in sessions) yaml was a bit faster and took up less space. Multilevel hashes of hashes were faster with json though.

bew:
@elorest I see, thanks for the precision! The time difference might come from the indentation parsing in yaml, that doesn't exists in json.

PR #123, opened 2017-06-27
Changed filename to match class name.

elorest:
### Description of the Change

File name is changed to message_verifier to match `MessageVerifier` class.


elorest:
I'm merging this tonight if there are no other comments.

PR #126, opened 2017-07-05
Fixes Static File Handler

eliasjpr:
Static File Handler has changed in crystal version 0.23

This adds a new argument called directory listing to the static handler
which allows to comply to the new changes in Crystal 0.23

* Directory listting can now be specified when adding the handler to the
pipeline

```crystal
pipeline :web do
  pipe Amber::Pipe::Static.new "/public", true # public directory, directory listing.
end
```

This makes it backwards compatible with crystal 0.22 and 0.23

Thank you @sdogruyol

### DIFF
--- a/src/amber/router/pipe/static.cr
+++ b/src/amber/router/pipe/static.cr
@@ -13,13 +13,6 @@ module Amber
         @directory_listing = directory_listing
       end
 
-      def initialize
-        @public_dir = File.expand_path "./public"
-        @default_file = "index.html"
-        @fallthrough = true
-        @static_config = {"dir_listing" => false, "gzip" => true}
-      end
-
       def call(context : HTTP::Server::Context)
         return call_next(context) if context.request.path.not_nil! == "/"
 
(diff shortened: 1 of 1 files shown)


sdogruyol:
Thank you 👍

mixflame:
:beers :pizza:

src/amber/router/pipe/static.cr
```diff
@@ -5,11 +5,12 @@ module Amber
...
-      def initialize(public_dir : String, fallthrough = true)
+      def initialize(public_dir : String, directory_listing = false, fallthrough = true)
         @public_dir = File.expand_path public_dir
         @fallthrough = !!fallthrough
         @default_file = "index.html"
         @static_config = {"dir_listing" => false, "gzip" => true}
+        @directory_listing = directory_listing
       end
 
       def initialize
```
gabrielrios: I think you're missing `@directory_listing = directory_listing` on this method.

straight-shoota:
Why is there a constructor overwrite at all? The only thing different from `HTTP::StaticFileHandler` is a static default value for `@static_config` which can easily be set at class scope.
The Crystal idiom would be overwriting `self.new` which also allows to add a parameter-less overload and keep the original `initialize` method from the super class.

eliasjpr:
@straight-shoota @gabrielrios  definitely was not the intention to overload the initialize method, and has now been removed! Thank you for pointing that out.

elorest:
Cool

PR #138, opened 2017-07-15
Rails like session/flash management for Amber.

eliasjpr:
### Description of the Change

> This PR adds Rails like session/flash management to Amber.

I have used the following amber blog app to test the functionality https://github.com/eliasjpr/blog

## The Session

A session store that uses an Amber:Router::Session::Store to store the sessions. This store is most useful if you don't store critical data in your sessions and you don't need them to live for extended periods of time.

This cookie-based session store is the Amber default. It is dramatically faster than the alternatives.

The cookie store used for storage is automatically configured to be the best possible option given your application's configuration.

Sessions typically contain at most a user_id and flash message; both fit within the 4K cookie size limit. A CookieOverflow exception is raised if you attempt to store more than 4K of data.

To configure the session:
```crystal
#  Cookie Store
Amber::Server.instance.session = {
  :key     => "name.session",
  :store   => :cookie,
  :expires => 120, # 0, will make the session last as long as the browser is open, upon closing, session will be terminated
  :secret  => "secret"
}

# Redis Store
Amber::Server.instance.session = {
  :key       => "name.session",
  :store     => :redis,
  :expires   => 120,
  :secret    => "secret",
  :redis_url => "redis://localhost:6379",
}
```
Configure the Pipeline

```crystal
 # Keep in mind the order of the Pipes. Session hash needs to be populated before trying to access the session flash scope, the flash depends on the session. 
 pipeline :web do
    plug Amber::Pipe::Session.new
    plug Amber::Pipe::Logger.new
    plug Amber::Pipe::Flash.new
    plug Amber::Pipe::CSRF.new
  end
```

Accessing the session

```crystal
 
  private
 
  # Finds the User with the ID stored in the session with the key
  # :current_user_id This is a common way to handle user login in
  # a Amber application; logging in sets the session value and
  # logging out removes it.
  def current_user
    @_current_user ||= session[:current_user_id] &&
      User.find_by(id: session[:current_user_id])
  end
end
```
To store something in the session, just assign it to the key like a hash:

```crystal
class LoginsController < ApplicationController
  # "Create" a login, aka "log the user in"
  def create
    if user = User.authenticate(params[:username], params[:password])
      # Save the user ID in the session so it can be used in
      # subsequent requests
      session[:current_user_id] = user.id
      redirect_to root_url
    end
  end
end
```

To remove something from the session, assign that key to be nil or use `session.delete(key)`

```crystal
class LoginsController < ApplicationController
  # "Delete" a login, aka "log the user out"
  def destroy
    # Remove the user id from the session
    @_current_user = session[:current_user_id] = nil
    redirect_to root_url
  end
end
```

## The Flash
The flash is a special part of the session which is cleared with each request. This means that values stored there will only be available on the next request, which is useful for passing error messages etc.

It is accessed in much the same way as the session, as a hash.

Let's use the act of logging out as an example. The controller can send a message which will be displayed to the user on the next request:

```crystal 
class LoginsController < ApplicationController
  def destroy
    session[:current_user_id] = nil
    #  Alternatively, `flash.notice=` could be use.
    flash[:notice] = "You have successfully logged out."
    redirect_to root_url
  end
end
```
Rendering the flash message

```erb
<html>
  <!-- <head/> -->
  <body>
    <% flash.each do |name, msg| -%>
      <%= content_tag :div, msg, class: name %>
    <% end -%>
 
    <!-- more content -->
  </body>
</html>
```
If you want a flash value to be carried over to another request, use the keep method:

```crystal
class MainController < ApplicationController
  # Let's say this action corresponds to root_url, but you want
  # all requests here to be redirected to UsersController#index.
  # If an action sets the flash and redirects here, the values
  # would normally be lost when another redirect happens, but you
  # can use 'keep' to make it persist for another request.
  def index
    # Will persist all flash values.
    flash.keep
 
    # You can also use a key to keep only some kind of value.
    # flash.keep(:notice)
    redirect_to users_url
  end
end
```

### Flash.now
By default, adding values to the flash will make them available to the next request, but sometimes you may want to access those values in the same request. For example, if the create action fails to save a resource and you render the new template directly, that's not going to result in a new request, but you may still want to display a message using the flash. To do this, you can use flash.now in the same way you use the normal flash.

```crystal
class ClientsController < ApplicationController
  def create
    @client = Client.new(params[:client])
    if @client.save
      # ...
    else
      flash.now[:error] = "Could not save client"
      render action: "new"
    end
  end
end
```

### Possible Drawbacks

Current all sessions are encrypted. It has been proposed to sign and/or only encrypt by passing a config param.


.travis.yml
```diff
@@ -1,4 +1,4 @@
-language: generic
+language: crystal
```
elorest: This should probably be in a different PR. I know Dru was in favor of using docker which was why I used this fix. Crystal Core team also uses docker for their travis tests.

src/amber.cr
```diff
@@ -49,10 +55,17 @@ module Amber
...
       @key_generator = Amber::Support::CachingKeyGenerator.new(
-        Amber::Support::KeyGenerator.new(secret, 1000)
+        Amber::Support::KeyGenerator.new(secret, 5)
       )
       @pubsub_adapter = WebSockets::Adapters::MemoryAdapter
       @redis_url = "redis://localhost:6379"
+      @session = {
+        :key       => "session_id",
+        :store     => :cookie,
+        :expires   => 2 * 60 * 60, # 2 Hours
```
elorest: Session should last as long as the browser is open like the other MVC's instead of 2 hours.

src/amber/router/pipe/session.cr
```diff
@@ -1,85 +1,14 @@
...
-        end
-
-        def [](key)
-          fetch(key, nil)
-        end
-
-        def find_entry(key)
-          super(key.to_s)
-        end
+        context.session.set_session
```
elorest: Writing cookies is expensive time wise, now more than ever due to encryption. We should only save cookies if something has changed.

src/amber/router/session/cookie_store.cr
```diff
@@ -0,0 +1,65 @@
+module Amber::Router::Session
+  class SessionHash < Hash(String, String)
```
elorest: I like that you utilized this.

spec/amber/controller/base_spec.cr
```diff
@@ -45,9 +76,10 @@ module Amber::Controller
       it "renders a form with a csrf tag" do
         request = HTTP::Request.new("GET", "/?test=test")
         context = create_context(request)
+        csrf_token = Amber::Pipe::CSRF.token(context)
         html_output = <<-HTML
         <form action="/posts" method="post">
-          <input type="hidden" name="_csrf" value="#{Amber::Pipe::CSRF.new.token(context)}" />
+          <input type="hidden" name="_csrf" value="#{csrf_token}" />
```
elorest: Like this more.

.travis.yml
eliasjpr: I decided to remove it since the crystal v0.23.1 works on Travis. Redis server was not being installed with the docker option.

src/amber/router/pipe/session.cr
eliasjpr: Let's iterate over this in a separate PR. In those upcoming PRs the signed and encryption should be addressed as well

eliasjpr:
Comments have been addressed. Now when passing expires: 0 sessions will only last until the browser is closed. 
screen_shot_2017-07-16_at_8_53_08_am

.travis.yml
drujensen: I recommend adding the  redis service to dockercompose.yml and reenabling docker for running our tests.  This will eliminate failures because of differences between environments and gives us finer control of the testing environment.

drujensen:
LGTM

Please squash the commits before merging. 

Add tickets for re-enabling cookie optimization and encryption options.

.travis.yml
eliasjpr: @drujensen I would not know how to do this. Let me know when you have some time I would like if you can walk me thru this setup.

eliasjpr:
@drujensen the reason that the flash message is not being rendered for the new generated app is that the session pipe needs to come before the flash and csrf pipe, as stated above in the PR description

```cr
 # Keep in mind the order of the Pipes. Session hash needs to be populatedd before trying to access the session flash scope, the flash depends of the session. 
 pipeline :web do
    plug Amber::Pipe::Session.new
    plug Amber::Pipe::Logger.new
    plug Amber::Pipe::Flash.new
    plug Amber::Pipe::CSRF.new
  end
```

drujensen:
@eliasjpr moving the session fixed it.  I also tested with redis and everything seems to be working correctly.

PR #144, opened 2017-07-20
use docker for travis

drujensen:
This PR changes travis to use docker and docker-compose to allow more control over the test environment.


drujensen:
@elorest do you have a recommendation on where you think this should go? `/docker` or `/config` or ???

elorest:
In this case I wonder if putting it in spec/ would be a good idea.

drujensen:
I moved the docker files to `/spec`

PR #146, opened 2017-07-22
Render using absolute path

faustinoaq:
### Description of the Change

This PR add absolute path to macro render_template.

### Alternate Designs

The absolute path is used if path is equal to `"src/view"`

Currently in my amber projects I have to rewrite the macro `render_template` adding support for absolute path, because [vscode-crystal-lang](https://github.com/faustinoaq/vscode-crystal-lang) extension can't find relative paths when analyze code.

My `application_controller.cr`

```crystal
  LAYOUT = "application.ecr"

  macro render_template(filename, path = "src/views")
    {% if filename.id.split("/").size > 2 %}
      Kilt.render("{{filename.id}}")
    {% else %}
      {% if path == "src/views" %}
        Kilt.render("{{__FILE__.split('/')[0..-4].join('/').id}}/#{{{path}}}/{{filename.id}}")
      {% else %}
        Kilt.render("#{{{path}}}/{{filename.id}}")
      {% end %}
    {% end %}
  end
end
```

Some screenshots:

Show macro errors related to relative path of views:

screenshot_20170721_205722

Errors fixed using absolute path:

screenshot_20170721_205917

### Benefits

This change add absolute path only if path is equal to `"src/view"` so, if path change the render process uses the path given.

### Possible Drawbacks

Absolute path is calculated using `__FILE__` only when path is default to views folder, if path changes then absolute path isn't used.

### Applicable Issues

Editor extensions like [Scry](https://github.com/kofno/scry) can't find relative paths used by macros, so this PR help to users to avoid unnecessary macro errors.

### DIFF
--- a/src/amber/controller/render.cr
+++ b/src/amber/controller/render.cr
@@ -6,7 +6,11 @@ module Amber::Controller
       {% if filename.id.split("/").size > 2 %}
         Kilt.render("{{filename.id}}")
       {% else %}
-        Kilt.render("#{{{path}}}/{{filename.id}}")
+        {% if path == "src/views" %}
+          Kilt.render("{{__FILE__.split('/')[0..-4].join('/').id}}/#{{{path}}}/{{filename.id}}")
+        {% else %}
+          Kilt.render("#{{{path}}}/{{filename.id}}")
+        {% end %}
       {% end %}
     end
 


src/amber/controller/render.cr
```diff
@@ -6,7 +6,11 @@ module Amber::Controller
       {% if filename.id.split("/").size > 2 %}
         Kilt.render("{{filename.id}}")
       {% else %}
-        Kilt.render("#{{{path}}}/{{filename.id}}")
+        {% if path == "src/views" %}
+          Kilt.render("{{__FILE__.split('/')[0..-4].join('/').id}}/#{{{path}}}/{{filename.id}}")
```
elorest: `__FILE__.split('/')[0..-4]` assumes that `__FILE__` is always going to be at the same level. If someone were to render a view from presenter, or even another ecr file it will fail.

faustinoaq:
Yes, I wil try to fix my editor since this PR seems a bit hacky
:sweat_smile:

faustinoaq:
This error also is happening with Kemal and other Crystal programs, so I opened a issue in vscode-crystal-lang instead :point_right:  https://github.com/faustinoaq/vscode-crystal-lang/issues/5

faustinoaq:
Now, My editor extension is fixed https://github.com/faustinoaq/vscode-crystal-lang/pull/10 :smile:

PR #150, opened 2017-07-26
added options to javascript Socket

fridgerator:
### Description of the Change

* Added options to `Amber.Socket#connect` method, with defaults.
  * Previously the socket connection would fail on ssl (wss) connections.  
  * I added some options to set port, location and protocol when connecting, defaulting to values stored in `window.location`

Let me know what you all think about this:

* Added the minified amber.min.js file webpack config file I've been using to generate that file.
  * install node plugins `webpack` and `uglifyjs-webpack-plugin` globally
  * build `amber.min.js` with `webpack --config webpack.config.js`


drujensen:
@fridgerator are we going to publish an npm package for `amber.min.js`?

fridgerator:
@drujensen I wasn't sure what to do with this file.  It will definitely need to be distributed some way, perhaps we can discuss this in the next meeting.

elorest:
Why not just distribute it through amber cmd? I definitely prefer not to
have an implicit npm dependency if possible.

On Jul 28, 2017 11:48 AM, "Nick Franken" <notifications@github.com> wrote:

@drujensen <https://github.com/drujensen> I wasn't sure what to do with
this file. It will definitely need to be distributed some way, perhaps we
can discuss this in the next meeting.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/amber-crystal/amber/pull/150#issuecomment-318719538>,
or mute the thread

Issue #151, opened 2017-07-30
Support SSL like Kemal.


faustinoaq:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem
    * Checked that your issue isn't already filed: https://github.com/Amber-Crystal/amber/issues
    * Checked that there is not already a Crystal Shard that provides the described functionality: http://crystalshards.xyz/

### Description

Currently Amber doesn't support SSL, see [gitter comment](https://gitter.im/Amber-Crystal/Lobby?at=597c8e2476a757f80858cd66)

### Steps to Reproduce

1. Install Amber:CMD
2. Create new Amber project
3. Try to configure SSL on Amber

**Expected behavior:** Amber supporting SSL

**Actual behavior:** Need to use external program like Nginx to use SSL

**Reproduces how often:** 100%

### Versions

0.1.5

### Additional Information

@sdogruyol, the creator of Kemal said:

> it's simple enough
> just pass SSL Context to HTTP::Server

Currently, Kemal does the following:

```
./your_kemal_app --ssl --ssl-key-file your_key_file --ssl-cert-file your_cert_file
```

See: http://kemalcr.com/docs/ssl/

eliasjpr:
Solved with https://github.com/amber-crystal/amber/pull/152

PR #152, opened 2017-07-30
Support SSL

c910335:
### Description of the Change

Implemented #151.

#### Usage

config/application.cr

```crystal
Amber::Server.instance.config do |app|
  # ...
  app.ssl_key_file = "config/privkey.pem"
  app.ssl_cert_file = "config/fullchain.pem"
end
```

SSL is important for every website, but it is too much work with Nginx or Apache or something like them.

### Benefits

Start with SSL easily.


src/amber.cr
```diff
@@ -89,7 +91,9 @@ module Amber
 
     def start
       time = Time.now
-      str_host = "http://#{host}:#{port}".colorize(:light_cyan).underline
+      ssl_enabled = ssl_key_file && ssl_cert_file
```

src/amber.cr
```diff
@@ -98,6 +102,13 @@ module Amber
 
       server = HTTP::Server.new(host, port, handler)
 
+      if ssl_enabled
```
eliasjpr: maybe we an refactor all the SSL stuff into a method?

elorest: Agreed

c910335:
`key.pem` and `cert.pem` were generated by `openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -nodes` with all fields empty.

spec/amber/ssl/cert.pem
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
fridgerator: @elorest these are just for specs

c910335: @fridgerator absolutely

PR #157, opened 2017-08-02
Fixed CSRF. :)

elorest:
### Description of the Change
CSRF Works now. Took for ever to track down due to the fact that it worked sometimes and didn't other times. Thanks for the help @drujensen.


mixflame:
:+1: :gem:

PR #160, opened 2017-08-02
Update Amber::VERSION

faustinoaq:
### Description of the Change

Update Amber to latest version.

### Alternate Designs

Maybe we can automate this?

Because Amber::VERSION is shown when app starts:

```
I, [2017-08-02 11:30:00 -0500 #25107]  INFO -- : [Amber 0.1.4] serving application "App." at http://0.0.0.0:3000
I, [2017-08-02 11:30:00 -0500 #25107]  INFO -- : Server started in development.
I, [2017-08-02 11:30:00 -0500 #25107]  INFO -- : Startup Time 00:00:00.0005190
```

### Benefits

Correct information is shown.

### Possible Drawbacks

Need to update manualy every version.

### Applicable Issues

https://github.com/amber-crystal/amber/issues/159

### DIFF
--- a/src/amber/version.cr
+++ b/src/amber/version.cr
@@ -1,3 +1,3 @@
 module Amber
-  VERSION = "0.1.4"
+  VERSION = "0.1.6"
 end


elorest:
This actually is automated but if one of them is updated manually without the other then version.cr will fail because  it's doing a find and replace on current version from shard.yml

faustinoaq:
> actually is automated but if one of them is updated manually without the other then version.cr will fail 

So, Somebody changed it manually?

PR #161, opened 2017-08-03
[WIP] Ep/halt request lifecycle

eliasjpr:
### Applicable Issues

Issue https://github.com/amber-crystal/amber/issues/156

### Requirements

> As a developer, If redirect_to or render is called in before_action, the action should not run, but it does run currently. Redirect_to or render in action is fine since we can simply return.
However, if they can cancel the action in before_action, before_action can be more useful and easy to do lots of things like authentication.

### Description of the Change

- Adds a halt method to short circuit the request lifecycle causing the response to be written and close at the moment this method is called. Concept thanks to Kemal halt @sdogruyol https://github.com/kemalcr/kemal/blob/0b4856b7417592743a27fc3e30867fd141395bf6/src/kemal/helpers/macros.cr
- Halt method takes two arguments a status_code and an output.
- Controllers actions don't need to return strings
- The response now gets printed within the route handler see change (https://github.com/amber-crystal/amber/compare/ep/halt-request-lifecycle?expand=1#diff-d58c8f9ba048a01c31b2beb56b72840cR43 and https://github.com/amber-crystal/amber/compare/ep/halt-request-lifecycle?expand=1#diff-7ecdd88ee52d1f44596e2e8f5184df65R7)

### Alternate Designs

**Raising an exception**
Generally, the use of exceptions for control flow is an anti-pattern, with notable situation- and language-specific cough exceptions cough.

As a quick summary for why, generally, it's an anti-pattern:

- Exceptions are, in essence, sophisticated GOTO statements
- Programming with exceptions, therefore, leads to more difficult to read, and understand code
- Most languages have existing control structures designed to solve your problems without the use of exceptions
- Arguments for efficiency tend to be moot for modern compilers, which tend to optimize with the assumption that exceptions are not used for control flow.

Raising an exception involves allocating memory, and executing an exception handler is generally slow. (https://crystal-lang.org/docs/syntax_and_semantics/exception_handling.html bottom section) 

**Catch and Throw**
Unfortunately this is unsupported and difficult to implement according to @straight-shoota

Halting a request is a convenient way to not process more than necessary and allows for more request lifecycle control flow/options.

### Benefits

Allows to skip processing code when not necessary.

```crystal
  before_action do
    only :halt_request { halt_lifecycle }
  end

  def show
    "Action show result"
  end

  def halt_lifecycle
    redirect_to "/something"
    halt(status_code = 302) # for redirects to work
  end

  def halt_request
    "Should not be in request"
  end
end
```

### DIFF
--- a/src/amber.cr
+++ b/src/amber.cr
@@ -62,9 +62,9 @@ module Amber
       @pubsub_adapter = WebSockets::Adapters::MemoryAdapter
       @redis_url = "redis://localhost:6379"
       @session = {
-        :key       => "session_id",
+        :key => "session_id",
         # store can be [:signed_cookie, :encrypted_cookie, :redis]
-        :store     => :signed_cookie, 
+        :store     => :signed_cookie,
         :expires   => 0,
         :secret    => secret,
         :redis_url => "redis://localhost:6379",
--- a/src/amber/controller/base.cr
+++ b/src/amber/controller/base.cr
@@ -32,5 +32,12 @@ module Amber::Controller
     def session
       context.session
     end
+
+    def halt(status_code = 200, output = "")
+      context.response.headers["Content-Type"] = "text/plain"
+      context.response.status_code = status_code
+      context.response.print output
+      context.response.close
+    end
   end
 end
--- a/src/amber/dsl/router.cr
+++ b/src/amber/dsl/router.cr
@@ -16,7 +16,8 @@ module Amber::DSL
         content = controller.{{ action.id }}
         controller.run_after_filter(action)
         controller.run_after_filter(:all)
-        content
+        context.response.print(content)
+        context.response.close
       }
       %verb = {{verb.upcase.id.stringify}}
       %route = Amber::Route.new(%verb, {{resource}}, %handler, {{action}}, valve, scope, "{{controller.id}}")
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -211,11 +211,11 @@ module Amber::Controller
 
   describe "#halt" do
     it "raises and Amber::Exceptions::HaltRequest error" do
-        controller = build_controller("")
+      controller = build_controller("")
 
-        controller.halt(status_code = 800, output = "Message")
+      controller.halt(status_code = 800, output = "Message")
 
-        controller.response.status_code.should eq 800
+      controller.response.status_code.should eq 800
     end
   end
 end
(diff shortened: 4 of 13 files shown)


straight-shoota:
I didn't say throw/catch would be difficult to implement. But it is not part of the language and I don't expect it ever will. So it comes to the same... ;)

eliasjpr:
Closing this still not working as expected :(

src/amber/controller/base.cr
```diff
@@ -32,5 +32,9 @@ module Amber::Controller
     def session
       context.session
     end
+
+    def halt(status_code = 200, output = "")
+      raise Amber::Exceptions::HaltRequest.new(status_code, output)
```
elorest: I thought that throwing and exception to halt execution was a bad idea?

eliasjpr: Yes, I reverted back something was not working and was wondering why I removing this commit from the PR. Hence the WIP

src/amber.cr
```diff
@@ -62,9 +62,9 @@ module Amber
       @pubsub_adapter = WebSockets::Adapters::MemoryAdapter
       @redis_url = "redis://localhost:6379"
       @session = {
-        :key       => "session_id",
+        :key => "session_id",
```
elorest: Could we name this like "amber_session" or something?

src/amber/controller/base.cr
```diff
@@ -32,5 +32,12 @@ module Amber::Controller
     def session
       context.session
     end
+
+    def halt(status_code = 200, output = "")
+      context.response.headers["Content-Type"] = "text/plain"
+      context.response.status_code = status_code
+      context.response.print output
+      context.response.close
```
elorest: I think closing this will be an issue since pipes need to write to response on fallback.

src/amber/dsl/router.cr
```diff
@@ -16,7 +16,8 @@ module Amber::DSL
         content = controller.{{ action.id }}
         controller.run_after_filter(action)
         controller.run_after_filter(:all)
-        content
+        context.response.print(content)
+        context.response.close
```
eliasjpr: This is the last handler in the pipeline without it halting will not work

src/amber/controller/base.cr
eliasjpr: This is the purpose for halting the request

src/amber/dsl/router.cr
straight-shoota: You could probably signal "halting" in a different way, maybe a custom property on Response or Context...

eliasjpr:
Closing this for now since it needs some work and cleanup

PR #163, opened 2017-08-04
Remove CSRF dead code

veelenga:
### Description of the Change

This change is about to remove dead code in CSRF module. All that is actually set [here](https://github.com/amber-crystal/amber/blob/master/src/amber/router/pipe/error.cr#L11-L19).


veelenga:
@elorest sorry, already removed? On a branch somewhere? It is still in master [amber/router/pipe/csrf.cr#L18-L200](https://github.com/amber-crystal/amber/blob/master/src/amber/router/pipe/csrf.cr#L18-L20)

eliasjpr:
👍

PR #164, opened 2017-08-04
Fixes Headers for Transfer Encoding Chunked

eliasjpr:
### Requirements

> Headers are not currently being printed at the end of the response
when transfer encoding is set to chunked.

### Description of the Change

This attempts to correct the issue by ensuring that cookies are written
at the end of the pipeline

- It seems that we cannot use session.changed? since
context.cookies.write(context.response.headers) writes not only the
session but also any other cookies

- context.session.set_session has to be called before call_next(context)
since it prepares the session scope. If done after none of the pipes
down the drain will have a session and no session will be ever set.

screen shot 2017-08-04 at 3 48 38 pm

### DIFF
--- a/src/amber/router/pipe/session.cr
+++ b/src/amber/router/pipe/session.cr
@@ -5,14 +5,9 @@ module Amber
     # be used to maintain data across requests.
     class Session < Base
       def call(context : HTTP::Server::Context)
-        # Session has to be set before it can be use down the pipeline
-        context.session.set_session
-
         call_next(context)
-      ensure
-        if context.session.changed?
-          context.cookies.write(context.response.headers)
-        end
+        # Writes the session to the store 
+        context.session.set_session
       end
     end
   end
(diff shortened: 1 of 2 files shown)


src/amber/router/pipe/session.cr
```diff
@@ -5,10 +5,12 @@ module Amber
...
-        call_next(context)
+        # Session has to be set before it can be use down the pipeline
+        context.session.set_session
 
+        call_next(context)
+      ensure
         if context.session.changed?
-          context.session.set_session
           context.cookies.write(context.response.headers)
         end
```
eliasjpr: Left this here but is not really needed since we always have to write to response headers regardless if the session has changed or not.

elorest: It's duplicate at this point and doesn't save any time :(

eliasjpr:
screen shot 2017-08-04 at 7 04 42 pm
screen shot 2017-08-04 at 7 05 30 pm

PR #165, opened 2017-08-04
Don't merge. Proof of concept. Fixes header issue in chunking but removes the option to not use session. :(

elorest:
Shouldn't be merged. Fixes chunking by writing headers before setting response. which is required in chunked mode. Bad fix in that session is used and written regardless of if session pipe is enabled. Should be refactored by for now it solves all the usability issues.

### DIFF
--- a/src/amber/router/pipe/pipeline.cr
+++ b/src/amber/router/pipe/pipeline.cr
@@ -45,6 +45,20 @@ module Amber
         end
       end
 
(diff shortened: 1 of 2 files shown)


elorest:
Yeah but that would be a bad idea. Like I said this isn't a good fix. But it does solve the problem we were having last night and csrf works.

PR #168, opened 2017-08-08
Halt Request Lifecycle

eliasjpr:
Issue: https://github.com/amber-crystal/amber/issues/156

should not run, but it does run currently. redirect_to or render
in action is fine since we can simply return. However, if they
can cancel the action in before_action, before_action can be
more useful and easy to do lots of things like authentication.

- Exits request action lifecycle by returning false from anywhere
in the controller including redirect_to, before/after filters and
actions.

- As soon the request Lifecycle is halted it will begin to unfold
the response.

- Alternatively you can use `return halt(status_code, content)` to
halt the request, this method sets status_code and content for the
response, and returns false.

### Description of the Change

- Introduces *halt!* controller method which cancels a request lifecycle
- *redirect_to* now halts the request

### Alternate Designs

```crystal

  before_action do
    only :halt_action { halt!(900) }
  end

  def halt_action
    raise "Should not reach this action"
    ""
  end

end

class FakeRedirectController < Amber::Controller::Base

  def redirect_action
    redirect_to "/"
    ""
  end
end
```
### Benefits

Avoids extra processing when calling *halt!*

### DIFF
--- a/src/amber/controller/base.cr
+++ b/src/amber/controller/base.cr
@@ -34,6 +34,7 @@ module Amber::Controller
     end
 
     def halt!(status_code : Int32 = 200, content = "")
+      response.headers["Content-Type"] = "text/plain"
       response.status_code = status_code
       response.print content
       response.close
--- a/src/amber/controller/base.cr
+++ b/src/amber/controller/base.cr
@@ -36,9 +36,7 @@ module Amber::Controller
     def halt!(status_code : Int32 = 200, content = "")
       response.headers["Content-Type"] = "text/plain"
       response.status_code = status_code
-      response.print content
-      response.close
-      context.halt = true
+      context.content = content
     end
   end
 end
--- a/src/amber/controller/base.cr
+++ b/src/amber/controller/base.cr
@@ -32,5 +32,12 @@ module Amber::Controller
     def session
       context.session
     end
+
+    def halt!(status_code : Int32 = 200, content = "")
+      response.status_code = status_code
+      response.print content
+      response.close
+      context.halt = true
+    end
   end
 end
--- a/src/amber/dsl/router.cr
+++ b/src/amber/dsl/router.cr
@@ -12,9 +12,9 @@ module Amber::DSL
       %handler = ->(context : HTTP::Server::Context, action : Symbol){
         controller = {{controller.id}}.new(context)
         controller.run_before_filter(action) unless context.halt
-        content = controller.{{ action.id }} unless context.halt
+        content = context.halt ? nil : controller.{{ action.id }}
         controller.run_after_filter(action) unless context.halt
-        content || ""
+        content.to_s
       }
       %verb = {{verb.upcase.id.stringify}}
       %route = Amber::Route.new(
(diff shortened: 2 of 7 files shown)


src/amber/dsl/router.cr
```diff
@@ -11,13 +11,15 @@ module Amber::DSL
...
       %handler = ->(context : HTTP::Server::Context, action : Symbol){
         controller = {{controller.id}}.new(context)
-        controller.run_before_filter(action)
-        content = controller.{{ action.id }}
-        controller.run_after_filter(action)
-        content
+        controller.run_before_filter(action) unless context.halt
+        content = controller.{{ action.id }} unless context.halt
+        controller.run_after_filter(action) unless context.halt
+        content || ""
```
elorest: Maybe we should do `content.to_s` instead.

src/amber/dsl/router.cr
```diff
@@ -11,13 +11,15 @@ module Amber::DSL
     macro route(verb, resource, controller, action)
       %handler = ->(context : HTTP::Server::Context, action : Symbol){
         controller = {{controller.id}}.new(context)
-        controller.run_before_filter(action)
-        content = controller.{{ action.id }}
-        controller.run_after_filter(action)
-        content
+        controller.run_before_filter(action) unless context.halt
```
elorest: Is there any reason to say unless `context.halt` on the first call in the chain?

src/amber/controller/base.cr
```diff
@@ -32,5 +32,12 @@ module Amber::Controller
     def session
       context.session
     end
+
+    def halt!(status_code : Int32 = 200, content = "")
```
elorest: If we're going to use this method to set status_code and content shouldn't it do headers as well?

src/amber/dsl/router.cr
eliasjpr: Developers can write a handler that halts the request before reaching the controller.

src/amber/controller/base.cr
eliasjpr: Im cleaning this up in a PR for including the flash in redirect.

eliasjpr:
@elorest updated the halt! method to write to headers also added `content.to_s`

src/amber/controller/base.cr
elorest: Sweet.

src/amber/dsl/router.cr
elorest: Handlers already have the ability to completely halt the request cycle but returning instead of calling next. Halt is only necessary because it all happens within the same handler proc.

src/amber/controller/base.cr
```diff
@@ -32,5 +32,13 @@ module Amber::Controller
...
+      response.headers["Content-Type"] = "text/plain"
+      response.status_code = status_code
+      response.print content
+      response.close
+      context.halt = true
```
elorest: I think it would make more sense to set `context.content = content` inside of halt and still have the pipeline handler print it.

Then instead of checking `context.halt` we could just check `context.content`. This is cleaner and won't try to print to response twice.

eliasjpr: yes I will change this I like that

PR #172, opened 2017-08-12
Adds Flash to Current Redirect Implementation

eliasjpr:
## Adds Flash to Current Redirect Implementation

**Issue:** https://github.com/amber-crystal/amber/issues/76

* This adds flash to the redirect_to method
* Arguments should be passed as named params
* status, params and flash as optional arguments
* Adds `controller_name` method to controller
* Adds `merge!` to **FlashStore** take a `Hash(String, String)?` as an argument
 
```crystal
redirect_to(:world, status: 302, flash: {"notice" => "Success!"})
redirect_to(:action, status: 302, params: { "some" => "value"}, flash: {"notice" => "Success!"})
redirect_to(:controller, :action, status: 302, params: { "some" => "value"}, flash: {"notice" => "Success!"})
```

### DIFF
--- a/src/amber/router/pipe/flash.cr
+++ b/src/amber/router/pipe/flash.cr
@@ -106,11 +106,9 @@ module Amber
           self
         end
 
-        def merge!(other : Hash(String, String)?)
-          if !other.nil?
-            other.each do |k, v|
-              @flashes[k.to_s] = v
-            end
+        def merge!(other : Hash(String, String))
+          other.each do |k, v|
+            @flashes[k.to_s] = v
           end
           self
         end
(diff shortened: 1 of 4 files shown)


src/amber/router/pipe/flash.cr
```diff
@@ -106,6 +106,15 @@ module Amber
           self
         end
 
+        def merge!(other : Hash(String, String)?)
```
elorest: Is there any reason to allow nil for the input value? It seems as if merge!(nil) should just exception and not compile.

eliasjpr: yes, I clean up. Since the flash hash could be nil to avoid the check on the other side. Should be cleaned now

elorest: Ok good enough for now then.

PR #174, opened 2017-08-15
Fixes Redirect When Controller Action is passed

eliasjpr:
Issue: https://github.com/amber-crystal/amber/issues/169
Issue:  https://github.com/amber-crystal/amber/issues/77

Redirect_to is rather limited in quite a few cases. Most resourceful routes will fail.

### Description of the Change

* Looks in the routes for the correct controller action resource path and if given will parse the path params.

```crystal
# config/routes.cr
get "/users/:id", UsersController, :show

# Within controller
redirect_to(controller: users, action: show, params: { "id" => "1", "utm" => "value" })

# Redirects to
"/users/1?utm=value"
```

This adds a match_by_controller_action(controller, action) method to Amber::Router::Router and performs a lookup when a match is found it returns the route. Then a `parse_params(params)` call can be made on the router to replace the resource params with actual values.

### DIFF
--- a/spec/amber/controller/redirect_spec.cr
+++ b/spec/amber/controller/redirect_spec.cr
@@ -66,28 +66,45 @@ module Amber::Controller
     end
 
     describe ".from_controller_action" do
-      Spec.before_each do
+      context "with params" do
+        it "redirects to correct location for given controller action" do
+          router = Amber::Router::Router.instance
+          router.draw :web do
+            get "/fake/:id", HelloController, :show
+            put "/hello/:id", HelloController, :update
+          end
+          controller = build_controller
+          redirector = Redirector.from_controller_action("hello", :show, params: {"id" => "11"})
+
+          redirector.redirect(controller)
+
+          controller.response.headers["location"].should eq "/fake/11"
+          controller.response.status_code.should eq 302
+          controller.context.content.nil?.should eq false
+        end
       end
 
-      it "redirects to correct location for given controller action" do
-        router = Amber::Router::Router.instance
-        router.draw :web do
-          get "/fake", HelloController, :index
-          get "/fake/:id", HelloController, :show
-          put "/hello/:id", HelloController, :update
+      context "without patams" do
+        it "redirects to correct location for given controller action" do
+          router = Amber::Router::Router.instance
+          router.draw :web do
+            get "/fake", HelloController, :index
+            put "/hello/:id", HelloController, :update
+          end
+          controller = build_controller
+          redirector = Redirector.from_controller_action("hello", :index)
+
+          redirector.redirect(controller)
+
+          controller.response.headers["location"].should eq "/fake"
+          controller.response.status_code.should eq 302
+          controller.context.content.nil?.should eq false
         end
-        controller = build_controller
-        redirector = Redirector.from_controller_action("hello", :show, params: {"id" => "11"})
-
-        redirector.redirect(controller)
-
-        controller.response.headers["location"].should eq "/fake/11"
-        controller.response.status_code.should eq 302
-        controller.context.content.nil?.should eq false
       end
     end
+
     describe "#redirect_back" do
-      context "and has a valid referer" do
+      context "and has a valid referrer" do
         it "sets the correct response headers" do
           hello_controller = build_controller("/world")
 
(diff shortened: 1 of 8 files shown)


spec/amber/controller/redirect_spec.cr
```diff
@@ -0,0 +1,112 @@
...
+
+      it "redirects to correct location for given controller action" do
+        router = Amber::Router::Router.instance
+        router.draw :web do
+          get "/fake", HelloController, :index
+          get "/fake/:id", HelloController, :show
+          put "/hello/:id", HelloController, :update
+        end
+        controller = build_controller
+        redirector = Redirector.from_controller_action("hello", :show, params: {"id" => "11"})
```
drujensen: Add a test without params like `Redirector.from_controller_action("hello", :index)`, it looks like params are necessary looking at the method.

src/amber/controller/redirect.cr
```diff
@@ -1,61 +1,68 @@
...
     @status : Int32 = 302
     @params : Hash(String, String)? = nil
     @flash : Hash(String, String)? = nil
 
-    def initialize(@location, @status = 302, @params = nil, @flash = nil)
-      raise_redirect_error(location) if location.empty?
+    def self.from_controller_action(controller : String, action : Symbol, **options)
+      router = Amber::Router::Router.instance
+      route = router.match_by_controller_action(controller, action)
+      if params = options[:params]?
```
drujensen: What if there are no params like an index page?

src/amber/controller/redirect.cr
```diff
@@ -1,61 +1,68 @@
...
-      response.headers.add "Location", url_path
-      response.status_code = status
+      controller.response.headers.add "Location", url_path
+      controller.halt!(status, "Redirecting to #{url_path}")
     end
 
-    def encode_query_string(location, params)
-      return location + "?" + HTTP::Params.encode(params).to_s if params
+    private def encode_query_string(location, params)
+      if !params.nil? && !params.empty?
```
drujensen: what do you think about using `unless` to avoid the negative?

src/amber/controller/redirect.cr
```diff
@@ -1,61 +1,68 @@
...
-    def redirect_to(action : Symbol, **args)
-      LocationRedirect.new("/#{controller_name}/#{action}", **args).redirect(context)
-      halt!(response.status_code, "Redirecting to #{response.headers["Location"]}")
-    end
-
-    def redirect_back(**args)
-      LocationRedirect.new(request.headers["Referer"], status: 302).redirect(context)
-      halt!(response.status_code, "Redirecting to #{response.headers["Location"]}")
+    private def set_flash(controller)
+      controller.flash.merge!(flash.not_nil!) if !flash.nil?
```
drujensen: unless flash.nil?

src/amber/router/route.cr
```diff
@@ -35,5 +35,20 @@ module Amber
     def call(context)
       handler.call(context)
     end
+
+    def parse_params(params)
```
drujensen: This feels more like a `generate_path` or `substitute_keys_in_path` method

elorest: Yeah I agree. `parse_params` sounds too much like something that happens earlier on.

src/amber/controller/redirect.cr
elorest: Then the if statement will be false right?

spec/amber/controller/redirect_spec.cr
eliasjpr: A very much missing test

src/amber/controller/redirect.cr
eliasjpr: Removed the if statement is not necessary

src/amber/controller/redirect.cr
eliasjpr: Unless for some reason raises a compile error on `params.empty?` **Undefined method 'empty?' for Nil**, forcing me to rewrite it aas `params.to_s.empty?`.

src/amber/router/route.cr
eliasjpr: I liked `substitute_keys_in_path`, thanks I really couldn't think of a better name at the time.

eliasjpr:
@drujensen I have addressed the comments. Please note that I had some issues with using unless I added a comment.

src/amber/controller/redirect.cr
drujensen: ah, ok

PR #175, opened 2017-08-15
channel.leave firing wrong event on leave

### LINKED ISSUE #155: Websockets handle_leave doesn't work
Try leaving a channel with .leave() in JavaScript and have handle_leave ready in Crystal. It won't be called.

fridgerator:
fixes #155 

### Description of the Change

channel.js `leave` method was calling the wrong even type.  Easy fix.


mixflame:
gj

PR #177, opened 2017-08-16
Adds Optional to Params Validation

eliasjpr:
> As a developer, I want to be able to define optional parameters and perform
validations only if the param is present.

Issue: https://github.com/amber-crystal/amber/issues/143

Acceptance Criteria:
*  I should be able to define an optional parameter as follow

```crystal
params do
  optional(:email) { |p| p.email? & p.size.between? 1..10 }
end
```

Optional params only performs validations when is present, otherwise will be skipped.


spec/amber/validations/params_spec.cr
```diff
@@ -12,8 +12,40 @@ module Amber
...
         end
 
-        validator.errors.size.should eq 0
         validator.valid?.should be_true
+        validator.errors.size.should eq 0
+      end
+
+      context "optional params" do
+        context "when missing" do
+          it "does not validate optional field" do
```
drujensen: As a rule, each context should have at least 2 tests; one positive and one negative.  Can you add a test for when `validator.valid?.should be_true`?

spec/amber/validations/params_spec.cr
```diff
@@ -12,8 +12,40 @@ module Amber
...
+              optional(:name) { |v| v.empty? }
+              required("last_name") { |v| v.empty? }
+            end
+
+            validator.valid?.should be_false
+            validator.errors.size.should eq 1
+          end
+        end
+
+        context "when present" do
```
drujensen: maybe this is your positive test? 

either merge this with the context above or add another test in this context for when `validator.valid?.should be_true`

spec/amber/validations/params_spec.cr
eliasjpr: Read next comment.

spec/amber/validations/params_spec.cr
eliasjpr: Yes this is my positive test and these tests are not testing the `#valid?` method, which has positive and negative tests. This tests are more like `integration tests` for the ValidationBuilder. Ideally we want to unit tests the `ValidationBuilder`, `OptionalRule` and `BaseRule`. I think I will do just that add unit tests  for each of the classes

drujensen: gotcha. looks good.

PR #179, opened 2017-08-16
passing client_socket to handle_message

fridgerator:
### Description of the Change

The channel `handle_message` now passes a (nilable) client socket parameter.  Because of this change, when used in a multiple server scenario (redis adapter) the client socket id had to be changed to a unique id (uuid), because obejct_id may not be unique across server instances.

**BREAKING CHANGE -** `handle_message` now takes 2 parameters instead of one.


drujensen:
Thanks @fridgerator

PR #183, opened 2017-08-21
Clean Up and Removed Code from amber.cr

eliasjpr:
Amber.cr needed tlc and for that reason proceeded to do some cleanup
and removed unused code from the file.

After the changes now the initializer methods is 1 line small,
properties are set with their corresponding default value and added
some ENV variables by default.

Added `Amber::Server.config` witch takes a block. Removes the need for `Amber::Server.instance.config`
```crystal

Amber::Server.config  do |amber|
  amber.name = "ProjectName"
  amber.port = 8080
  amber.host = "localhost"
end
```

### DIFF
--- a/src/amber.cr
+++ b/src/amber.cr
@@ -10,6 +10,32 @@ require "./amber/**"
 
 module Amber
   class Server
+    alias WebSocketAdapter = WebSockets::Adapters::RedisAdapter.class | WebSockets::Adapters::MemoryAdapter.class
+
+    property port_reuse = true
+    property port = 8080
+    property name = "amber_project"
+    property env = ARGV[0]? || ENV["AMBER_ENV"]? || "development"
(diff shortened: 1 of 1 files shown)


eliasjpr:
Ping!

src/amber.cr
```diff
@@ -18,8 +44,8 @@ module Amber
       instance.all_routes
     end
 
-    def self.settings
-      instance
+    def self.config(&block)
```
drujensen: 🥇 I really like this.  Should we should consider this pattern for the pipes as well?  I had something similar in Kemalyst.

eliasjpr: So people can configure pipelines like?

```crystal
Amber::Server.pipeline :web do
  plug Amber::Plug::Session.new
end
```

PR #184, opened 2017-08-24
Adds useful methods to Context

eliasjpr:
Issue: https://github.com/amber-crystal/amber/issues/180

The request object contains a lot of useful information about the
request coming in from the client. To get a full list of the available
methods, refer to the Crystal API Documentation
https://crystal-lang.org/api/0.23.0/HTTP/Request.html

This adds several useful methods

* format to determine The content type requested by the client.
(JSON, XML, HTML etc)
* get?, post?, patch?, put?, delete?, head? Returns true if the
HTTP method is GET/POST/PATCH/PUT/DELETE/HEAD.
* port The port number (integer) used for the request.
* requested_url The entire URL used for the request.
* client_ip The IP Address of the client. This does not attempt to get
the actual machine remote IP Address but rather relies on request headers
to obtain it

### DIFF
--- a/src/amber/support/mime_types.cr
+++ b/src/amber/support/mime_types.cr
@@ -197,7 +197,6 @@ module Amber
         "hqx"       => "application/mac-binhex40",
         "htc"       => "text/x-component",
         "htke"      => "application/vnd.kenameaapp",
-        "htm"       => "text/html",
         "html"      => "text/html",
         "hvd"       => "application/vnd.yamaha.hv-dic",
         "hvp"       => "application/vnd.yamaha.hv-voice",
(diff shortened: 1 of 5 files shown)


src/amber/support/mime_types.cr
```diff
@@ -629,6 +628,10 @@ module Amber
       def self.zip_types(path) # https://github.com/h5bp/server-configs-nginx/blob/master/nginx.conf
         [".htm", ".html", ".txt", ".css", ".js", ".svg", ".json", ".xml", ".otf", ".ttf", ".woff", ".woff2"].includes? File.extname(path)
       end
+
+	  def self.format(accepts)
+		MIME_TYPES.key?(accepts)
```
jasdeepsingh: I see some formatting inconsistency here?

jasdeepsingh: Indentation*

eliasjpr:
Ping! This is still WIP. 

The controller was basically using the same methods from the context
class with no real differences. In order to clean up the controller
all methods related to the context are now delegated to the Context
class. This allowed to clean up the Base controller class and specs

Note:

It would be nice to move the halt! method to the context, left this
out from this commit since it involves more changes that could affect
the request lifecycle.

PR #186, opened 2017-08-25
provide auth generation

drujensen:
This change provides the ability to generate an authenticated resource.

Unfortunately, there is some templating that is duplicated with minor modifications. We may want to look at a way to sharing templates between generators.

To test:

amber new project --deps
cd project
amber g auth User name:string
crystal db/seeds.cr
This will generate a User model with scaffolding.
It forces generation of email:string and encrypted_password:string fields.
It adds the SessionController with login/create/logout methods with routes.
It adds a new middleware/Authenticate to the pipeline.
It adds a db/seeds.cr that will create a new user.
It adds a new field type password for correct generation of the forms.

### DIFF
--- a/src/templates/app/src/{{name}}.cr
+++ b/src/templates/app/src/{{name}}.cr
@@ -2,6 +2,9 @@ require "amber"
 require "../config/*"
 require "./models/**"
 require "./mailers/**"
+require "./middleware/**"
+
+# load the application_controller before controllers which depend on it
 require "./controllers/application_controller"
 require "./controllers/**"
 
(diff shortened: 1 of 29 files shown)


src/templates/app/src/{{name}}.cr
```diff
@@ -2,6 +2,9 @@ require "amber"
 require "../config/*"
 require "./models/**"
 require "./mailers/**"
+require "./middleware/**"
```
eliasjpr: can we rename this to `handlers` instead since is more the nomenclature that Crystal uses?

drujensen: yeah, sounds good.

PR #191, opened 2017-08-26
npm, webpack, babel integration for assets compilation

drujensen:
### Description of the Change

This provides an asset pipeline for Amber using npm and webpack.

- all assets are placed in `src/assets`
- support for loading images, fonts, css, javascript
- support for sass transpiler
- support for es2015 transpiler using babel
- includes `lib/amber/assets/amber.js` for web socket support
- different config for dev and production.  
   - Dev provides inline source mapping
   - Prod provides uglify compression

In order to try this out, you will need to open a new terminal and run:
```
npm install

// build dev bundle
npm run build

// watch for changes and build dev bundle
npm run watch

// build prod bundle
npm run release
```

### Alternate Designs

There are many alternatives to this approach.  We could look into a pure crystal solution or use other packaging solutions like Gulp.

npm, webpack and babel are fairly standard in the JS community.  By providing a standard set of configuration settings simplifies the developers life so they don't have to deal with the complexities of choosing.

This provides the minimal amount of npm packages necessary.  This does not include any packages that would be controversial and everything included is recommended by the webpack guidelines.

### Benefits

All assets are compiled into one bundle.  Ability to load images or fonts directly into css.  Ability to write sass.  Ability to write es2015.  Ability to compile and minify for production.

### Possible Drawbacks

This change will require users to install node and npm.  Maintaining npm packages and supporting node/npm/webpack/babel will take valuable time from the core team.

### Applicable Issues

The initial install will provide pre-compiled assets bundle but any changes to css/js will require the user to install/learn node and npm to recompile the assets bundle.


src/templates/app/shard.yml.ecr
```diff
@@ -6,7 +6,7 @@ authors:
 
 crystal: <%= Crystal::VERSION %>
 
-license: MIT
+license: UNLICENSED
```
elorest: Rather than UNLICENSED wouldn't it be better off just to delete the field?

src/templates/app/config/webpack/dev.js
```diff
@@ -0,0 +1,6 @@
+const merge = require('webpack-merge');
+const common = require('./common.js');
```
elorest: We'll probably want to coordinate this with crystal environments as well. Maybe call them development and production unless you think we want to shorten them for the whole project.

src/templates/app/shard.yml.ecr
drujensen: npm requires a license field and they recommend using "UNLICENSED".  I think for consistency sake we should keep this in.

src/templates/app/config/webpack/dev.js
drujensen: good call.  renamed to `development` and `production`

drujensen:
@eliasjpr is this approved?

eliasjpr:
Approved! 👍🏻

eliasjpr:
As an idea, maybe we can add an option to generate app with front end support. I personally incline to have any frontend related code in the public folder node modules, js, CSS and not wihin the src directory

drujensen:
@eliasjpr for traditional request/response apps, I think the `assets` should be in `src` .  This is the same as rails and feels right to me.

For SPA single page apps, their should be a separation between front and back end and I agree in that case that assets should not be in `src`. Either a `client` directory or a separate repo altogether.   Also, they would be structured differently where assets would be managed with the components instead of a globally maintained folders like `stylesheets`. 

What if we specify when they create a new application if they want to build an SPA `--spa`?  Then  we could change the generators to build APIs and front end components in a `client` directory instead?

faustinoaq:
SPA support would be awesome! :smile:

PR #199, opened 2017-08-27
Small controller generator change.

elorest:
### Description of the Change

Cleaned up controller generator a tiny bit.

### DIFF
--- a/src/amber/cmd/templates/controller/src/controllers/{{name}}_controller.cr.ecr
+++ b/src/amber/cmd/templates/controller/src/controllers/{{name}}_controller.cr.ecr
@@ -1,8 +1,12 @@
 class <%= @name.capitalize %>Controller < ApplicationController 
-<% @actions.each do |action, verb| -%>
-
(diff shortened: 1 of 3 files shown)


src/amber/cmd/templates/controller/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -1,8 +1,12 @@
 class <%= @name.capitalize %>Controller < ApplicationController 
-<% @actions.each do |action, verb| -%>
-
-  def <%= action %>
-    render("<%= action %>.<%= @language %>")
+<%= 
+  actions = @actions.map do |action, verb| 
+    <<-ACTION
```
eliasjpr: I kinda like how it was but all good :)

elorest: Lol. Yeah the template definitely looked better before. The generated code looks better now though and I think that's more important.

PR #204, opened 2017-08-28
[WIP] Remove amber directories from spec and src

eliasjpr:
### Description

Moves the **cmd** directory outside of the amber directory and renames the **core** directory to Amber.

```bash
-- spec
  -- amber
  -- cmd
  -- support
  spec_helper.cr
  docker-compose.yml
  Dockerfile

-- src
  -- amber
  -- cmd
  amber.cr
  amber_cmd.cr
  version.cr
```
### Why this change?

It turns out that modules namespaces conventions should follow a directory structure where/when possible to keep the project cohesive and to understand where things are. For instance `Amber::Controller` should point to `amber/controller` directory. 

Currently there is a mismatched directory to module structure because of the core directory we have `amber/core/controller` but the module is still `Amber::Controller` I tried/consider renaming the modules to include core, but it was just not right and the module namespacing became unnecessarily too long eg `Amber::Core::Controller`. For this reason, I think is cleaner to keep **amber** instead of **core**.

### Alternatives
As an alternative we can have the following directory structure
```
--src
   --amber
      -- controller
      -- cmd # Command Line files
      -- dsl
      ...
```

### Changes
- Removes the extra amber directory from **src** and **spec** directory.
- Fixes the specs to run
- Moves the version.cr file to the root of the src directory.
- Make still works as expected
- Moves the shared spec support directory to the root of spec


eliasjpr:
This is actually a reflection of that I would still like to hear from @drujensen and @fridgerator

fridgerator:
Makes sense to me

drujensen:
Elias, my first comment to Isaac was asking why do we need the extra amber directory? I was thinking just like you did.  It seemed unnecessary. 

The reason is name spacing conventions where you should have a directory for each namespace. 

I think we should keep the extra directory to stay with that convention. What do you think?

eliasjpr:
So you would like to stay with  `Amber/Core` even if we do not have any modules namespace with Core? If we decide I will advocate for us to namespace all the modules with Amber::Core IMHO.

drujensen:
Your right. We either add the namespace to match the core and cmd directories or remove them.  I forgot that we don't have the core namespace yet.

eliasjpr:
Spoke to @drujensen about the PR IRL and came up with this directory structure

screen shot 2017-08-28 at 9 42 52 pm
screen shot 2017-08-28 at 9 49 19 pm

eliasjpr:
Updates:

- Renames CMD to CLI
- Namespaced Amber::CMD to Amber::CLI
- Cleans CLI specs
- Addss Server directory for server related classes/modules 

- Source directory looks is now
```crystal
-- src
  -- amber
  -- cli
  amber.cr
```

eliasjpr:
Ok took care of the directory. This should be good to go after tests passes

drujensen:
whew, that was close.  almost didn't get that last bug fix in.  should have rejected it when I found more issue.

PR #207, opened 2017-08-30
add `npm run watch` to docker-compose.yml

drujensen:
### Description of the Change

- Add node.js and npm to the docker image.  
- Create a new docker container that runs `npm install && npm run watch`. Any changes to the assets will automatically recompile.

### Alternate Designs

We could look at adding this capability to the `amber watch` with a configuration file similar to foreman.

Docker and Docker Compose provide easily configurable containers to setup a development environment.  This change adds another container to handle the new asset management via npm and web pack.

### Benefits

This allows any developer who has docker installed to run `docker-compose up -d` and they will have a running development environment with all the needed services.

### Possible Drawbacks

Docker has a fairly steep learning curve and is intimidating for new users.  We already provide Docker and Docker Compose configuration files so adding another container should not be much of an issue.

### Applicable Issues

None that I'm aware of


drujensen:
Lol

PR #209, opened 2017-08-31
Refactors Amber::Server

eliasjpr:
### Description of the Change

Some of the calls to the server instance was invoking multiple methods
for example Amber::Server.instance.pubsub_adapter.instance, with the
changes this now reads as Amber::Server.pubsub_adapter making it easy to
change without reworking the caller.

### Requirements

* All new code requires tests to ensure against regressions

### Alternate Designs

We could dynamically generate the class methods for each of the properties to reduce the number of methods define at code level

In other to make the framework attractive to developer, it should readable and accessible.

### Benefits

The resulting changes tends to be more maintainable and adaptable making
the code more readable to the developer.

### Possible Drawbacks

Narrow interfaces, giving access to only as much information as it needs to do
its job, as each method needs to know about a small set of methods of closely related objects

High number of class methods


eliasjpr:
@elorest @drujensen  I had to add a test_shard.yml to run tests for the CLI, basically, this will look for the current amber context that the tests are running.

I am open to suggestions

spec/support/sample/test_shard.yml
```diff
@@ -0,0 +1,43 @@
...
+  amber:
+    main: lib/amber/src/amber/cli.cr
+
+development_dependencies:
+  amber_spec:
+    github: amber-crystal/amber-spec
+
+dependencies:
+  amber:
+    path: ../../amber
```
eliasjpr: This test_shard.yml file is used to run tests in the current amber context. 

With this file in place, we can replace the generated **shard.yml** for the `test_app`, causing the tests to look at the current amber context (very useful for testing changes to the framework to make sure the CLI does not break).

spec/amber/cli/commands_spec.cr
```diff
@@ -38,6 +38,8 @@ module Amber::CLI
         MainCommand.run ["new", TESTING_APP, "--deps"]
         Dir.cd(TESTING_APP)
         MainCommand.run ["generate", "scaffold", "Animal", "name:string"]
+		`rm ./shard.yml`
```
eliasjpr: Replace the generated shard.yml with a test shard to test the current amber context.

eliasjpr:
This adds a test_shard.yml file is used to run tests in the current amber context.

With this file in place, we can replace the generated shard.yml for the test_app, causing the tests to look at the current amber context (very useful for testing changes to the framework to make sure the CLI does not break).

For example https://github.com/amber-crystal/amber/pull/209/files#diff-e3abc1c7c34e94cf42d32a206fe06768R41

spec/amber/cli/commands_spec.cr
drujensen: @eliasjpr is there a way to just add the `path: ../amber` to the shard instead of replacing the whole file?  I'm concerned that we will forget to update this second shard.yml and it will get out of date.

spec/amber/router/pipe/session_spec.cr
```diff
@@ -30,7 +30,7 @@ module Amber
 
         context "Redis Store" do
           it "sets session value in controller" do
-            Amber::Server.instance.session = {
+            Amber::Server.settings.session = {
```
drujensen: I really like this. ✨

spec/amber/cli/commands_spec.cr
eliasjpr: Use the `sed` command instead to replace just the line. Note that this is looking by line number.

eliasjpr:
@drujensen made some changes to just replace the line of the shard.yml. Note that this is replacing by line number.

elorest:
Could you just gsub instead?

PR #219, opened 2017-09-02
generate secret one time at creation

drujensen:
### Description of the Change

Configure and generate a secret for the cookie session store. Currently the default "secret" is not secure and could lead to vulnerabilities.

### Alternate Designs

n/a

### Benefits

The user no longer needs to do this manually.

### Possible Drawbacks

n/a

### DIFF
--- a/src/amber/cli/templates/app/config/application.cr.ecr
+++ b/src/amber/cli/templates/app/config/application.cr.ecr
@@ -7,4 +7,11 @@ Amber::Server.configure do |setting|
   setting.env = (ENV["AMBER_ENV"] ||= "development").to_s
   setting.log = ::Logger.new(STDOUT)
   setting.log.level = ::Logger::INFO
+
+  setting.session = {
+    :key       => "<%= @name %>.session",
+    :store     => :signed_cookie,
+    :expires   => 0,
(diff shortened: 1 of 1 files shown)


drujensen:
just checked out @elorest changes and he takes care of this.  closing.

drujensen:
Re-opening this.  We should do it this way instead.

src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -7,4 +7,11 @@ Amber::Server.configure do |setting|
   setting.env = (ENV["AMBER_ENV"] ||= "development").to_s
   setting.log = ::Logger.new(STDOUT)
   setting.log.level = ::Logger::INFO
+
+  setting.session = {
+    :key       => "<%= @name %>.session",
+    :store     => :signed_cookie,
+    :expires   => 0,
+    :secret    => "<%= SecureRandom.hex(64) %>"
```
elorest: This doesn't actually fix the issue because this secret doesn't actually get used by the current implementation. It gets passed into session::cookie_store but is abandoned there. New keys are then generated by signed cookie and encrypted_cookie which change every time it restarts. I still have a decent amount of work to make it use my new secret_key_base and I'd love to pair.

PR #221, opened 2017-09-02
[WIP] Url Helper for generating full urls

eliasjpr:
### Description of the Change

Returns a URL that has been rewritten according to the named
tuple options.

Url Options:

- protocol - Overrides the default (current) protocol if provided.
- port - Optionally specify the port to connect to.
- host - Overrides the default (current) host if provided.
- only_path - If true, returns the relative URL (omitting the protocol,
- host name, and port) (false by default).
- trailing_slash - If true, adds a trailing slash, as in "/archive/2005/".
Note that this is currently not recommended since it breaks caching.
- anchor - Appends anchor to the end of the URL

### Alternate Designs

- Would be nice to have just one method to define all routes.

### Benefits

- This will allow support to define routes helper methods.
- TODO: Make accessible from your controllers, views, and mailers


eliasjpr:
Yes, this will be tied into the router and used for the generated routes helper method. Something like 

```crystal
macro route_helper(route)
  def {{route.name}}_path(**options)
      # Some logic...
      Url.new(**options)
  end 
end
```

elorest:
I'm concerned that as this currently stands we're just taking a simple task like `link_to "/listings/1"` or `link_to "https://www.google.com"` and adding a bunch of processing and extra syntax too it. When we tie this into the router I think it will be useful.

eliasjpr:
@drujensen this class it to be used to generate routes helper methods.

Eg user_path, posts_path, category_url.

This class does not intend to be used for link_to although it can be used.

eliasjpr:
@drujensen you're right, I should be using the URI class instead, I'm going to rework this to do just that. I'm going to rework

PR #224, opened 2017-09-03
Dj/websocket route fix

drujensen:
### Description of the Change

Add a macro to allow users to easily configure a websocket route.
```
route :web do
  websocket "/user", UserSocket
  websocket "/room", RoomSocket
end
```

### Alternate Designs

Create a separate pipeline for web sockets.

### Benefits

Simplifies the use of web sockets.


mixflame:
I can test it, but I can't merge this pull request anymore.

mixflame:
I can't find the address of Dru's fork or I would use than and test

fridgerator:
Does websocket within in a pipeline make sense?  A pipeline might include CRSF or CORS plugs, which doesn't have anything to do with a websocket connection since its not a normal http request.

drujensen:
@fridgerator probably should have read your comment before merging.  Let me know if you think we need to change it.  I think this is a better way of defining sockets but understand your concern.  Maybe we can add this to the :static pipeline or create a separate pipeline.

PR #225, opened 2017-09-03
Fix quotes

faustinoaq:
### Description of the Change

Add quotes

Currently quotes aren't ended in some ecr templates

screenshot_20170903_183228

### Alternate Designs

Nop

### Benefits

Correctly highlighting and syntax

### Possible Drawbacks

Nop


drujensen:
@faustinoaq thanks for the fix

PR #235, opened 2017-09-05
Environment Settings Enhancements (For Discusión Only)

eliasjpr:
> This PR intention is to open discusión of alternate implementation of the environment loader to address some spec related concerns

### Description of the Change

Environment settings loaded via the run command imposes some
vulnerability concerns since it allows arbitrary code execution
on the server. One can have a YAML file that its contents can be
changed to load other objects to the application and expose sensitive
information.

You might be tempted to think that since we’re only creating objects that have
to be defined on our servers, this object instantiation ability wouldn’t be too
bad. Unfortunately, attackers can use this method to execute arbitrary code.

This PR takes the environment.cr and converts it to a class providing:

- Testability: Needed in order to manage specs settings be reloaded or
changed to specific file environment.

- Portable: Can load YAML settings from user-specified locations not just
from within the project.

- Safer: Since is not run as a macro and the code is executed within the
context.

If the user wishes to load environment variables he can safely use the
config/application.cr is the ideal place to use environment variables.

Settings class uses the core lib in YAML.parser which maps to a specific
class.

YAML Mapping

```crystal
    YAML.mapping(
      env: {type: String, default: "development"},
      host: {type: String, default: "localhost"},
      name: {type: String, default: "Amber_App"},
      redis_url: {type: String?, default: nil},
      port_reuse: {type: Bool, default: true},
      port: {type: Int32, default: 3000},
      process_count: {type: Int32, default: 1},
      secret_key_base: {type: String, default: SecureRandom.urlsafe_base64(32)},
      ssl_key_file: {type: String?, default: nil},
      ssl_cert_file: {type: String?, default: nil},
      secrets: {type: Hash(String, String), default: {} of String => String},
      session: {type: Hash(String, Int32 | String), default: {
        "key" => "amber.session", "store" => "signed_cookie", "expires" => 0,
      }},
    )
```

YAML Config File
```yml
name: amber_test_app
secret_key_base: ox7cTo_408i4WZkKZ_5OZZtB5plqJYhD4rxrz2hriA4
port: 4000
port_reuse: true
process_count: 16
ssl_key_file: some key file
ssl_cert_file: some cert file
host: 0.0.0.0
redis_url: "redis://localhost:6379"
session: {
  key: "myapp.session",
  store: "encrypted_cookie",
  expires: 10
}
secrets:  {
  description: "Store your test secrets credentials and settings here.",
  database: "mysql://root@localhost:3306/amber_test_app_test"
}
```

### Benefits

- Testability: Needed in order to manage specs settings be reloaded or
changed to specific file environment.

- Portable: Can load YAML settings from user-specified locations not just
from within the project.

- Safer: Since is not run as a macro and the code is executed within the
context.

### Possible Drawbacks

- No code execution in YAML


spec/support/config/test.yml
```diff
@@ -1,18 +1,18 @@
 name: amber_test_app
```
eliasjpr: Change the some of the contents of this file to ensure that the defaults are not used

src/amber/router/session.cr
```diff
@@ -1,6 +1,6 @@
 module Amber::Router::Session
   class Store
-    getter session_config : Hash(Symbol, String) = Amber::Server.settings.session
+    getter session_config : Hash(String, Int32 | String) = Amber::Server.settings.session
```
drujensen: I prefer symbols for keys.  Can we change this back to a symbol?

src/amber/router/session/cookie_store.cr
```diff
@@ -36,7 +36,7 @@ module Amber::Router::Session
     forward_missing_to session
 
     def self.build(store, settings)
-      new(store, settings[:key].to_s, settings[:expires].to_i)
+      new(store, settings["key"].to_s, settings["expires"].not_nil!.to_i)
```
drujensen: will the expires `not_nil!` assertion fail if they don't provide a value?  if so, should we consider checking and providing a default of 0?

src/amber/server/settings.cr
```diff
@@ -6,28 +6,42 @@ module Amber
...
+    property pubsub_adapter : WebSocketAdapter = WebSockets::Adapters::MemoryAdapter
+    property session : Hash(String, Int32 | String)
+    property ssl_key_file : String?
+    property ssl_cert_file : String?
+    property secrets : Hash(String, String)
 
-    #TODO: Remove as ASAP!
-    class_getter key_generator = Amber::Support::CachingKeyGenerator.new(
-      Amber::Support::KeyGenerator.new(@@secret_key_base.to_s, 5)
+    YAML.mapping(
```
drujensen: nice! 👍

src/amber/server/settings.cr
```diff
@@ -6,28 +6,42 @@ module Amber
...
+    getter key_generator = Amber::Support::CachingKeyGenerator.new(
+      Amber::Support::KeyGenerator.new(@secret_key_base.to_s, 5)
+    )
+
+    property env : String
+    property host : String
+    property name : String
+    property log : ::Logger = ::Logger.new(STDOUT)
+    property redis_url : String?
+    property process_count : Int32
```
elorest: I do like these being instance variables a ton better.

drujensen:
@eliasjpr - @elorest and I discussed the idea you had of changing this from a YAML to a CR.  It sounds like this would provide the most flexibility in setting environment configuration and with a quick spike, @elorest was able to demonstrate how much easier that approach would be.

We would still support encrypt/decrypt but it would be the {environment}.cr file instead of .yaml file.

We also talked about providing a secrets hash for custom settings.

What do you think of this approach instead?

eliasjpr:
I like the idea of using cr files instead. I guess I will wait for a PR to look at the proposed concept maybe we can clean up the current Settings singleton class and make it just a class with properties.

elorest:
@drujensen @eliasjpr I did find a couple issue with my above suggestion. I'd like to have a meeting and discuss pro's and cons. 

Regardless of how we go on this I definitely want to get rid of the singleton class and instantiate settings. That alone will solve many of the issues that both you and I have had with testing this. I also think we need to add `Amber::Server.instance.config` back for backward compatibility reasons.

elorest:
### Current Solution:
#### Pros:
* Allows for secrets of any depth, including nested hashes and arrays which in rails secrets are super useful for storing mailer settings. This is accomplished by creating a namedtuple at compile time. See example A below.
* All keys in secrets are symbols which is more clean and also the same as rails.
* Only a single file has to be created or edited to add a new environment like staging or ci. This is because code like ENV variables can be evaluated at compile time, which eliminates the need for configurations in multiple files. The settings are few enough that I think this makes sense.

#### Cons:
* the load_environment file which gets called via run macro creates string containing code which it inserts into the compilation process via printing to stdout which is coopted by the compiler. This is a little bit script looking and ugly but is short. The crystal standard library makes use of this feature in many places in a similar fashion. Ecr, slang, teeplate, and baked_file_system are examples that wouldn't be possible without it.
* Is opinionated about environment.yml locations (I think this is a good thing). The only way to configure locations is with environment vars, defaults or fall back list of locations. While not optimal I think that the power of this solution outweighs not being able to easily change env.yml locations.

### This PR:
####  Pros:
* Yaml mapping looks better.
* File location is configurable; Because of this tdd unit testing is easier.
* Doesn't require a run using a `run` macro.

#### Cons:
* Code can't be evaluated, requiring additional environment files to create or edit an environment.
* Secrets are flat key/value pairs and with keys being strings. Storing nested settings traditional in secrets will result in compile error with current yaml mapping. See example A below. I've experimented with other options but without being able to predict all possible settings it will fail in yaml mapping.

### Suggestions above that @drujensen and I discussed.

#### Pros
* Evaling from Cr files seems more natural than Yaml probably.
* Macro `run` code (envirionment.cr) can be a lot smaller. Just Read file, decrypt and return crystal code to compile.

#### Cons:
* Secrets are still one dimensional can't be a tuple unless the #{envirionment}.cr is where settings are defined with an optional overload later.
* Default values will be harder to do and could be broken easily by someone defining an invalid type in settings cr.
* Still requires `run` macro.

Example A
```yaml
secrets:
  smtp_settings:
    address: "smtp.mailgun.org"
    port: 587 
    domain: "ls.example.com"
    authentication: "plain"
    enable_starttls_auto: true
    user_name: "postmaster@ls.example.com"
    password: "62b9782c80e8a62eccb852c3826172c5"
```

src/amber/router/session.cr
fridgerator: I prefer symbols as well, I would use whatever is consistent across our api.

eliasjpr:
These are my notes on the comments above:

> Code can't be evaluated, requiring additional environment files to create or edit an environment.

I don't see the necessity to have logic and variables in YAML files if that would have been a need the YAML language would have already supported this naturally. If we need to evaluate logic and variables I would use the application.cr for this use. 

> Secrets are flat key/value pairs and with keys being strings. Storing nested settings traditional in secrets will result in compile error with current YAML mapping. See example A below. I've experimented with other options but without being able to predict all possible settings it will fail in YAML mapping.

I disagree, if the Crystal implementation of YAML is followed we can accomplish having nested types, Crystal YAML lib will return a hash. We should not reinvent the Crystal implementation of YAML by creating our own parsing. 

The feature should start small with a basic implementation and grow from there.

Let's remove the evaluated variables from the YAML and the run command, and start from there and see if there is a greater need. In the instance, we want to use environment variables we can always user **application.cr** to do this since **application.cr** will overwrite the YAML files. Which the framework is already doing.

Since ENV is a hash like hash-like accessor for environment variables we can probably loop over, something like below.
 
```crystal
# server/configuration.cr
class Amber::Configuration
   def load_from_env
     {% for key  in ENV_KEYS do %}
     # load settings.{{key}} = ENV[key] (Int | String conversions needs to happen)
     {% end %}
   end
end

# config/application.cr
# Load from env variables overwritting the YAML settings where necessary
Amber::Server.from_env 
```

If you guys need more clarification on this let me know I would like for us to pair.

elorest:
To me being able to evaluate code in the environments is really important.
I also really like being able setup new environments by just creating one new file.
But we might not ever agree on these features.

I do need to respond to that you said above though is this:
"I disagree, if the Crystal implementation of YAML is followed we can accomplish having nested types, Crystal YAML lib will return a hash. We should not reinvent the Crystal implementation of YAML by creating our own parsing."

Crystal YAML mapping requires that the types and nesting be the same every-time. I've tested all of this. I'm not just making it up. Its not possible to map secrets to be `Hash(String, String | Hash(String, String))`. If you just map to `Hash(String, String)` then nested YAML blows up on `Settings.from_yaml(yml)`

https://carc.in/#/r/2oq0

eliasjpr:
Hey tried this out and it seems it was missing a union definition to support that type of nested hashes check this out and let me know if you think this works https://carc.in/#/r/2oql

elorest:
I don't like that we're deleting a lot of useful functionality but other than having string keys instead of symbols this should work.

https://carc.in/#/r/2oqy
I can’t seem to get recursive aliases to work. In mapping. When I tried it last week Jason said it seemed to be broken in Yaml mapping. So unless we can solve that we’re still limited to whatever depth and options we program from the beginning.
https://crystal-lang.org/docs/syntax_and_semantics/alias.html
See if you can figure it out.

PR #238, opened 2017-09-07
Fix for docker deploy

elorest:
### Description of the Change

Updated path for docker cli.

### DIFF
--- a/src/amber/cli/templates/app/config/deploy/Dockerfile
+++ b/src/amber/cli/templates/app/config/deploy/Dockerfile
@@ -1,10 +1,10 @@
-FROM drujensen/crystal:0.22.0-1
+FROM drujensen/crystal:0.23.1
 
-ENV AMBER_VERSION 0.1.15
+ENV AMBER_VERSION 0.2.3
 
-RUN curl -L https://github.com/Amber-Crystal/amber_cmd/archive/v$AMBER_VERSION.tar.gz | tar xvz -C /usr/local/share/. && cd /usr/local/share/amber_cmd-$AMBER_VERSION && crystal deps && make
+RUN curl -L https://github.com/Amber-Crystal/amber/archive/v$AMBER_VERSION.tar.gz | tar xvz -C /usr/local/share/. && cd /usr/local/share/amber-$AMBER_VERSION && crystal deps && make
 
-RUN ln -sf /usr/local/share/amber_cmd-$AMBER_VERSION/bin/amber /usr/local/bin/amber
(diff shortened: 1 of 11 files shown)


src/amber/cli/templates/app/config/deploy/Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM drujensen/crystal:0.22.0-1
+FROM drujensen/crystal:0.23.1
```
drujensen: can we use the official image?

src/amber/cli/templates/app/config/deploy/Dockerfile
```diff
@@ -1,10 +1,10 @@
...
 
-ENV AMBER_VERSION 0.1.15
+ENV AMBER_VERSION 0.2.3
 
-RUN curl -L https://github.com/Amber-Crystal/amber_cmd/archive/v$AMBER_VERSION.tar.gz | tar xvz -C /usr/local/share/. && cd /usr/local/share/amber_cmd-$AMBER_VERSION && crystal deps && make
+RUN curl -L https://github.com/Amber-Crystal/amber/archive/v$AMBER_VERSION.tar.gz | tar xvz -C /usr/local/share/. && cd /usr/local/share/amber-$AMBER_VERSION && crystal deps && make
```
drujensen: should we look at building this like we are in the root Dockerfile instead of needing to remember to update the version number every time?

elorest: Yeah we definitely should. I'll change that.

src/amber/cli/templates/app/config/deploy/Dockerfile
elorest: We probably should. It would take some refactoring at a few other places though. When this was built there were a few commands backed into the root dockerfile for dev that didn't make a lot of sense for deployment. Root dockerfile is completely different now though.

src/amber/cli/templates/app/config/deploy/Dockerfile
faustinoaq: @elorest I think `Amber-Crystal` should be lowercase `amber-crystal`

elorest: Technically that's true but it used to be `Amber-Crystal` and github works with both.

elorest: Changed that.

PR #242, opened 2017-09-08
Removes deprecated method config

eliasjpr:
Config method was replaced with Configure

### Description of the Change

The config method has been deprecated in favor of `settings` and `configure`

### DIFF
--- a/src/amber/server/configuration.cr
+++ b/src/amber/server/configuration.cr
@@ -8,11 +8,6 @@ module Amber
       def self.configure
         with settings yield settings
       end
-      
-      # NOTE: Deprecated. Here so old releases work.
-      def config
-        with settings yield settings
-      end
 
       def self.settings
         Settings


elorest:
You've got to be kidding. @drujensen and I just added that back yesterday because it broke everything.

eliasjpr:
Why not ask to update the method instead of adding it back?

elorest:
I've mentioned in conversation multiple times over the last few days that this broke the apps for countless people who are using amber and everyone who had the cli from just a few days ago. Including the benchmarks. Dru and I talked about it at length and decided it was a good temporary fix.

elorest:
Because deprecation is the standard practice for removing methods from OSS not compiler errors.

eliasjpr:
Why wasn't an issue opened? It would have avoided me opening a PR, since I don't see any context in regards to this. People expressed on Gitter having issues and those were addressed

faustinoaq:
@elorest Amber is still in development, so breaking changes could happen. People that use Amber should be aware of that.

I think we can make the codebase simpler and effective (even with breaking changes) until Amber 1.0 is released.

> Why wasn't an issue opened? It would have avoided me opening a PR

I think the same.

> Also if the user does not upgrade the version their app should still work because `shards.lock`

@eliasjpr Yep, you are right, I have multiples project with Amber 1.3 and 2.2 and `shard.lock` allow me to compile without any issue.

drujensen:
@elorest @eliasjpr let's create an issue to remember to remove this but I vote for leaving it for now.  

There is an issue with `shards install` on the TechEmpower benchmarks that is performing an update instead of a install so the shards.lock is not being adhered too.  Once that is resolved, we can look at removing this.

@faustinoaq I'm glad you are ok with breaking changes but I worry that people will avoid using Amber because of the instability we have had lately. I know I have been pretty frustrated by it myself so I imagine I'm not the only one.

drujensen:
This PR addresses issue #243

Issue #194, opened 2017-08-27
Cookie and session keys don't persist across server restarts.


elorest:
### Description

Every time server restarts `Support::CachingKeyGenerator` generates new keys causing session, and cookies to all become invalid. This is a bug since rails, sinatra, phoenix, laravel, and hanami all have persistent keys usually set via Env or secrets.yml.

### Steps to Reproduce

1. Create an action and set a session.
2. Created another action which reads and prints the session.
3. Run server and visit both pages.
4. Restart the server and refresh the session printing page without visiting the session writing page first.

**Expected behavior:** Session should print.

**Actual behavior:** Session is nil.

**Reproduces how often:** Everytime.

### Solution

I can work on this while working on environments and secrets. My suggestion would be to pull these keys from secrets and only generate them if they're missing.

drujensen:
This bug is pretty annoying since every code change will log me out of my website.

eliasjpr:
Should we add `amber g secret` command?

elorest:
I'm working on that.

eliasjpr:
Whats the progress on this?

PR #246, opened 2017-09-09
Dj/fix session secret

drujensen:
### Description of the Change

Fixes issue #194 

Sorry for the large PR.  This removes the KeyGenerator and uses the secret_key_base for signing and encrypting the cookies.  

Please be critical of this PR since it may have some security concerns and we want to make we don't introduce any vulnerabilities.

### Benefits

This change avoids the session being lost between restarts of an application.  By using the same key to encrypt or sign the cookies, the session doesn't change and doesn't seem as if its been tampered with.

### Possible Drawbacks

nothing I can think of.


spec/amber/router/cookies_spec.cr
```diff
@@ -108,11 +108,9 @@ module Amber::Router
 
       it "gets an encrypted cookie" do
         cookies = new_cookie_store
-        cookie = HTTP::Cookie::Parser.parse_cookies("user_name=LByguEoiSsJqc1iG%2FPrIujkr5ha0yUi%2Fng2fT4XSX3I%3D--qRpa7wr%2FuVEx5xfyfDrCHzrXjnJv44q1xhqG1XdgaAQ%3D; path=/").first
-
-        cookies[cookie.name] = cookie
-
-        cookies.encrypted["user_name"].should eq "david"
+        cookie = HTTP::Cookie::Parser.parse_cookies("user_name=GfnZpsL07kZjaNvgx7WJuc4OG0ANKyYdmqZ3xPAHX9w%3D--etXkFCeEbVhl4g%2FsdytpbCp0I%2B47WQfOwB1GDRqv4Ac%3D; path=/").first
+        #TODO: Need help generating this
```
drujensen: @eliasjpr @elorest could use some help here.  Let me know how to fix this test.

drujensen:
Also, can someone help with the breaking spec?

src/amber/router/cookies.cr
```diff
@@ -11,15 +11,19 @@ module Amber::Router
...
+        @encrypted ||= EncryptedStore.new(self, @options)
       end
 
       def signed
-        @signed ||= SignedStore.new(self, @key_generator, @options)
+        @signed ||= SignedStore.new(self, @options)
+      end
+
+      def secret
+        @secret ||= Amber::Server.settings.secret_key_base || SecureRandom.urlsafe_base64(32)
```
eliasjpr: Can we pass in on initialize `Amber::Server.settings.secret_key_base` or a property something like:

```crystal
@secret = Amber::Server.settings.secret_key_base
# or 
def initialize(@host = nil, @secure = false, @secret = Amber::Server.settings.secret_key_base)
```

I would also consider removing `|| SecureRandom.urlsafe_base64(32)` since secret_key_base always exists. I think having this can cause the same effect you are trying to resolve with this PR.

src/amber/server/configuration.cr
```diff
@@ -22,10 +22,6 @@ module Amber
         Settings
       end
 
-      def self.key_generator
-        settings.key_generator
-      end
-
       def self.secret_key_base
```
eliasjpr: You can get rid of this method if you are using `Amber::Server.settings.secret_key_base`. This is a helper method for `Amber::Server.secret_key_base`

src/amber/router/cookies.cr
drujensen: :+1:

src/amber/server/configuration.cr
drujensen: aaah, ok.  I will use the helper method since its shorter.

drujensen:
@eliasjpr when you get a chance can you take another look?  I removed the dependency on `Amber::Server` by passing in the `secret` as a parameter to `build` method.  That is much cleaner and removes the dependency.  Thanks!

src/amber/server/configuration.cr
eliasjpr: 👍🏻

drujensen:
@eliasjpr curious, why are we using a chaining pattern?  

Today, this is possible:
```
cookies.encrypted.signed.permanent("strange")
```

what is the use case for something like this?

eliasjpr:
Great that you passed in the secret, that would have been my first option. 

In regards to the chaining the idea was to be able to set signed and encrypted cookie that expires. I did not realize that it gave you those chaining options, that's a side effect of the implementation. The ideas was to have cookies.encrypted.permanent and cookies.signed.permanent

PR #251, opened 2017-09-10
Removes deprecated method config

drujensen:
The TechBenchmark has been updated to the latest Amber so this issue will no longer occur.  We should be good to remove this now.


drujensen:
@eliasjpr we didn't change the templates.  they always used the newer `configure` method.  we only added this for backward compatibility since there was an issue with shard.lock on the TF benchmarks.

PR #253, opened 2017-09-12
combine websockets js into single file

fridgerator:
### Description of the Change

Combine js files into single asset / file.  I'm going to begin working on client side heartbeat and reconnection functionality and wanted to do this in a separate PR.

### Benefits

I had the javascript files separated out in case further non-websockets functionality wanted to be written to the javascript library, however for now a single file will be easy to work with and read.

### Possible Drawbacks

None


drujensen:
after writing my comment, a question came to mind.  

Will someone want to use a socket without a pubsub channel?  Is it something we should consider?

fridgerator:
My initial thought when writing the javascript library was that it could be optional, and extra functionality beyond what it provides could be done manually.  I suppose though it depends on how many users need `x` feature.

Would it even be possible to use sockets without channels given the server implementation?  I'm not sure I would have to think how it could be used without.

Let me know what you think.

PR #254, opened 2017-09-12
fix links on homepage

drujensen:
### Description of the Change

The links on the generated homepage did not work.  This also add a target="_blank" so it will open a new tab.

### Alternate Designs

Please provide feedback on other resources and links we should provide on this page.

PR #255, opened 2017-09-12
Fix links.

faustinoaq:
### Description of the Change

Currently links in default home page don't works.

### Alternate Designs

Nop

### Benefits

Fix links location.

### Possible Drawbacks

Nop


faustinoaq:
@drujensen Can we include this fix in current release or we need to publish a new one?

drujensen:
@faustinoaq beat you to it.  PR #254

PR #259, opened 2017-09-13
Add X-Powered-By header

faustinoaq:
### Description of the Change

Show `X-Powered-By` in response headers.

screenshot_20170913_121131

Similar as Kemal is doing:

kemal

### Alternate Designs

Nop

### Benefits

People can see what sites do use Amber :)

### Possible Drawbacks

Nop

### DIFF
--- a/spec/amber/router/context_spec.cr
+++ b/spec/amber/router/context_spec.cr
@@ -267,4 +267,14 @@ describe HTTP::Server::Context do
       end
     end
   end
+
+  describe "X-Powered-By Header" do
+    it "initializes context with X-Powered-By: Amber" do
+      headers = HTTP::Headers.new
+      request = HTTP::Request.new("GET", "/", headers)
+      context = create_context(request)
+      context.response.headers.add "X-Powered-By", "Amber"
+      context.response.headers["X-Powered-By"].should eq "Amber"
+    end
+  end
 end
--- a/spec/amber/router/context_spec.cr
+++ b/spec/amber/router/context_spec.cr
@@ -273,7 +273,8 @@ describe HTTP::Server::Context do
       headers = HTTP::Headers.new
       request = HTTP::Request.new("GET", "/", headers)
       context = create_context(request)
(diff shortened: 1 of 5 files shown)


spec/amber/router/context_spec.cr
```diff
@@ -267,4 +267,14 @@ describe HTTP::Server::Context do
       end
     end
   end
+
+  describe "X-Powered-By Header" do
+    it "initializes context with X-Powered-By: Amber" do
+      headers = HTTP::Headers.new
+      request = HTTP::Request.new("GET", "/", headers)
+      context = create_context(request)
+      context.response.headers.add "X-Powered-By", "Amber"
```
drujensen: In order to test if the pipeline is correctly adding the header, you shouldn't set this in the test.  you should let the pipeline add it.

faustinoaq: Ok, let me fix this

drujensen: Try something like this:
```
pipeline = Amber::Pipe::Pipeline.new
pipeline.call(context)
context.response.headers["X-Powered-By"].should eq "Amber"
```

eliasjpr: @faustinoaq @drujensen is correct you can unit test the Pipeline and verify that the header is set. That's a simple test setup.

faustinoaq:
Travis is failing with latest commit, but it works in my local machine:

Some clue ?

image

faustinoaq:
Hi @drujensen @eliasjpr  I think `Amber::Pipe:::Pipeline.new` is not working well :sweat_smile:

faustinoaq:
Oh, I see my local spec are failing too if I execute all specs

screenshot_20170914_191116

Why if I execute the single test it pass?

ok

faustinoaq:
Travis is happy now, and the pipeline is well tested :)

@drujensen Can we squash and merge this PR now ?

drujensen:
@faustinoaq thanks!

PR #268, opened 2017-09-19
Add reference field type to scaffolding

drujensen:
### Description of the Change

Now that Granite has relationships, this PR provides the ability to easily generate a reference (ref for short) between two models.

```
amber g auth user
amber g scaffold post user:reference title:string body:text
amber g scaffold comment user:ref post:ref comment:text
```
This will:
- create the foreign key of `{model}_id : BIGINT` in the migrations
- create an index on the foreign key in the migrations
- create the `belongs_to` macro in the model
- include the `{model}_id` in the controller for accepted attributes
- add a `select_field` with a collection using the `id` in the form view
- add a column for the model with the id column in the show and index views
- supports both `ref` and `reference` as the keyword for the field type

screen shot 2017-09-18 at 10 04 30 pm

### Alternate Designs

can't think of any

### Benefits

This simplifies the generation of relationships between models and will allow you to quickly build out your persistence layer with little to no coding.

### Possible Drawbacks

can't think of any


src/amber/cli/templates/app/config/database.yml.ecr
```diff
@@ -3,7 +3,7 @@
    when "mysql" -%>
   database: mysql://root@localhost:3306/<%= "#{@name}_development" %>
 <% when "pg" -%>
-  database: postgres://postgres:@pg:5432/<%= "#{@name}_development" %>
+  database: postgres://root:@localhost:5432/<%= "#{@name}_development" %>
```
eliasjpr: 🎉

src/amber/cli/commands/generate.cr
```diff
@@ -6,7 +6,7 @@ module Amber::CLI
       class Options
         arg "type", desc: "scaffold, model, controller, migration, mailer, socket, channel, auth", required: true
         arg "name", desc: "name of resource", required: true
-        arg_array "fields", desc: "name:string body:text age:integer published:bool"
+        arg_array "fields", desc: "user:ref name:string body:text age:integer published:bool"
```
eliasjpr: should we be consistent and not use abbreviations `references`? See https://www.google.com/search?q=ref+abbreviation&oq=ref+abbr&aqs=chrome.0.0j69i57j0l4.3968j0j1&sourceid=chrome&ie=UTF-8

Also, also what do you think of something like `Post  title:string body:text belongs_to:user has_many:comments`

src/amber/cli/templates/field.cr
```diff
@@ -65,5 +67,9 @@ module Amber::CLI
         raise "type #{@type} not available"
       end
     end
+
+    def ref?
+      self.type == "ref"
```
eliasjpr: See above comment

src/amber/cli/templates/field.cr
```diff
@@ -17,6 +17,8 @@ module Amber::CLI
         time:      ["time", "Time", "TIMESTAMP"],
         timestamp: ["time", "Time", "TIMESTAMP"],
         password:  ["password", "String", "VARCHAR"],
+        ref:       ["ref", "Int64", "BIGINT"],
+        reference: ["ref", "Int64", "BIGINT"],
```
eliasjpr: Ahh good. Is `reference` the correct keywork or `references` plural?

src/amber/cli/templates/app/config/database.yml.ecr
drujensen: Yes, I opened a ticket https://github.com/amberframework/granite-orm/issues/67 to look at supporting Amber new environment yml.  Not this PR though.

src/amber/cli/commands/generate.cr
drujensen: This PR supports both `ref` and `reference`.  

Regarding `references`, This is describing the type of field so it should not be plural since its a noun, not an action. It answers the question of what type the field is.  `This field is a reference`, not `This field is a references`.

I will change the code to use the unabbreviated `reference` so it's clear.

Regarding the `belongs_to:user` and `has_many:comments`, we should discuss if we want to allow for mixing the order of the `name:type` and how we would do that.  I will leave that for a different PR.

src/amber/cli/templates/field.cr
drujensen: This is describing the type of field so it should not be plural since its a noun, not an action. It answers the question of what type the field is.  `This field is a reference`, not `This field is a references`.

src/amber/cli/templates/field.cr
drujensen: :+1:

drujensen:
@eliasjpr I changed the `ref` to `reference` throughout the code base.  When you get a chance, can you take another look at it? thx!

drujensen:
@eliasjpr I checkout and rails supports:
```
$ rails g scaffold campaign name:string account:belongs_to
```
I've added this as an optional type as well.

drujensen:
@eliasjpr although I disagree with the pluralism of the type, it looks like rails used `references`.  ugh.

For ease of use for rails developers, should we add this as an option as well?

drujensen:
Added `references` as another alias for rails developers.  :-?

eliasjpr:
I was just wondering, we don`t have to comply with rails way. I was just wondering if it was proper English :)

PR #274, opened 2017-09-20
Refactoring error pipe

### LINKED ISSUE #237: Add easy way to customize error pages
### Description

Currently I need to hack Amber classes to customize error pages.

### Steps to Reproduce

1. Create a new amber project.
2.  Try to customize an error page as 404.

**Expected behavior:** Easy way to add error pages like Kemal, see: http://kemalcr.com/guide/#helpers

**Actual behavior:** No easy way to add error pages.

### Versions

```
$ lsb_release -a
LSB Version:    n/a
Distributor ID: ManjaroLinux
Description:    Manjaro Linux
Release:        17.0.2
Codename:       Gellivara
$ amber -v
Amber CMD (ambercr.io) - v0.2.3
$ crystal -v
Crystal 0.23.1 (2017-07-17) LLVM 4.0.1
```

### Additional Information

Currently I'm using this code:

```crystal
module Amber::Exceptions
  class RouteNotFound < Base
    include Amber::Controller::Render
    LAYOUT = "application.ecr"

    def initialize(request)
      @status_code = 404
      io = IO::Memory.new
      response = HTTP::Server::Response.new(io)
      context = HTTP::Server::Context.new(request, response)
      flash = context.flash
      super(render("404.ecr"))
    end

    def set_response(response)
      response.headers["Content-Type"] = "text/html"
      response.print message
      response.status_code = status_code
    end
  end
end

class ErrorHandler < Amber::Pipe::Base
  include Amber::Controller::Render
  LAYOUT = "application.ecr"

  def call(context)
    call_next(context)
  rescue ex
    flash = context.flash
    context.response.status_code = 500
    context.response.content_type = "text/html"
    context.response.print render("500.ecr")
  end
end
```

faustinoaq:
### Description of the Change

`Amber::Error::Pipe` was refactored and added inside `config/router.cr` being part of web and static pipelines, allowing to add customized error pages more easily.

- Add Amber::Controller::Error
- Add Amber::Pipe::Error.new to configured pipelines
- Add error_controller.cr
- Format files
- Raise RouteNotFound on StaticController
- Simplify Amber::Pipe::Error
- Call Pipe::Error on Exceptions::Base inside Pipe::Pipeline
- Fix error spec
- Update docker image

#### Error scenarios

Currently Errors have 3 scenarios:

1. `Amber::Error::Pipe` is disabled and shows default Exception message.

screenshot_20170920_075142

screenshot_20170920_075749

2. `Amber::Error::Pipe` is enabled and shows default responses for `not_found`, `internal_server_error`, etc.

screenshot_20170920_075231

screenshot_20170920_075705

3. `Amber::Error::Pipe` is enabled and `src/controllers/error_controller.cr` is customized, so developers can use templates or customized messages.

```crystal
# src/controllers/error_controller.cr
class Amber::Controller::Error < Amber::Controller::Base
  LAYOUT = "application.slang"

  def not_found
    render("404.slang") # => src/views/error/404.slang
  end

  def internal_server_error
    render("500.slang") # => src/views/error/500.slang
  end
end
```

screenshot_20170920_075415

screenshot_20170920_075844

### Alternate Designs

Overwrite Exceptions and Amber::Pipe:Error every time a developer want to customize error pages, by example [error_controller.cr](https://github.com/amberframework/amberframework.org/blob/4c5960bacc9210ca93f161704a82fe163065a905/src/controllers/error_controller.cr) used in amberframework.org (I think this approach isn't easy and very repetitive)

### Benefits

Fix #237 

### Possible Drawbacks

Nop

/cc @eliasjpr @drujensen @elorest @fridgerator


src/amber/controller/error.cr
```diff
@@ -0,0 +1,47 @@
...
+      message_based_on_content_type(@ex.message, content_type)
+    end
+
+    def internal_server_error
+      context.response.content_type = "text/plain"
+      "ERROR: #{@ex.inspect_with_backtrace}"
+    end
+
+    def forbidden
+      if context.request.headers["Accept"]?
```
eliasjpr: Maybe
```crystal
response.content_type = if context.request.headers["Accept"]?
  request.headers["Accept"].split(",").first
else
  "text/plain"
end

private def response_format(message)
  case response.content_type
  when "application/json"
    {"error": message}.to_json
  when "text/html"
    "<html><body>#{message}</body></html>"
  else
    message
  end
end
```

faustinoaq: Done! :+1:

faustinoaq:
phoenix

Something like Phoenix error pages [0] will be amazing, we already have the back trace info, just need to add some interactivity. It will be a whole PR to achieve that.

[0] - https://github.com/phoenixframework/phoenix/issues/1776

PR #278, opened 2017-09-27
Render Refactor [Breaking Change] - Also fixes some things

elorest:
### Description of the Change

Fixes bug that @eliasjpr found where you couldn't use render multiple times in the same project as content would overwrite the previous content.

Adds `partial` and `template` as  named params to render, this makes render closer to rails so it should be familiar to more people.

1. `render(partial: "_form.slang")` will render supplied file without layout.
2. `render("show.slang", layout: false)` will render template without layout.
3. `render("show.slang")` will render supplied file with default layout.

### Possible issues.

Breaking change as `render("template", false)` has been replaced with the more verbose  `render("template", layout: false)`.
`render_template` is an internal helper macro and shouldn't be use. It is now private.

### DIFF
--- a/src/amber/controller/render.cr
+++ b/src/amber/controller/render.cr
@@ -2,7 +2,7 @@ module Amber::Controller
   module Render
     LAYOUT = "application.slang"
 
-    macro render_template(filename, path = "src/views")
+    private macro render_template(filename, path = "src/views")
       {% if filename.id.split("/").size > 2 %}
         Kilt.render("{{filename.id}}")
       {% else %}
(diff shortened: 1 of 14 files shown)


src/amber/controller/render.cr
```diff
@@ -11,27 +11,27 @@ module Amber::Controller
...
-        {% end %}
+      {{filename = template || partial}}
 
-        {% if layout && !partial %}
-          content = %content
-          %content = render_template("layouts/#{{{layout.class_name == "StringLiteral" ? layout : LAYOUT}}}", {{path}})
+      {% if filename.id.split("/").size > 1 %}
+        %content = render_template("#{{{filename}}}", {{path}})
+      {% else %}
+        {% if folder.id.ends_with?(".ecr") %}
```
drujensen: what about `slang`, `mustache`, `temel` and `crikey`?

elorest: Slang already works. We paired on this line a few months ago, in order to fix an issue with ecr working differently than lang. As far as temel, crikey, hopefully it will work with kilt unless they don't something weird at which point we might have to handle them specifically too.

drujensen: oh yes, I recall.  That looks so out of place. oh well.

PR #281, opened 2017-09-28
Some missing changes about Fix for video in safari

faustinoaq:
### Description of the Change

Adds some missing changes of the previous PR

Please see https://github.com/kemalcr/kemal/pull/405/files

### Alternate Designs

Nop

### Benefits

This update fixes definitely video on safari

### Possible Drawbacks

Nop


faustinoaq:
@eliasjpr should we add `env.response.headers["X-Content-Type-Options"] = "nosniff"` too ? [0]

[0] - https://github.com/kemalcr/kemal/blob/master/src/kemal/helpers/helpers.cr#L102

PR #283, opened 2017-09-29
only apply optional rule if param is present and pass otherwise

sgargan:
### Description of the Change
The optional parameter rule checked to see if the param was present and only ran the validation block if it was found. If the param is not present it returned nil and this nil was interpreted by the unless clause that checked the rules as a failure. This change allows the rule to pass if the param is not present.


drujensen:
@sgargan thanks for the fix

eliasjpr:
👍 Thank you!

Issue #280, opened 2017-09-28
Amber console command requiring files that do not exists in path


eliasjpr:
### Description

When running `amber console` and directories under `/src` do not exists it outputs message

```shell
± amber c
while requiring "./src/mailers/*": can't find file './src/mailers/*' relative to '/Users/eliasperez/Sites/amberframework.org'

require "amber";require "./src/controllers/application_controller";require "./src/controllers/*";require "./src/models/*";require "./src/mailers/*";require "./src/views/*"
                                                                                                                          ^
icr(0.23.1) >
``` 

It would be great to require files and directories dynamically.

### Steps to Reproduce

1. Generate amber app `amber new somename`
2. remove directory under `./src`
3. Run `amber console`

**Expected behavior:** Directories and files should be required as they are present

**Actual behavior:** Directories/Files are required regardless of their existence

**Reproduces how often:** 100%

### Versions

Amber CMD (ambercr.io) - v0.2.6

eliasjpr:
Addressed with #285

PR #285, opened 2017-10-01
#280 fix requires for console

drujensen:
### Description of the Change

Fix for issue #280 

This change adds a new helper module for the generators so you can easily add a route, pipeline or dependency.

Dependencies are now in the `config/application.cr` instead of statically listed in each binary that needs it.  The console, the main application and the specs now only `require ../config/*` that will load the `application.cr` which will have all the required dependencies.

If a new source directory is created by a generator, it will now also add a dependency to the `application.cr`

### Alternate Designs

The existing design required modifying the list of dependencies to multiple places.  This change moves that to one place so you can include everything you need by including the "application.cr"

### Benefits

The helpers provide the ability to insert code in appropriate places.  This is done in several of the configuration files when using a generator.

This places all the required dependencies in one place.  The application, specs and console now can include a single file and they will have all the dependencies needed.

We may also use this if we want to include one amber project in another using a shards `config/application.cr`. 

### Possible Drawbacks

none that I can think of

### DIFF
--- a/src/amber/cli/templates/auth.cr
+++ b/src/amber/cli/templates/auth.cr
@@ -2,6 +2,7 @@ require "./field.cr"
 
 module Amber::CLI
   class Auth < Teeplate::FileTree
+    include Amber::CLI::Helpers
     directory "#{__DIR__}/auth"
 
     @name : String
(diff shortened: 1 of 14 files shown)


src/amber/cli/templates/auth.cr
```diff
@@ -24,7 +25,23 @@ module Amber::CLI
       @timestamp = Time.now.to_s("%Y%m%d%H%M%S")
       @primary_key = primary_key
       @visible_fields = @fields.reject(&.hidden).map(&.name)
-      add_route
```
elorest: Is there a good reason to get rid of the method and write it all inline here? I know you're calling a function that builds this, but maybe it should still be extracted out from initialize?

src/amber/cli/templates/helpers.cr
```diff
@@ -0,0 +1,37 @@
+module Amber::CLI::Helpers
+  def add_routes(pipeline, route)
```
elorest: Ok I like this after I figured out what you were doing.

PR #286, opened 2017-10-03
Dj/54 support model flag

drujensen:
### Description of the Change

WIP - still needs specs

Addresses #50 

This provides the ability to use `-m crecto` as the repo pattern with Amber.  This also restructures the models, views, controllers, and migrations into separate groups and then leverage the same templates where possible in the generators.

NOTE: The `auth` generator is not yet supported for crecto.

### Benefits

This allows developers to choose if they prefer the ActiveRecord approach or the classical Repo pattern.

### Possible Drawbacks

Too many choices is confusing for the new developers who don't have the experience yet to pick the best one for their needs.


src/amber/cli/templates/controller.cr
```diff
@@ -3,7 +3,7 @@ require "./field.cr"
 module Amber::CLI
   class Controller < Teeplate::FileTree
     include Amber::CLI::Helpers
-    directory "#{__DIR__}/controller"
+    directory "#{__DIR__}/controller/plain"
```
eliasjpr: What does plain mean?

drujensen: I was struggling to come up with a good name for it. It’s a controller without a model or resource.  It doesn’t have the restful 7 built in methods.

Can you think of a better name?  Obviously Plain isn’t clear. ;-)

drujensen:
@elorest made the requested changes.

PR #287, opened 2017-10-04
remove static controller

drujensen:
### Description of the Change

Remove unneeded StaticController from project and add it to amber core until we can remove it completely.

### Alternate Designs

In the future, would love to remove this altogether.

### Benefits

Remove the clutter from the controllers directory.

### Possible Drawbacks

None that I can think of


src/amber/cli/templates/app/src/controllers/application_controller.cr.ecr
```diff
@@ -3,4 +3,9 @@ require "jasper_helpers"
 class ApplicationController < Amber::Controller::Base
   include JasperHelpers
   LAYOUT = "application.<%= @language %>"
+
+  # If static resource is not found then raise an exception
+  def index
```
elorest: Don't think this will cause a failure, but it could be bad that all controller that inherit from application_controller will have an index action that raises a 404 when called. What do you think of moving this controller into amber core `Amber::Controller::Static < Amber::Controller::Base`. Then at least it doesn't clutter the project, until we figure out a way to get rid of it entirely.

src/amber/controller/base.cr
```diff
@@ -26,5 +26,10 @@ module Amber::Controller
     def controller_name
       self.class.name.gsub(/Controller/i, "")
     end
+
+    # If static resource is not found then raise an exception
+    def index
```
elorest: Still my concern with this is that all controllers inherit it. My initial comment on this suggested moving it to a new controller in amber/controller/ until we remove it. That way it doesn't clutter the project and it gives us an incentive to remove the need for it so it doesn't clutter amber core either.

`Amber::Controller::Static` is what I would suggest calling it for now. What are your thoughts on that?

drujensen: oooh, right.  i see. ok.

drujensen:
@elorest thanks for the feedback.  This still feels wrong but at least we eliminated the extra controller.

elorest:
Yeah I agree. Thanks for working on this.

PR #289, opened 2017-10-05
ClientSocket `on_disconnect` public method

### LINKED ISSUE #288: ClientSocket needs on_disconnect method
For functionality needed when a user disconnects, a public `on_disconnect` method needs implemented in instances of the `ClientSocket` struct

fridgerator:
closes #288 

### Description of the Change

A method doesn't exist for handling `ClientSocket` disconnects within the application.  There was already an `on_connect` public method, this adds a public `on_disconnect` method that can be optionally overritten for functionality when the client disconnects.

### DIFF
--- a/src/amber/websockets/client_socket.cr
+++ b/src/amber/websockets/client_socket.cr
@@ -73,17 +73,20 @@ module Amber
         end
       end
 
-      def session
-        @session ||= @context.session
-      end
-
-      def cookies
-        @cookies ||= @context.cookies
-      end
-
       # Authentication and authorization can happen here
       def on_connect : Bool
         true
+      end
+
+      # On socket disconnect functionality
+      def on_disconnect; end
+
+      protected def session
+        @session ||= @context.session
+      end
+
+      protected def cookies
+        @cookies ||= @context.cookies
       end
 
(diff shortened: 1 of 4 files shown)


src/amber/websockets/client_socket.cr
```diff
@@ -73,17 +73,20 @@ module Amber
         end
       end
 
-      def session
-        @session ||= @context.session
+      # Authentication and authorization can happen here
+      def on_connect : Bool
```
eliasjpr: why not `abstract def on_connect : Bool`?

src/amber/websockets/client_socket.cr
```diff
@@ -73,17 +73,20 @@ module Amber
...
+        true
       end
 
-      def cookies
-        @cookies ||= @context.cookies
+      # On socket disconnect functionality
+      def on_disconnect; end
```
eliasjpr: why not `abstract def on_disconnect; end`?

fridgerator: I didn't make it abstract, because its optional

src/amber/websockets/client_socket.cr
fridgerator: also optional, abstract makes it required

PR #299, opened 2017-10-11
Prevent lower case request methods (i.e. put)

veelenga:
### Description of the Change

`a href="/path?_method=PUT` fails because of CSRF, but `a href="/path?_method=put` doesn't, which opens a hole for cross site forgery attack. This change is about to fix it.

### DIFF
--- a/src/amber/router/params.cr
+++ b/src/amber/router/params.cr
@@ -16,6 +16,8 @@ module Amber::Router
     OVERRIDE_HEADER  = "X-HTTP-Method-Override"
     OVERRIDE_METHODS = %w(PATCH PUT DELETE)
 
+    @override_method : String?
+
     alias ParamTypes = Nil | String | Int64 | Float64 | Bool | Hash(String, JSON::Type) | Array(JSON::Type)
 
     # clear the params.
@@ -54,15 +56,24 @@ module Amber::Router
     # don't do it. Use a POST instead.
     def override_request_method!
       # If the current request method is not GET or POST it means that it was
-      # already overrided
+      # already overridden
       return unless %(GET POST).includes? request.method
-      method = params[METHOD]? || override_header?
-      request.method = method if method && OVERRIDE_METHODS.includes? method.upcase
+      if (method = override_method) && override_method?
+        request.method = method
+      end
     end
 
     # Determines the existence of HTTP Request Method Override in headers
     def override_header?
       request.headers[OVERRIDE_HEADER]?
+    end
+
+    private def override_method?
+      OVERRIDE_METHODS.includes? override_method
+    end
+
+    private def override_method
+      @override_method ||= (params[METHOD]? || override_header?).try &.upcase
     end
 
     def merge_route_params
--- a/spec/amber/router/context_spec.cr
+++ b/spec/amber/router/context_spec.cr
@@ -39,6 +39,15 @@ describe HTTP::Server::Context do
           context.request.method.should eq "PATCH"
         end
       end
+
+      it "overrides form request method only by upper case value" do
+        header = HTTP::Headers.new
+        header["content-type"] = "application/x-www-form-urlencoded"
+        request = HTTP::Request.new("GET", "/?test=test", header, "_method=put")
+
+        context = create_context(request)
+        context.request.method.should eq "PUT"
+      end
     end
 
     it "overrides form POST method to PUT, PATCH, DELETE" do
--- a/spec/amber/router/pipe/csrf_spec.cr
+++ b/spec/amber/router/pipe/csrf_spec.cr
@@ -5,7 +5,7 @@ module Amber
     describe CSRF do
       context "when requests have HTTP methods" do
         CSRF::CHECK_METHODS.each do |method|
-          it "raises forbbiden error for PUT request" do
+          it "raises forbidden error for #{method} request" do
             csrf = CSRF.new
 
             request = HTTP::Request.new(method, "/")


hugoabonizio:
I think Amber's behavior of accepting method overriding using query string on a GET request a little weird.

AFAIR Rails doesn't have that. Rails uses a link helper to create a form under the hood, similar to what we're doing [here](https://github.com/crystal-community/crystal-ann/blob/master/public/javascripts/main.js#L17-L42).

In my opinion method overriding should be applied to POST requests only.

drujensen:
What I find weird is that browsers still don't support PUT and DELETE as HTTP actions.

@hugoabonizio why restrict it to only POST?  Is there a security concern?

hugoabonizio:
@drujensen Yes, I agree that browsers should accept these methods, but there are semantic and security concerns regarding mimicking other HTTP verbs in a GET request.

### Semantic

GET requests are supposed to have no request body, while POST/PUT/PATCH are usually used to send data through its body (forms, upload, etc). Having a _request body_ on a URI to override a method behavior doesn't make sense if we have another method that accepts requests with a body (POST).[[1](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)]

### Security

Sending a request payload on a GET request might lead to sensitive information leak. GET based forms can expose client data on URI to third parties and are bad pratice. [[2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec15.html#sec15.1.3)]

[Kemal](https://github.com/kemalcr/kemal/blob/a1520de7ed3865fa73258343a80fad4f20666a99/src/kemal/ext/request.cr#L22-L32), for example, checks ```_method``` param only on POST requests.

Ambers users may be aware of those security issues and don't abuse method overriding, but I think the _secury by default_ principle fits well here.



1. https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html
2. https://www.w3.org/Protocols/rfc2616/rfc2616-sec15.html#sec15.1.3

drujensen:
Do you consider the method type sensitive data?

drujensen:
@hugoabonizio I see what you are saying.  The data would need to be sent in the URL params, its not just the method overload.

So the use case that drives me nuts is having to use POST for DELETE.  There are many times when I just want to use an anchor link to perform a DELETE.

What are your thoughts here?

hugoabonizio:
@drujensen Since DELETE requests usually don't use request body, can be reasonable to use a GET as a DELETE, but I still think having a ```?_method=DELETE``` on URI feels wrong.

Maybe a nice approach is to add a helper that makes links with ```data-method="delete"``` create a form behind the scenes, like Rails. WDYT?

drujensen:
To me, creating a helper to handle DELETE feels more wrong.  They both feel wrong but I deem the _method=DELETE less wrong.  (The real wrongness is having to method override in the first place.)

When I built this, I felt restrictions like these were over zealous.  Rails did this and then others followed.  Even if we do restrict this, there is nothing stopping someone from sending URL params and using a GET method to modify the data.  I don't see how restricting this will prevent that.

hugoabonizio:
@drujensen The helper would require JS to create a plain form, no need to AJAX request.

Another argument against using ```_method``` on URIs is making cache (in front of a amber app) useless, since you need to pass a ```csrf``` on query string too, and the token is different for each request.

drujensen:
@hugoabonizio that is very true and a good argument for creating a helper.

drujensen:
@hugoabonizio Actually, I don't think you can client-side cache a form either since the CSRF will change per request.  You would need to use headers.

elorest:
@drujensen I believe he's talking about serverside caching. 

@hugoabonizio Varnish or nginx could be configured to cache with query string as well. In fact _method=delete could trigger it to delete the cache of the object on the same url. I don't think this alone is a good enough reason not to allow method override on gets. Rails use to do it this way before they added their js helper in rails-ujs, which has always seemed a bit more hackish to me.

hugoabonizio:
@drujensen @elorest Disregard my comment about caching, I realiazed that it doesn't make sense after publishing :smile: 

That was not my point anyway, the problem to me is more of semantics, but if this behavior is desired by core team I have no complains!

Let's get this merged :bowtie:

elorest:
Could we have a discussion about my comment above? Why not just capitalize put to PUT instead of rejecting it? It will be a bit confusing if it's just ignored and the wrong route is returned.

src/amber/router/params.cr
```diff
@@ -56,8 +56,8 @@ module Amber::Router
       # If the current request method is not GET or POST it means that it was
       # already overrided
       return unless %(GET POST).includes? request.method
-      method = params[METHOD]? || override_header?
-      request.method = method if method && OVERRIDE_METHODS.includes? method.upcase
+      method = (params[METHOD]? || override_header?).try &.upcase
```
eliasjpr: Small recommendation. I didn't realize this till now, but this method has two responsibilities if not three, can we extract the parsing of the `request method` into its own function. Something like

```crystal
def override_request_method!
  return unless %(GET POST).includes? request.method
  request.method = override_method if override_method?
end 

private

def override_method?
  OVERRIDE_METHODS.includes? override_method
end

def override_method
  @_override_method ||= (params[METHOD]? || override_header?).try &.upcase
end
```

What do you think?

hugoabonizio:
@elorest I didn't understand, isn't the code doing exactly what you're saying?

veelenga:
Guys, thanks for review and discussion.

@eliasjpr done with a slight modification so it compiles :)

> Why not just capitalize put to PUT instead of rejecting it?

@elorest sorry, probably I had to be more descriptive, but this is exactly what I was trying to achieve :) Have I missed something?

elorest:
Looking at it now it looks as I would expect. Might have missed something this morning. Can't seem to see that commit anymore though. Looks good now. 👍

marksiemers:
@drujensen - I always wondered about http verb support in browsers, but then I found out that it isn't the browsers but the HTML standard itself that doesn't support other verbs in forms:
> Browsers do support PUT and DELETE, but it's HTML that doesn't.
> 
> For example, a browser will initiate a PUT request via Javascript (AJAX), but not via HTML `<form>` submission.
> 
> This is because HTML 4.01 and the final W3C HTML 5.0 spec both say that the only HTTP methods that their form elements should allow are GET and POST.
> 
> There was much discussion about this during the development of HTML 5, and at one point they got added to HTML 5, only to be removed again. The reason the additional methods were removed from the HTML 5 spec is because HTML 4-level browsers could never support them (not being part of HTML at the time they were made); and there is no way to allow them to do so without a JavaScript shim; thus, you may as well use AJAX.

Source: https://stackoverflow.com/questions/16805956/why-dont-browsers-support-put-and-delete-requests-and-when-will-they

PR #300, opened 2017-10-11
Use quartz_mailer

faustinoaq:
### Description of the Change

Change `quartz-mailer` by `quartz_mailer` depends on https://github.com/amberframework/quartz-mailer/pull/3

Currently amber requirements on amberframework organization are using `_` to separate words (amber_spec, granite_orm, amber_session, jasper_helpers) I think we should follow the same pattern.

### Alternate Designs

No

### Benefits

Same pattern on require names

### Possible Drawbacks

No


evangelinux:
You forgot to change the name of the quartz_mailer shard to quartz_mailer in <project>/shard.yaml when you generate it with the command `amber new <project> ` in amber v0.2.7. You get this error when you type `shards install`:
`Error shard name (quartz_mailer) doesn't match dependency name (quartz-mailer)`

faustinoaq:
Yes @evangelinux I forgot modify `shard.yml` we need to fix that :+1:

faustinoaq:
Seems it was done 😅 https://github.com/amberframework/amber/blob/4a4d59d5c16ba2e0c1524e28513d844462ba92dc/src/amber/cli/templates/app/shard.yml.ecr#L32

We need to publish a new version to fix this issue, because old amber (0.2.7) is installing latest amber (master), so dependencies mismatch.

Maybe we can lock amber on `shard.yml` to the stable branch, @elorest WDYT ?

faustinoaq:
Oh, I was wrong, I see a fix PR here https://github.com/amberframework/quartz-mailer/pull/4

drujensen:
@faustinoaq  I thought we wanted to use the underscore for the names. why do you think you were wrong?

PR #301, opened 2017-10-11
Extract CSS from JS bundle

faustinoaq:
### Description of the Change

It moves all the required *.css modules in entry chunks into a separate CSS file.

~~Also add the capability of minify CSS on production.~~

### Alternate Designs

No

### Benefits

- Faster runtime (less code and DOM operations)
- Styles are no longer inlined into the JS bundle.
-  If your total stylesheet volume is big, it will be faster because the CSS bundle is loaded in parallel to the JS bundle.

### Possible Drawbacks

No


drujensen:
@faustinoaq My hope was to keep the web pack and npm packages to a bare minimum.   This adds two more npm packages and a plugin that is a beta version.

Is there a way to do this without the extra packages?

faustinoaq:
@drujensen Yeah, we can remove `CssoWebpackPlugin` but we lose the capability of minify extracted CSS. I think minified css isn't too important here, developers can configure webpack on theirs own needs.

In the other hand, `ExtractTextPlugin` is the official plugin recommended by webpack to extract CSS, so we need this one. [0]

[0] - https://webpack.js.org/plugins/extract-text-webpack-plugin/#usage-example-with-css

drujensen:
@faustinoaq thanks.  lgtm

PR #302, opened 2017-10-12
Adds environment boolean checks

eliasjpr:
Amber currently ships with 3 different environments: development, test
and production.

As a developer I would like to sometimes run code conditionally
depending on the environment that the application is.

```crystal
if Amber.env.development?
  # ... code here ...
end
```

This PR addresses the issue by creating a Amber::Env module that
contains 3 module methods:

```crystal
Amber.env.development?
Amber.env.test?
Amber.env.production?
Amber.env.integration?
Amber.env.staging?
Amber.env.sandbox?

Amber.env.in? %w(development test)
Amber.env.in? %i(development test)

Amber.env.is? "test"
Amber.env.is? :test
```
With the changes in this PR a developer should be able to call the
appropriate method to check for the desired environment.

### DIFF
--- a/src/amber/server/env.cr
+++ b/src/amber/server/env.cr
@@ -1,0 +1,34 @@
+module Amber
+  def self.env
+    Env.new(Settings.env.not_nil!.downcase)
+  end
+
+  class Env
+    def initialize(@env : String | Symbol)
+    end
+
+    def in?(environment_list : Array(String | Symbol))
+      (environment_list.map &.to_s.downcase).includes? @env
+    end
+
+    def is?(environment : String | Symbol)
+      @env == environment.to_s.downcase
+    end
+
+    def to_s(io)
+      io << @env
+    end
+
+    private def environments
+      (Dir.entries(CONFIG_DIR).map &.downcase.tr(".yml", "") + ENVIRONMENTS).to_set
+    end
+
+    macro method_missing(call)
+      environment = {{call.name.id.stringify.downcase}}
+      if environment.chars.last == '?'
+        environment  = environment.downcase.tr("?", "")
+        (@env == environment)
+      end
+    end
+  end
+end
--- a/spec/amber/amber_spec.cr
+++ b/spec/amber/amber_spec.cr
@@ -6,14 +6,13 @@ struct RoomSocket < Amber::WebSockets::ClientSocket; end
 
 describe Amber::Server do
   describe ".configure" do
-    it "it loads environment settings from test.yml" do
+    it "loads environment settings from test.yml" do
       settings = Amber::Server.settings
 
       settings.name.should eq "amber_test_app"
       settings.port_reuse.should eq true
       settings.redis_url.should eq "#{ENV["REDIS_URL"]? || "redis://localhost:6379"}"
       settings.port.should eq 3000
-      settings.env.should eq "test"
       settings.color.should eq true
       settings.secret_key_base.should eq "mV6kTmG3k1yVFh-fPYpugSn0wbZveDvrvfQuv88DPF8"
       # Sometimes settings get over written by other tests first and this fails
@@ -26,7 +25,7 @@ describe Amber::Server do
       settings.secrets.should eq expected_secrets
     end
 
-    it "allows you to overide enviroment settings" do
+    it "overrides enviroment settings" do
       Amber::Server.configure do |server|
         server.name = "Hello World App"
         server.port = 8080
@@ -45,7 +44,7 @@ describe Amber::Server do
       settings.secret_key_base.should eq "mV6kTmG3k1yVFh-fPYpugSn0wbZveDvrvfQuv88DPF8"
     end
 
-    it "should still retain environment.yml settings that haven't been overwritten" do
+    it "retains environment.yml settings that haven't been overwritten" do
       # NOTE: Any changes to settings here remain for all specs run afterwards.
       # This is a problem.
 
--- a/spec/amber/cli/commands/generator_spec.cr
+++ b/spec/amber/cli/commands/generator_spec.cr
@@ -5,6 +5,8 @@ module Amber::CLI
     describe MainCommand::Generate do
       context "scaffold" do
         it "generates and compile generated app" do
+          ENV["AMBER_ENV"] = "test"
+
           MainCommand.run ["new", TESTING_APP]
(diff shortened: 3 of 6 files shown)


elorest:
Could we add it directly to Amber like it is in Rails. So `Amber.env` `Amber.development?` etc.

eliasjpr:
I considered `Amber.development?` and do like the aesthetics of this, but then I thought that if we move the check all the way, would people expect all the settings to live in this scope? So I opted to be safe by introducing the `Amber::Env` I consider the module approach to be more appropriate `Amber::Env` since it allow us to extend the module we can more methods and stuff pertinent to the Env here if we wish to in the future. 

Also I believe the method chaining in rails `Rails.env.development?` violates some principles, `law of demeter` for instance. Also I find it counter intuitive that rails returns the current environment if you just do `Rails.env` but it returns a bool check if you do `Rails.env.development?` I would expect something like `Rails.env.current`, which if we have to do something similar it could look something like `Amber::Env.current`

elorest:
I agree we shouldn't mimic or copy rails in a lot of things, but creating a whole new module to return `ENV["AMBER_ENV"] == "development"` etc seems like unnecessary complexity. Rails.env automatically calls to_s. This is the same for every instance in crystal and ruby.

https://carc.in/#/r/2w68

veelenga:
@eliasjpr looks nice 👍 

What about user-defined environments? `staging`, `integration`, `sandbox`, `beta` etc. often can be seen in web apps. I understand what it is not possible to dynamically generate `.staging?` method, but it could be good to have some kind of fallback for this, for example, `Amber::Env.is? "staging"` or something...

drujensen:
@veelenga really like the `Amber::Env.is? :staging` idea.

veelenga:
Another use case might be to check for inclusion in a list of environments. 

`Amber::Env.in? %w(staging integration sandbox)` WDYT?

eliasjpr:
I have added the `in?` and `is?` methods as suggested @veelenga Thanks, this is a great addition

It would be nice to populate the `ENVIRONMENTS` constant at precompile time appending the contents of the `./config/environments` files present. This way the `ENVIRONMENTS` will contain the all the environments defined. 

I attempted to approach the problem with `method_missing` and converting the `Env` to a class, so the call would look something like `Amber::Env.new.development?`. `method_missing` will not work in modules, the macro is invoked when a method call on an instance fails, modules cannot be instantiated.

I'm going to look at the failing specs.

src/amber/server/env.cr
```diff
@@ -0,0 +1,27 @@
+module Amber
+  module Env
+    ENVIRONMENTS = %w(development test production)
```
elorest: These could be dynamically defined at compile time by looking at file names in `./config/environments`

src/amber/server/env.cr
```diff
@@ -0,0 +1,27 @@
...
+    def self.in?(environments : Array(String | Symbol))
+      (ENVIRONMENTS.map &.to_s.downcase).includes? current
+    end
+
+    def self.is?(environment : String | Symbol)
+      current == environment.to_s.downcase
+    end
+
+    def self.current
+      Settings.env.not_nil!.downcase
```
elorest: Store env here is just duplicating values already stored in ENV["AMBER_ENV"]. Is this necessary? I would rather `current` just wrap `ENV["AMBER_ENV"]` and return it's values.

src/amber/server/env.cr
eliasjpr: Do you have an example on how to do this? I got stuck...

src/amber/server/env.cr
eliasjpr: I would trust the Settings.env more than the ENV["AMBER_ENV"] since the latter can be modified outside of the APP after it has been started. What do you think?

elorest: ENV["AMBER_ENV"] is defined by the env var at runtime and can't be changed outside of the app. Since the env var is where this value originates I forsee some possible issues where they could get out of sync and cause confusion if they're not pointing t

src/amber/server/env.cr
elorest: Yeah. Give me a second.

src/amber/server/env.cr
eliasjpr: We should be fine using `Settings.env` since its pointing to `ENV["AMBER_ENV"]`. See https://github.com/amberframework/amber/pull/302/files#diff-b8648cef4f24d74b1b3b0cee94f1c9deR15

src/amber/server/env.cr
elorest: There are 2 ways to do this.

### Eval
This way requires less files but isn't quite as fast for compile time.

`ENVIRONMENTS = {{ system("crystal eval \'puts Dir.glob(\"./config/environments/**\").map(&.split(/[\\/\\.]/)[-2]?)\'") }}`

### New File using Macro Run

This way is a bit faster as the compile can cache the compiled `environment_names.cr`

```cr
# environment_names.cr
puts Dir.glob("./config/environments/**").map(&.split(/[\/\.]/)[-2]?)
```

Then call `ENVIRONMENTS = {{run("./path/to/environment_names.cr")}}` in env.cr.

src/amber/server/env.cr
elorest: Ok. I'm ok with that then.

eliasjpr:
Updates.

@elorest was right by using the Rails way it turned out to be very clean and feature rich 

kudos @elorest 🎉 

This now supports the following. 

```crystal
Amber.env.development?
Amber.env.test?
Amber.env.production?
Amber.env.integration?
Amber.env.staging?
Amber.env.sandbox?

Amber.env.in? %w(development test)
Amber.env.in? %i(development test)

Amber.env.is? "test"
Amber.env.is? :test
```

spec/amber/scripts/environment_loader_spec.cr
```diff
@@ -0,0 +1,47 @@
+require "../../../spec_helper"
+
+describe Amber::Server do
```
elorest: I like the idea of moving this under scripts. Would it be possible to move it and maintain git file history though instead of adding a new file?

eliasjpr: I can undo the commit, and you can go ahead and move it in a separate PR. WDYT?

PR #304, opened 2017-10-13
Lock shard.yml Amber version to version of cli that generated it.

elorest:
### Description of the Change

Locks Amber version in `shard.yml` to `Amber::VERSION` of CLI that generated it.


src/amber/cli/templates/app/shard.yml.ecr
```diff
@@ -18,6 +18,8 @@ targets:
 dependencies:
   amber:
     github: amberframework/amber
+    version: <%= Amber::VERSION %>
+    #branch: master
```
fridgerator: did you mean to leave this here?

PR #305, opened 2017-10-13
Added == and DRY'd up env

elorest:
### Description of the Change

1. Added `env.==(environment)` to Env.
2. Pointed every comparison to `.is?(environment)` so that the comparison logic is only defined once. 
3. Used `ends_with?('?')` in `method_missing` macro for speed increase.
4. Added Tuple params options for `.in?(:development, :test)`.
5. `in?` returns immediately upon true which results in less string conversions.
6. Tested `Amber::Env` class in a vacuum as well testing Amber.env.
7. Retains complete object functionality and adds `==`.


src/amber/server/env.cr
```diff
@@ -1,33 +1,38 @@
 module Amber
   def self.env
-    Env.new(Settings.env.not_nil!.downcase)
+    @@env ||= Env.new(ENV["AMBER_ENV"]? || "development") # Would be best to only define this once.
```
eliasjpr: I don’t understand the need for memorization here. How much significant is the speed gain?

src/amber/server/env.cr
```diff
@@ -1,33 +1,38 @@
...
+    def initialize(@env : String) end
+
+    def is?(environment : EnvType)
+      @env == environment.to_s.downcase
     end
 
-    def in?(environment_list : Array(String | Symbol))
-      (environment_list.map &.to_s.downcase).includes? @env
+    def in?(environment_list : Array(EnvType))
+      environment_list.each do |other_env|
```
eliasjpr: Could have just removed the map &.to_s.downcase since includes?, any? Does the same thing as above

src/amber/server/env.cr
```diff
@@ -1,33 +1,38 @@
...
+        return true if is?(other_env)
+      end
     end
 
-    def is?(environment : String | Symbol)
-      @env == environment.to_s.downcase
+    def in?(*environment_list : Object)
+      in?(environment_list.to_a)
```
eliasjpr: This is a nice addition

src/amber/server/env.cr
```diff
@@ -1,33 +1,38 @@
...
     end
 
     def to_s(io)
       io << @env
     end
 
-    private def environments
-      (Dir.entries(CONFIG_DIR).map &.downcase.tr(".yml", "") + ENVIRONMENTS).to_set
+    def ==(other : EnvType)
```
eliasjpr: Also a nice addition

src/amber/server/settings.cr
```diff
@@ -12,7 +12,7 @@ module Amber
     class_property port_reuse : Bool
     class_property log : ::Logger
     class_property color : Bool = true
-    class_property env = ENV["AMBER_ENV"]? || "development"
+    class_property env = ENV["AMBER_ENV"]? || "development" #This doesn't need to be here.
```
eliasjpr: Can we remove the inline comment and put it above the code. Also what exactly does not need to be there?

src/amber/server/env.cr
eliasjpr: Also each check is down casing the other_env which is more operations than downcasing everything at once. Again don’t understand the performance gain here

src/amber/server/env.cr
eliasjpr: The ENV[“AMBER_ENV”] || “development” this is already happening in Setting.env the reason why it was using the Setting.env. Also initializing a class like this is confusing. What is the use case for Setting.env now with this environment class. 

I would like to recommend to use a variable assignment 
Environment = ENV[“AMBER_ENV”] || “. And remove Settings.env

spec/amber/server/env_spec.cr
```diff
@@ -4,40 +4,85 @@ describe Amber do
...
       it "returns false when the environment does not match" do
-        Amber::Settings.env = "invalid environment"
-        Amber.env.{{env.id}}?.should be_falsey
+        amber_env = Amber::Env.new "invalid environment"
+        amber_env.{{env.id}}?.should be_falsey
       end
     end
   {% end %}
 
+  describe "Amber.env" do
```
eliasjpr: Can you move this out into its own test. This can be move to the amber_spec file. 

Describe Amber do
   Describe “.env” do 
      ....
   End 
End

src/amber/server/env.cr
```diff
@@ -1,33 +1,38 @@
...
+      is?(other)
     end
 
     macro method_missing(call)
-      environment = {{call.name.id.stringify.downcase}}
-      if environment.chars.last == '?'
-        environment  = environment.downcase.tr("?", "")
-        (@env == environment)
+      if (environment = {{call.name.id.stringify}}).ends_with? '?'
```
eliasjpr: Inline assignment I’m against it stylistically, since it can escape your eyes if you are doing an assignment or comparison. I’m ruby rubicon will complain about this ñ, but this is not ruby. 

End_width kudos!!

eliasjpr: Rubocop

src/amber/server/settings.cr
eliasjpr: In fact we can get rid of this line completely

veelenga:
@elorest i think `is?` is useless if there is `==`. And it violates Crystal best practices regarding alias methods :)

src/amber/server/env.cr
```diff
@@ -1,33 +1,38 @@
...
+    @@env ||= Env.new(ENV["AMBER_ENV"]? || "development") # Would be best to only define this once.
   end
 
   class Env
-    def initialize(@env : String | Symbol)
+    alias EnvType = String | Symbol 
+
+    def initialize(@env : String) end
+
+    def is?(environment : EnvType)
```
eliasjpr: I think we can remove this method with the new addition of the `==`

src/amber/server/env.cr
eliasjpr: You can remove `: Object`

elorest: If you remove `: Object` it will respond to nil i.e, `env.in?()` rather than throwing a method not found. This will also cause it to error when I convert the tuple to an array.

src/amber/server/env.cr
elorest: Good point.

src/amber/server/env.cr
elorest: With this I didn't want to change any part of the outside of the blackbox. I'm fine with removing this but sort of thought of it last night as [] and fetch in a hash.
https://github.com/crystal-lang/crystal/blob/d1f8c42f82d3e6491cb7b3b536984c19a51b94ba/src/hash.cr#L59

spec/amber/server/env_spec.cr
elorest: Yeah.

src/amber/server/settings.cr
elorest: That's my plan but I thought maybe we should discuss it first.

src/amber/server/env.cr
elorest: Again I'm fine with removing. You guys had wanted it yesterday though so I didn't want to just go delete it. Also I sort of though of it as [] and fetch in hash. https://github.com/crystal-lang/crystal/blob/d1f8c42f82d3e6491cb7b3b536984c19a51b94ba/src/hash.cr#L59

src/amber/server/env.cr
elorest: In the past `&.map.to_s.downcase` was also modifying the strings one at a time. In this case we're just relying on is? or == to do the to_s.downcasing which is DRYer. Also in the case that the first one matches it doesn't go on to the second or third and no string operations are performed.

src/amber/server/env.cr
elorest: In terms of garbage collection we're potentially saving a lot of time here depending on how many times Amber.env gets called.

elorest: I think we're on the same page as far as making this the definitely location of env instead of Settings.

elorest:
"Remember that ENV[“AMBER_ENV”] reads the system environment variable that can change from outside the app."
- After run time ENV values don't change unless we change them in the ENV hash manually.

"Maybe we should just use ENV[“AMBER_ENV”] inside the env class"
- Probably a good idea.

"Beware that every time you are calling is? Is doing a downcase, specially in loops."
- I'm aware of this. Before it was just calling to_s.downcase manually at every place. Now it gets called at most the same amount of times, with a potential for less calls.

src/amber/server/env.cr
elorest: WDYT?

src/amber/server/env.cr
eliasjpr: What do you think of 

```crystal
environment_list.any? { |env| self.== env.to_s.downcase }
```

src/amber/server/env.cr
eliasjpr: So by passing `: Object` it kinda makes the argument `required`, interesting. 

I would think the user of the method is responsible to not pass nil. Do you have an example of your comment above I can be learning something new today 😄

src/amber/server/env.cr
veelenga: @elorest you made a good improvement on this. Yesterday it was `Amber::Env.is?`, but now it is `Amber.env ==` which is much more natural 🎉

Besides, personally I'm indifferent to `[]` approach.

eliasjpr: I think we can just remove it. We covered all the current use cases. Less is more in this case.

eliasjpr:
I will thumbs up after requested changes

src/amber/server/env.cr
elorest: Does any? return as soon as it finds one?

elorest:
Thanks for your review. I'm moving offices today so might not be able to fix until later on. Have a good day.

src/amber/server/env.cr
eliasjpr: Yes it exits when the first instance  is found. See https://github.com/crystal-lang/crystal/blob/e2a1389e8165fb097c785524337d7bb7b550a086/src/enumerable.cr#L68

elorest: Perfect. Thanks.

elorest:
@eliasjpr Should be done. Let me know what you think.

PR #307, opened 2017-10-16
Fixes bug with release being undefined and adds --no-debug to production run

elorest:
### Description of the Change

This allows it to work with LLVM 4.0 on linux.
Also saves a bit of time on compiling.
Fixes bug where if run command was used for development or test it would error because #{release} didn't exist.
Fixes issue https://github.com/amberframework/amber/issues/308

### DIFF
--- a/src/amber/cli/commands/run.cr
+++ b/src/amber/cli/commands/run.cr
@@ -19,9 +19,13 @@ module Amber::CLI
       end
 
       def run
-        release = "--release" if options.e.downcase == "production"
+        build_options = Array(String)
+        build_options << "--release" if options.e.downcase == "production"
+        build_options << "--no-debug" if options.e.downcase == "production"
+        build_options << " " if build_options.size > 0
+
         puts colorize("💎  Crystalizing...", :dark_gray)
-        `crystal build #{release} $(ls ./src/*.cr | sort -n | head -1) -o app`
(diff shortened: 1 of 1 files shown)


src/amber/cli/commands/run.cr
```diff
@@ -19,9 +19,13 @@ module Amber::CLI
       end
 
       def run
-        release = "--release" if options.e.downcase == "production"
+        build_options = Array(String).new
+        build_options << "--release" if options.e.downcase == "production"
+        build_options << "--no-debug" if options.e.downcase == "production"
+        build_options << " " if build_options.size > 0
```
veelenga: This may be a bit cleaner:

```crystal
if options.e.downcase == "production"
  build_options.concat ["--release", "--no-debug", " "]
end

```

elorest: I wrote it this way so that we could add different options for different reasons but only have a space if there was at least one option.

If it will never change we could just write `build_options += "--release --no-debug "`

veelenga: Right, but `.downcase` and `==` are done twice, what is not really good

eliasjpr: I agree with the above is much cleaner. I will also remove the additional space `" "` and just add the space on line 28  between the `} $`

elorest: But then it ends up with 2 spaces in the case that the env isn't production.

Issue #306, opened 2017-10-16
render text / plain text in controller response?


sam0x17:
What is the amber equivalent of `render text: 'asfasdf'` (or in newer rails: `render body: 'asdfasdf'`)? I need to render raw text as part of an API.

elorest:
```cr
def text_api
  "hello world!"
end
```

sam0x17:
omg why didn't rails do it this way so simple!! :+1:

elorest:
Yeah I love rails but this is definitely one of the things I wanted to do differently.

c910335:
And don't forget `response.content_type = "test/plain"` or other Content-Type you need.

sam0x17:
@c910335 is there a way I can do that in my base ApiController or do I need to do it in every single method. Also would super appreciate a JSON example?

c910335:
[Custom Pipe](https://amberframework.org/guides#getting-started/routing/pipelines.md) or [Filters](https://amberframework.org/guides#getting-started/controllers/filters.md) should be good for you.

elorest:
Yeah. Use just write a simple custom pipe to do that.

elorest:
For JSON just write 
```cr
def api_action
  {"name" => "elorest"}.to_json
end
```

Also make sure to set the response type somehow.

sam0x17:
excellent, thanks all

elorest:
@sam0x17 Here's an example handler/pipe you that would set content_types for you.

```cr
    class TextType
      include HTTP::Handler
      def call(context : HTTP::Server::Context)
        content.response.content_type = "test/plain"
        call_next context
      end
    end
```

sam0x17:
@elorest thanks!

watzon:
Don't know how anyone else feels about this, but for purposes of readability and ease of use it could be nice to have a similar functionality to Rails here by using the render method with a named `json` attribute. That way the render method could take the hash, variable, etc, apply `to_json` if necessary, set the content type, and render the json response.

drujensen:
@watzon something like this?  

https://github.com/kemalyst/kemalyst/blob/master/src/kemalyst/helpers.cr#L36

watzon:
@drujensen something like that would work. Did that code not get ported over when kemalyst became amber?

drujensen:
[...]

PR #313, opened 2017-10-18
Methods to set content type and content

elorest:
### Description of the Change

Resolves discussion in this issue. https://github.com/amberframework/amber/issues/306

Adds `respond_with` block as a helper method to set content_types.

Returning values and render by itself still work unless a value has already been set via these methods.

### Benefits

```cr
respond_with do
 html render("index.slang")
 json posts.to_json
 xml render("index.xml.slang")
 text "hello world"
end
```

Provides and easier and cleaner way to set content_types.

### DIFF
--- a/src/amber/controller/helpers/render.cr
+++ b/src/amber/controller/helpers/render.cr
@@ -37,30 +37,5 @@ module Amber::Controller::Helpers
         Kilt.render("#{{{path}}}/{{filename.id}}")
       {% end %}
     end
-
-    protected def respond_with(html : String? = nil, json : Hash | String? = nil, xml : String? = nil, text : String? = nil)
-      puts context.request.headers["Accept"]
-      accepts = context.request.headers["Accept"].split(";").first.try(&.split(/,|,\s/))
-      # TODO: add JS type simlar to rails.
-      if accepts.includes?(Content::TYPE[:html]) && html
-        respond_with_html(html)
-      elsif accepts.includes?(Content::TYPE[:json]) && json
-        respond_with_json(json.is_a?(Hash) ? json.to_json : json)
-      elsif accepts.includes?(Content::TYPE[:xml]) && xml
-        respond_with_xml(xml)
-      elsif accepts.includes?(Content::TYPE[:text]) && text
-        respond_with_text(text)
-      elsif html
-        respond_with_html(html)
-      elsif text
-        respond_with_text(text)
-      elsif json
-        respond_with_json(json.is_a?(Hash) ? json.to_json : json)
-      elsif xml
-        respond_with_xml(xml)
-      else
-        respond_with_text("Response not acceptable", 406)
-      end
-    end
   end
 end
--- a/src/amber/controller/helpers/render.cr
+++ b/src/amber/controller/helpers/render.cr
@@ -1,0 +1,66 @@
+module Amber::Controller::Helpers
+  module Render
+    LAYOUT = "application.slang"
+
+    macro render(template = nil, layout = true, partial = nil, path = "src/views", folder = __FILE__)
+      {% if !(template || partial) %}
+        raise "Template or partial required!"
+      {% end %}
+
+      {{ filename = template || partial }}
+
+      # Render Template and return content
+      {% if filename.id.split("/").size > 1 %}
+        %content = render_template("#{{{filename}}}", {{path}})
+      {% else %}
+        {{ short_path = folder.gsub(/^.+?(?:controllers|views)\//, "") }}
+        {% if folder.id.ends_with?(".ecr") %}
+          %content = render_template("#{{{path}}}/#{{{short_path.gsub(/\/[^\.\/]+\.ecr/, "")}}}/#{{{filename}}}")
+        {% else %}
+          %content = render_template("#{{{path}}}/#{{{short_path.gsub(/\_controller\.cr|\.cr/, "")}}}/#{{{filename}}}")
+        {% end %}
+      {% end %}
+
+      # Render Layout
+      {% if layout && !partial %}
+        content = %content
+        render_template("#{{{path}}}/layouts/#{{{layout.class_name == "StringLiteral" ? layout : LAYOUT}}}")
+      {% else %}
+        %content
+      {% end %}
+    end
+
+    private macro render_template(filename, path = "src/views")
+      {% if filename.id.split("/").size > 2 %}
+        Kilt.render("{{filename.id}}")
+      {% else %}
+        Kilt.render("#{{{path}}}/{{filename.id}}")
+      {% end %}
+    end
+
+    protected def respond_with(html : String? = nil, json : Hash | String? = nil, xml : String? = nil, text : String? = nil)
+      puts context.request.headers["Accept"]
+      accepts = context.request.headers["Accept"].split(";").first.try(&.split(/,|,\s/))
+      # TODO: add JS type simlar to rails.
+      if accepts.includes?(Content::TYPE[:html]) && html
+        respond_with_html(html)
+      elsif accepts.includes?(Content::TYPE[:json]) && json
+        respond_with_json(json.is_a?(Hash) ? json.to_json : json)
+      elsif accepts.includes?(Content::TYPE[:xml]) && xml
+        respond_with_xml(xml)
+      elsif accepts.includes?(Content::TYPE[:text]) && text
+        respond_with_text(text)
+      elsif html
+        respond_with_html(html)
+      elsif text
+        respond_with_text(text)
+      elsif json
+        respond_with_json(json.is_a?(Hash) ? json.to_json : json)
+      elsif xml
+        respond_with_xml(xml)
+      else
+        respond_with_text("Response not acceptable", 406)
+      end
+    end
+  end
+end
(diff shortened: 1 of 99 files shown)


watzon:
Very happy to see this being worked on so quickly

src/amber/controller/render.cr
```diff
@@ -2,6 +2,24 @@ module Amber::Controller
   module Render
     LAYOUT = "application.slang"
 
+    protected def render_text(body, status_code = 200)
```
drujensen: @elorest what do you think of dropping the `render_` and just using `text`, `json`, `html`?

c910335: I think it would be better if their names are verbs, since verbs are more readable and nouns may collide with local variables.

drujensen: yup, your right.  It reads better.

elorest: @drujensen Which ones. I just thought that text, json, html might be more likely to collide with something else.

watzon: I don't know how you all feel about it, but I like the idea of just using the normal `render` command with named params like Rails does.

```crystal
def index
    render json: { key: "Value }
end
```

elorest: IMHO render is already complicated enough. Also text and json aren't truly rendered, but rather just copied over into the response without changes. As these helper methods currently  are they don't have to be macro's or complicate the already complex workings of the render macro.

elorest: Do you think there's a better word like `set_text, set_html, set_json`?

watzon: The verb could be `respond_with` instead of render? Maybe something like:

```crystal
def respond_with(json: Hash(String | Symbol, Amber::Router::ParamsType))
    ...
end

def respond_with(text: String)
    ...
end

def respond_with(html: XML | String)
    ...
end
```

Or something like that.

elorest: I'm ok with that. @drujensen @eliasjpr @fridgerator WDYT?

elorest: I'm ok with that. @drujensen @eliasjpr @fridgerator WDYT?

drujensen: @watzon that looks like a good solution.  

Should it support the ability to detect the `Accept` header and allow you to define multiple response types?
```
respond_with(
  json: {"hello" => "world"},
  text: "hello world", 
  html: render("index.slang")
)
```

eliasjpr: I like the signature above is very intuitive. Crystal supports defining method with external names so the above can be written as

```crystal
respond with: { "json" }

def respond(with type)
  ...
end
```

See External Names here https://crystal-lang.org/docs/syntax_and_semantics/default_values_named_arguments_splats_tuples_and_overloading.html

elorest: Yeah it's on my crystal lang coffee mug lol.

@eliasjpr In your code block above there's nothing that defines which type it's responding with though.

I think that this makes more sense.

```cr
respond_with(
  json: {"hello" => "world"},
  text: "hello world", 
  html: render("index.slang")
)
```

eliasjpr:
Was thinking about the new methods, they dont seem to be calling the render. What do you think of organizing them in such:

```crystal
module Render
  module Helpers
    def render_text end;
    def render_html end;
    def render_json end;
  end 
end 
```

And include them in the controllers...

elorest:
Could you look over this and let me know if you see any issues? @drujensen @eliasjpr 
I like that it's simple and easy to use, not sure if the accepts logic is missing anything though.

elorest:
@eliasjpr There is already a module called render. I would just prefer to leave them in there so they're included with the render macros. We can talk about it though. I'm more concerned with the functionality right now.

src/amber/controller/render.cr
```diff
@@ -37,5 +37,43 @@ module Amber::Controller
         Kilt.render("#{{{path}}}/{{filename.id}}")
       {% end %}
     end
+
+    protected def respond_with(html : String? = nil, json : Hash | String? = nil, xml : String? = nil, text : String? = nil)
+      accepts = context.request.headers["Accept"].split(";").try(&.split(","))
```
eliasjpr: There should probably be a guard check for when Accept is `*/*`. also sometimes the accept header could be in this format`text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8` Also we have some helper methods that you can make use of in this case, see https://github.com/amberframework/amber/blob/master/src/amber/support/mime_types.cr

elorest: @eliasjpr Thanks. I'm thinking that we should probably just default to html or text or something it doesn't match any of accept types. Seems like pretty much all browsers have */* as the last set of accepts parameters.

How would you suggest the mime_type helpers could be used: It seems like `Amber::Support::Mime.mime_type("json")` is more complicated than `"application/json"`. Since all of the ones we care about only point to one mime type I feel that it's probably bests just to hardcode the string. Thoughts?

elorest:
@drujensen @eliasjpr What would the expectation be when accepts and available response types don't match up? For instance it asks for `text` but the only thing available on the action is `html`

src/amber/controller/render.cr
eliasjpr: Well, the Amber::Support::Mime has a Hash of mime-types you can probably write a nice matcher method in that module to match the one from the header, this will keep the code centralized in one place instead of hardcoding the string.

eliasjpr:
According to  the RFC (https://tools.ietf.org/html/rfc7231#section-5.3.2) return 406 (Not Acceptable) response or disregard the header field by treating the response as if it is not subject to content negotiation.

drujensen:
My understanding (reading the spec) is that you should send a 406 (not acceptable) response.  

I don't know if you want to add this capability, but we could also support priority based on the `q` parameter.

`q` defaults to 1 so if none of the `Accept` content types has a `q`, then the order in which they are requested is the priority order.

If any of the `Accept` content types includes the `q` parameter, then we are supposed to use that as the priority order.

let me know if this is how you interpret it.  I'm ok if we ignore the `q` parameter.  I think this parameter is mainly used for audio/video quality.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1

drujensen:
nm, I was reading the older spec.  The newer spec:
```
Note: Use of the "q" parameter name to separate media type
      parameters from Accept extension parameters is due to historical
      practice. 
```
Looks like q is not recommended for media types.

elorest:
Thanks guys. I believe that's what I'm doing with the 406.

elorest:
It seems like everything after the first list semicolon is just png, jpg, */*. I can't see any reason why we should care about those.

src/amber/controller/render.cr
```diff
@@ -37,5 +37,43 @@ module Amber::Controller
         Kilt.render("#{{{path}}}/{{filename.id}}")
       {% end %}
     end
+
+    protected def respond_with(html : String? = nil, json : Hash | String? = nil, xml : String? = nil, text : String? = nil)
+      accepts = context.request.headers["Accept"].split(";").try(&.split(/,|,\s/))
+
+      if accepts.includes?("text/html") && html
```
eliasjpr: Can this be a constant if you want to keep the string as such:

```crystal
RESPON_WITH = { 
  html: "text/html", 
  json: "application/json", 
  text: "text/plain", 
  xml: "application/xml"
}
```

elorest: Yeah that looks good.

src/amber/controller/helpers/render.cr
```diff
@@ -37,5 +37,30 @@ module Amber::Controller
         Kilt.render("#{{{path}}}/{{filename.id}}")
       {% end %}
     end
+
+    protected def respond_with(html : String? = nil, json : Hash | String? = nil, xml : String? = nil, text : String? = nil)
```
eliasjpr: Can we remove this method?`responders.cr` does the format very nicely.

src/amber/controller/helpers/render.cr
```diff
@@ -37,5 +37,30 @@ module Amber::Controller
...
+      elsif html
+        respond_with_html(html)
+      elsif text
+        respond_with_text(text)
+      elsif json
+        respond_with_json(json.is_a?(Hash) ? json.to_json : json)
+      elsif xml
+        respond_with_xml(xml)
+      else
+        respond_with_text("Response not acceptable", 406)
```
drujensen: I tried `curl -H "Accept: application/pdf" http://localhost:3000` against this branch but for some reason the html was still returned instead of an error 406.

elorest: We decided that since all browsers include "*/*" as a final parameter that if nothing matched up just to return the first defined content type.

src/amber/controller/helpers/render.cr
elorest: Yeah. Weird that shouldn't be there.

elorest: Done!

elorest:
@drujensen It does allow for hashes and namedtuples right now. Just not nested.

watzon:
Yay! Great job guys!

Issue #310, opened 2017-10-17
Nested JSON, numbers, and boolean params being parsed as strings


watzon:
### Description

I have been trying for the past couple hours to get this working and finally figured it must be a bug. I am attempting to set up a route to receive a webhook from GitLab/GitHub. These webhooks looks something like this:

```
{
  "action": "opened",
  "issue": {
    "url": "https://api.github.com/repos/octocat/Hello-World/issues/1347",
    "number": 1347,
    ...
  },
  "repository" : {
    "id": 1296269,
    "full_name": "octocat/Hello-World",
    "owner": {
      "login": "octocat",
      "id": 1,
      ...
    },
    ...
  },
  "sender": {
    "login": "octocat",
    "id": 1,
    ...
  }
}
```

Each have several nested objects. This is something that shouldn't be an issue. The problem is that all nested objects get parsed into a string, so the above example becomes this:

```
{
  "action": "opened",
  "issue": "{
    \"url\": \"https://api.github.com/repos/octocat/Hello-World/issues/1347\",
    \"number\": \"1347\",
    ...
  }",
  "repository" : "{
    \"id\": \"1296269\",
    \"full_name\": \"octocat/Hello-World\",
    \"owner\": \"{
      \"login\": \"octocat\",
      \"id\": \"1\",
      ...
    }",
    ...
  },
  "sender": \"{
    \"login\": \"octocat\",
    \"id\": \"1\",
    ...
  }"
}
```

Numbers are also becoming strings which is a slightly smaller problem.

### Steps to Reproduce

1. Create a `POST` route and route it to a controller.
2. In the controller method you're routing to put `pp params.raw_params` and return something.
3. Go to GitHub/GitLab webhooks and create a webhook (for this method you will need ngrok or something. Alternately you could just use Postman or the like.
4. Sent a test hook, or if you went with the Postman method send a JSON payload with nested objects.

**Expected behavior:** Params should be parsed correctly as a hash. Nested objects should become hashes as well rather than strings. Numbers should be numbers if they were sent as numbers, not strings. Booleans should also be boolean values.

**Actual behavior:** Nested objects, numbers, and booleans get stringified.

**Reproduces how often:** Every time.

### Versions

Amber CMD (amberframework.org) - v0.3.0
Windows Subsystem for Linux

### Additional Information

This is most likely due to `raw_params` being typed as `Hash(String, String)`, at least that is my thought. I don't know if there's a way around this, but I feel like there should be or this makes things super difficult to use.

[...]

PR #315, opened 2017-10-19
Fix Params JSON Parsing

eliasjpr:
#### Depends on https://github.com/amberframework/granite-orm/pull/78 
☝️ (This needs to be merged in order for this branch to pass)

The framework should be able to parse JSON request body appropriately by returning a `JSON::Any` object.

Currently, there is a bug when trying to parse JSON parameters from the body of a request where nested JSON payload is not being parsed correctly

```crystal
# Given a JSON payload 
{ "test": "test", "address": { "city": "New York" }
# Should be accessed 
params["address"].as(Hash)["city"]
```
Params is now a custom Hash of the following signature `Hash(String | Symbol, Amber::Router::ParamsType)` or `Amber::Router::ParamsHash.new`

Issue https://github.com/amberframework/amber/issues/310

## UPDATED:

To give more context all parameters where being parsed using the **HTTP::Params.parse** method which returns an object of type **Hash(String, Array(String))** this caused issues when parsing JSON params from the body, It turns out that nested JSON keys where being parsed as a string. Basically, all `params` values are currently returning as a string. 

With the changes presented in this PR users can now define model objects with JSON mapping that should be evaluated correctly based on the mapping defined. So for instance:

```crystal
params["user_id"].as(Int64) => 123
params["address"].as(Hash).each ...
params["user"].as(User) => User.name 
# or 
User.from_json(params["user"]) => User.name
```
See https://github.com/amberframework/amber/pull/315/files#diff-707c9269bb28ed65da046758b2df5c48R156 
✋ @watzon

### DIFF
--- a/spec/amber/amber_spec.cr
+++ b/spec/amber/amber_spec.cr
@@ -4,8 +4,7 @@ struct UserSocket < Amber::WebSockets::ClientSocket; end
 
 struct RoomSocket < Amber::WebSockets::ClientSocket; end
 
-
-describe Amber do 
+describe Amber do
   describe ".env" do
     it "should return test" do
       Amber.env.test?.should be_truthy
@@ -54,9 +53,10 @@ describe Amber do
         settings.secret_key_base.should eq "mV6kTmG3k1yVFh-fPYpugSn0wbZveDvrvfQuv88DPF8"
       end
 
-      it "retains environment.yml settings that haven't been overwritten" do # NOTE: Any changes to settings here remain for all specs run afterwards.
-        # This is a problem.
-
+      it "retains environment.yml settings that haven't been overwritten" do
+        # NOTE: Any changes to settings here remain for all specs run afterwards.
+        # Added for convenience until settings is change to an instances
+        # instead of singleton this is still an "a problem".
         Amber::Server.configure do |server|
           server.name = "Hello World App"
           server.port = 8080
--- a/spec/amber/router/context_spec.cr
+++ b/spec/amber/router/context_spec.cr
@@ -2,7 +2,7 @@ require "../../../spec_helper"
 
 describe HTTP::Server::Context do
   describe "#override_request_method!" do
-    describe "when X-HTTP-Method-Overrid is present" do
+    context "when X-HTTP-Method-Override is present" do
       it "overrides form POST method to PUT, PATCH, DELETE" do
         %w(PUT PATCH DELETE).each do |method|
           header = HTTP::Headers.new
@@ -147,23 +147,24 @@ describe HTTP::Server::Context do
   it "parses json hash" do
     headers = HTTP::Headers.new
     headers["Content-Type"] = "application/json"
-    body = "{\"test\":\"test\"}"
+    body = %({ "test": "test", "address": { "city": "New York" }})
     request = HTTP::Request.new("POST", "/", headers, body)
 
     context = create_context(request)
 
     context.params["test"].should eq "test"
+    context.params["address"].as(Hash)["city"].should eq "New York"
   end
 
   it "parses json array" do
     headers = HTTP::Headers.new
     headers["Content-Type"] = "application/json"
-    body = "[\"test\",\"test2\"]"
+    body = %(["test", "test2"])
     request = HTTP::Request.new("POST", "/", headers, body)
 
     context = create_context(request)
 
-    context.params["_json"].should eq "[\"test\", \"test2\"]"
+    context.params["_json"].should eq %w(test test2)
   end
 
   it "parses files from multipart forms" do
--- a/spec/amber/validations/params_spec.cr
+++ b/spec/amber/validations/params_spec.cr
@@ -4,7 +4,7 @@ module Amber::Validators
   describe BaseRule do
     describe "#apply" do
       it "raises error when missing field" do
-        params = HTTP::Params.parse("")
+        params = params_builder("")
         rule = BaseRule.new("field", "") do
           false
         end
@@ -15,7 +15,7 @@ module Amber::Validators
       end
 
       it "returns true for given block" do
-        params = HTTP::Params.parse("field=val")
+        params = params_builder("field=val")
         rule = BaseRule.new("field", "") do
           true
         end
@@ -24,7 +24,7 @@ module Amber::Validators
       end
 
       it "returns false for the given block" do
-        params = HTTP::Params.parse("field=val")
+        params = params_builder("field=val")
         rule = BaseRule.new("field", "") do
           false
         end
@@ -35,7 +35,7 @@ module Amber::Validators
 
     describe "#error" do
       it "returns default error message" do
-        params = HTTP::Params.parse("field=val")
+        params = params_builder("field=val")
         error_message = "Field field is required"
         rule = BaseRule.new("field", nil) do
           false
@@ -47,7 +47,7 @@ module Amber::Validators
       end
 
       it "returns the given error message" do
-        params = HTTP::Params.parse("field=val")
+        params = params_builder("field=val")
         error_message = "You must provide this field"
         rule = BaseRule.new("field", error_message) do
           false
@@ -63,17 +63,17 @@ module Amber::Validators
   describe OptionalRule do
     describe "#apply" do
       it "does not apply rule when field is missing" do
-        params = HTTP::Params.parse("")
+        params = params_builder("")
         error_message = "You must provide this field"
         rule = OptionalRule.new("field", error_message) do
-          false
+          true
         end
 
         rule.apply(params).should be_true
       end
 
       it "applies rule when field is present" do
-        params = HTTP::Params.parse("field=val")
+        params = params_builder("field=val")
         error_message = "You must provide this field"
         rule = OptionalRule.new("field", error_message) do
           true
@@ -87,12 +87,12 @@ module Amber::Validators
   describe Params do
     describe "#validation" do
       it "validates required field" do
-        http_params = HTTP::Params.parse("name=elias&last_name=perez&middle=j")
+        http_params = params_builder("name=elias&last_name=perez&middle=j")
         validator = Validators::Params.new(http_params)
 
         result = validator.validation do
-          required(:name) { |v| v.str? & !v.empty? }
-          required("last_name") { |v| v.str? & !v.empty? }
+          required(:name) { |v| !v.nil? }
+          required("last_name") { |v| !v.nil? }
         end
 
         validator.valid?.should be_true
@@ -102,12 +102,12 @@ module Amber::Validators
       context "optional params" do
         context "when missing" do
           it "does not validate optional field" do
-            http_params = HTTP::Params.parse("last_name=&middle=j")
+            http_params = params_builder("last_name=&middle=j")
             validator = Validators::Params.new(http_params)
 
             result = validator.validation do
-              optional(:name) { |v| v.empty? }
-              required("last_name") { |v| v.empty? }
+              optional(:name) { |v| !v.nil? }
+              required("last_name") { |v| !v.nil? }
             end
 
             validator.valid?.should be_true
@@ -117,15 +117,33 @@ module Amber::Validators
 
         context "when present" do
           it "validates optional field" do
-            http_params = HTTP::Params.parse("name=")
+            http_params = params_builder("name=")
             validator = Validators::Params.new(http_params)
 
             result = validator.validation do
-              optional(:name) { |v| !v.empty? }
+              optional(:name) { |v| v.nil? }
             end
 
             validator.valid?.should be_false
             validator.errors.size.should eq 1
+          end
+        end
+
+        context "casting" do
+          it "returns false for the given block" do
+            params = params_builder("name=john&number=1&price=3.45&list=[1,2,3]")
+            validator = Validators::Params.new(params)
+            validator.validation do
+              required(:name) { |f| !f.to_s.empty? }
+              required(:number) { |f| f.as(String).to_i > 0 }
+              required(:price) { |f| f.as(String).to_f == 3.45 }
+              required(:list) do |f|
+                list = JSON.parse(f.as(String)).as_a
+                (list == [1, 2, 3] && !list.includes? 6)
+              end
+            end
+
+            validator.valid?.should be_truthy
           end
         end
       end
@@ -133,24 +151,24 @@ module Amber::Validators
 
     describe "#valid?" do
       it "returns false with invalid fields" do
-        http_params = HTTP::Params.parse("name=john&last_name=doe&middle=j")
+        http_params = params_builder("name=john&last_name=doe&middle=j")
         validator = Validators::Params.new(http_params)
 
         validator.validation do
-          required("name") { |v| v.empty? }
-          required("last_name") { |v| v.str? & !v.empty? }
+          required("name") { |v| v.nil? }
+          required("last_name") { |v| !v.nil? }
         end
 
         validator.valid?.should be_false
       end
 
       it "returns false when key does not exist" do
-        http_params = HTTP::Params.parse("name=elias")
+        http_params = params_builder("name=elias")
         validator = Validators::Params.new(http_params)
         result = {"nonexisting" => {nil, "Param [nonexisting] does not exist."}}
 
         validator.validation do
-          required("nonexisting") { |v| v.str? & !v.empty? }
+          required("nonexisting") { |v| !v.nil? }
         end
 
         expect_raises Exceptions::Validator::InvalidParam do
@@ -159,12 +177,12 @@ module Amber::Validators
       end
 
       it "returns true with valid fields" do
-        http_params = HTTP::Params.parse("name=elias&last_name=perez&middle=j")
+        http_params = params_builder("name=elias&last_name=perez&middle=j")
         validator = Validators::Params.new(http_params)
 
         validator.validation do
-          required("name") { |v| v.str? & !v.empty? }
-          required("last_name") { |v| v.str? & !v.empty? }
+          required("name") { |v| !v.nil? }
+          required("last_name") { |v| !v.nil? }
         end
 
         validator.valid?.should be_true
@@ -173,12 +191,12 @@ module Amber::Validators
 
     describe "#validate!" do
       it "raises error on failed validation" do
-        http_params = HTTP::Params.parse("name=elias&last_name=perez&middle=j")
+        http_params = params_builder("name=&last_name=&middle=j")
         validator = Validators::Params.new(http_params)
 
         validator.validation do
-          required("name") { |v| v.str? & v.empty? }
-          required("last_name") { |v| v.str? & !v.empty? }
+          required("name") { |v| !v.to_s.empty? }
+          required("last_name") { |v| !v.to_s.empty? }
         end
 
         expect_raises Exceptions::Validator::ValidationFailed do
@@ -187,17 +205,27 @@ module Amber::Validators
       end
 
       it "returns validated params on successful validation" do
-        http_params = HTTP::Params.parse("name=elias&last_name=perez&middle=j")
+        http_params = params_builder("name=elias&last_name=perez&middle=j")
         validator = Validators::Params.new(http_params)
         result : Hash(String, String) = {"name" => "elias", "last_name" => "perez"}
 
         validator.validation do
-          required("name") { |v| v.str? & !v.empty? }
-          required("last_name") { |v| v.str? & !v.empty? }
+          required("name") { |v| !v.nil? }
+          required("last_name") { |v| !v.nil? }
         end
 
         validator.validate!.should eq result
       end
     end
+  end
+end
+
+def params_builder(body = "")
+  params = {} of String | Symbol => Amber::Router::ParamsType
+  return params if body.empty?
+
+  body.tr("?", "").split("&").each_with_object(params) do |item, params|
+    key, value = item.split("=")
+    params[key] = value
   end
 end
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -1,12 +1,13 @@
 # NOTE: Constants should be set before require begins.
 
 ENV["AMBER_ENV"] = "test"
-TEST_PATH    = "spec/support/sample"
-PUBLIC_PATH  = TEST_PATH + "/public"
-VIEWS_PATH   = TEST_PATH + "/views"
-TESTING_APP  = "./tmp/test_app"
-APP_TPL_PATH = "./src/amber/cli/templates/app"
-CURRENT_DIR  = Dir.current
+TEST_PATH     = "spec/support/sample"
+PUBLIC_PATH   = TEST_PATH + "/public"
+VIEWS_PATH    = TEST_PATH + "/views"
+TEST_APP_NAME = "test_app"
+TESTING_APP   = "./tmp/#{TEST_APP_NAME}"
+APP_TPL_PATH  = "./src/amber/cli/templates/app"
+CURRENT_DIR   = Dir.current
 
 require "http"
 require "spec"
--- a/src/amber/cli/templates/auth.cr
+++ b/src/amber/cli/templates/auth.cr
@@ -25,7 +25,7 @@ module Amber::CLI
       @timestamp = Time.now.to_s("%Y%m%d%H%M%S")
       @primary_key = primary_key
       @visible_fields = @fields.reject(&.hidden).map(&.name)
-      
+
       add_routes :web, <<-ROUTES
         get "/signin", SessionController, :new
           post "/session", SessionController, :create
@@ -37,7 +37,7 @@ module Amber::CLI
       add_plugs :web, <<-PLUGS
         plug Authenticate.new
       PLUGS
-      
+
       add_dependencies <<-DEPENDENCY
       require "../src/models/**"
       require "../src/handlers/**"
--- a/src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
+++ b/src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
@@ -19,7 +19,7 @@ class <%= @name.capitalize %>Controller < ApplicationController
   end
 
   def create
-    <%= @name %> = <%= @name.capitalize %>.new(params.to_h.select(<%= @visible_fields %>))
+    <%= @name %> = <%= @name.capitalize %>.new(<%= @name.downcase%>_params.validate!)
 
     if <%= @name %>.valid? && <%= @name %>.save
       flash["success"] = "Created <%= @name.capitalize %> successfully."
@@ -41,7 +41,7 @@ class <%= @name.capitalize %>Controller < ApplicationController
 
   def update
     if <%= @name %> = <%= @name.capitalize %>.find(params["id"])
-      <%= @name %>.set_attributes(params.to_h.select(<%= @visible_fields %>))
+      <%= @name %>.set_attributes(<%= @name.downcase%>_params.validate!)
       if <%= @name %>.valid? && <%= @name %>.save
         flash["success"] = "Updated <%= @name.capitalize %> successfully."
         redirect_to "/<%= @name %>s"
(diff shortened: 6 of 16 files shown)


watzon:
Just to clarify the bug affected/affects basically anything that's not a string; booleans, integers, floats, and nested objects.

elorest:
What needed fixed with param and form parameters?

eliasjpr:
Updated PR description

src/amber/router/params.cr
```diff
@@ -1,129 +1,4 @@
-require "json"
-
 module Amber::Router
-  # The Params module will parse parameters from a URL, a form post or a JSON
```
elorest: Why was everything but 4 lines deleted from this file and copied to params_parser?

src/amber/router/params.cr
```diff
@@ -1,129 +1,4 @@
...
-               else
-                 ""
-               end
-
-      HTTP::Params.parse(values) do |key, value|
-        params.add(key, value)
-      end
-    end
-  end
+  alias ParamsType = Nil | String | Int64 | Float64 | Bool | Hash(String, JSON::Type) | Array(JSON::Type)
```
elorest: I like this.

src/amber/router/params_parser.cr
```diff
@@ -0,0 +1,127 @@
+require "json"
```
elorest: I don't like creating a whole new file with mostly the same contents of params.cr. This screws up git history.

elorest:
It's still named, "Fixes Parsing URL, Form, JSON Parameters" but I'm unclear what has been changed with parsing URL and Form parameters.

elorest:
Now that params is a Hash instead of HTTP::Params we can include files in it as well which is nice.

src/amber/router/params.cr
eliasjpr: The rename of the file is necessary here since it adds context on what the file does.

src/amber/router/params_parser.cr
eliasjpr: I renamed the file to something more meaningful.

elorest:
The changes that affect the generators and add validation code into controllers and models should be a separate PR.

eliasjpr:
## Update Friday Oct 20 2017

Recently added another commit which addresses the issue with the generator_spec.cr:7 (Very useful spec).

The new changes modifies the app generator specifically the controller and models. 

**Controller**
The controller now generates the app using `params.validation` similar to Rails strong params. It adds a method to the Controller and overloads the Granite initializer and set_attributes methods. 

```crystal
# AnimalsController
private def animal_params
  params.validation do
    required("name") { |v| v.str? & !v.empty? }
  end
end
# animal_params.validate! returns a hash of `Hash(String | Symbol, Amber::Router::ParamsType)`
# only containing the attributes defined in the `params.validation`
animal = Animal.new(animal_params.validate!) 
```

**Model**
Models generate for the Granite::ORM now includes the following line `include Amber::Models::Granite` this will allow models to be populated with the controller params using the `params.validations`. The reason this is added here and not to the Granite::ORM repo is to keep Granite loosly couple from any framework, since the `params.validation` is specific to amber.

The module defines 2 methods an initializer and set attributes. See https://github.com/amberframework/amber/pull/315/files#diff-d748cfb908fdf84d3356f37cb13c2d0d 
 
```crystal
class Animal < Granite::ORM
  include Amber::Models::Granite
end
```

eliasjpr:
This is now passing here https://travis-ci.org/amberframework/amber/builds/290448518

src/amber/cli/templates/model/granite/src/models/{{name}}.cr.ecr
```diff
@@ -1,6 +1,7 @@
 require "granite_orm/adapter/<%= @database %>"
 
 class <%= @name.capitalize %> < Granite::ORM::Base
+  include Amber::Models::Granite
```
drujensen: yuck, don't like this at all.  Anyway of avoiding special Granite logic in Amber?

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -63,4 +63,14 @@ class <%= @name.capitalize %>Controller < ApplicationController
     end
     redirect_to "/<%= @name %>s"
   end
+
+  def <%= @name %>_params
+	params.validation do
+	  <% @fields.each do |field| %>
+	  <% unless field.cr_type() != "time" %>
+	  required(:<%= field.name %>) { |p| !p.as(<%= field.cr_type.capitalize %>) }
```
drujensen: I'm not following this logic.  Why would all the fields be required except times?

src/amber/models/granite.cr
```diff
@@ -0,0 +1,16 @@
...
+	  def initialize
+	  end
+
+	  def initialize(args : Hash(String | Symbol, Amber::Router::ParamsType))
+        set_attributes(args)
+  	  end
+
+	  def set_attributes(args : Hash(Symbol | String, Amber::Router::ParamsType))
+        args.each do |k, v|
+          cast_to_field(k, v.as(DB::Any))
```
drujensen: instead of doing this here, can we cast it in the controller before passing into Granite?  Also, how does Crecto handle this?

src/amber/cli/templates/model/granite/src/models/{{name}}.cr.ecr
eliasjpr: We can have the initializer and set_attributes defined in Granite. I thought of that, but this is specific to Amber.

This will clean everything

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
eliasjpr: This are the incoming URL/Form/Body params  that gets parsed. The time types are in reference to 'created_at' and 'updated_at' maybe it should be checking for those specific fields. 🤔

src/amber/models/granite.cr
eliasjpr: Well, the controller params does not care about the Model. The model should be responsible for casting to what it needs. 

Also the `cast_to_field` also live in Granite::ORM. We can easily just move the initialize and set_attributes to Granite

src/amber/cli/templates/model/granite/src/models/{{name}}.cr.ecr
drujensen: ok,  I'm torn.  This feels wrong to me but I see your argument for putting it here.  Are you ok with moving it to granite?

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
drujensen: yeah, I think you should iterate the `@visible_fields` since it takes care of which fields would be passed in the params.  There may be other fields that are `time` fields that are not created_at or updated_at.

src/amber/models/granite.cr
drujensen: I would prefer not requiring users to remember they need the included this module.  Can we move this to Granite?

eliasjpr:
@drujensen this was my initial thought. I have remove the granite  file from Amber and opened this PR in Granite-ORM

Depends on https://github.com/amberframework/granite-orm/pull/78 (This needs to be merged first in order for this branch to pass)

This actually worked quite well and is clean with no dependencies between each other.

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
eliasjpr: Unfortunately `@visible_fields` is an array of only names (String) and I need the type as well.

src/amber/router/params_parser.cr
```diff
@@ -1,28 +1,29 @@
 require "json"
 
 module Amber::Router
-  # The Params module will parse parameters from a URL, a form post or a JSON
+  alias ParamsType = Nil | String | Int64 | Float64 | Bool | Hash(String, JSON::Type) | Array(JSON::Type)
```
drujensen: `Float32 | Int32 | Time | Bytes`?

src/amber/cli/templates/auth.cr
```diff
@@ -25,7 +25,7 @@ module Amber::CLI
       @timestamp = Time.now.to_s("%Y%m%d%H%M%S")
       @primary_key = primary_key
       @visible_fields = @fields.reject(&.hidden).map(&.name)
```
drujensen: we should remove this if its no longer being used.

eliasjpr: **@visible_fields** is still being use in other places

spec/amber/router/context_spec.cr
```diff
@@ -147,23 +147,24 @@ describe HTTP::Server::Context do
...
     headers = HTTP::Headers.new
     headers["Content-Type"] = "application/json"
-    body = "[\"test\",\"test2\"]"
+    body = %(["test", "test2"])
     request = HTTP::Request.new("POST", "/", headers, body)
 
     context = create_context(request)
 
-    context.params["_json"].should eq "[\"test\", \"test2\"]"
+    context.params["_json"].should eq ["test", "test2"]
```
veelenga: Might be handy: `%w(test test2)`

spec/amber/validations/params_spec.cr
```diff
@@ -187,17 +187,25 @@ module Amber::Validators
...
     end
   end
 end
+
+def params_builder(body = "")
+  params = {} of String | Symbol => Amber::Router::ParamsType
+  HTTP::Params.parse(body).each do |key, value|
+    params[key] = value.as(Amber::Router::ParamsType)
+  end
+  params
```
veelenga: [Enumerable.each_with_object](https://crystal-lang.org/api/0.23.1/Enumerable.html#each_with_object%28obj%2C%26block%29-instance-method) returns that object, so you don't need that extra line to return it.

```crystal
params = {} of String | Symbol => Amber::Router::ParamsType

HTTP::Params.parse(body).each_with_object(params) do |key, value|
  params[key] = value.as(Amber::Router::ParamsType)
end
```

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -63,4 +63,14 @@ class <%= @name.capitalize %>Controller < ApplicationController
     end
     redirect_to "/<%= @name %>s"
   end
+
+  def <%= @name %>_params
+	params.validation do
+	  <% @fields.each do |field| %>
+	  <% unless ["created_at", "updated_at"].includes? field.name %>
```
veelenga: `%w(created_at updated_at).includes? ...`

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -63,4 +63,14 @@ class <%= @name.capitalize %>Controller < ApplicationController
...
+	  required(:<%= field.name %>) { |p| !p.as(<%= field.cr_type.capitalize %>) }
```
veelenga: Shouldn't it be `as?` instead of `as` ? `as` rises an exception which would result in 500 error code instead of validation error.

eliasjpr: I personally prefer the exception knowing that is not being cast to the type I am expecting, but it can be changed to `as?`, I guess  it depends in the use case. IMHO.

spec/amber/validations/params_spec.cr
eliasjpr: I tried this and got an error laving it as is for now

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
veelenga: @eliasjpr imagine a view with a form that requires integer value. User enters "blah" and submits. In my understanding controller will throw exception and user will be responded with status code 500, instead of a flash message "you have to enter integer value". Am I wrong?

spec/amber/validations/params_spec.cr
veelenga: Sorry, this is the correct one:

```crystal
params = {} of String | Symbol => Amber::Router::ParamsType

HTTP::Params.parse(body).each_with_object(params) do |(key, value), params|
  params[key] = value.as(Amber::Router::ParamsType)
end

```

eliasjpr:
@veelenga should be good now. Thanks for the nice suggestions!

veelenga:
@eliasjpr i still have doubts regrading `as` method. I compiled amber on this branch and generated new app:

```sh

$ amber new testapp
$ cd testapp
$ crystal deps
$ amber generate scaffold Post title:string no:integer
$ amber w 
```

First of all there is a strange formatting in PostController:

```crystal
  def post_params
	params.validation do
	  
	  
	  required(:title) { |p| !p.as(String) }
	  
	  
	  
	  required(:no) { |p| !p.as(Int32) }
	  
	  
	  
	  
	  
	  
	end
  end
```

Second of all is it does not compile:

```crystal
in src/controllers/post_controller.cr:22: instantiating 'post_params()'

  post = Post.new(post_params.validate!)
                  ^~~~~~~~~~~

in src/controllers/post_controller.cr:75: can't cast String to Int32

   required(:no) { |p| !p.as(Int32) }

```

I think it still requires some improvements/fixes.

eliasjpr:
Yes you’re right the post_params should not be, just params. I’ll clean/fix it up. Thanks

eliasjpr:
@veelenga I completely had a misconception of `as()` I thought this was a way of casting

veelenga:
@eliasjpr it is a way of casting, but with some type inference features:

```crystal
"string".as(Int32) #=> can't cast String to Int32 (compile error)
```

```crystal
if false
  a = 1
else
  a = "hello"
end

a.as(String) #=> ok 
a.as(Int32)  #=> cast from String to Int32 failed (runtime error, TypeCaseError)

```

eliasjpr:
@veelenga yeah that looks more like unboxing. 

Anyways I made changes to this PR. The only workaround I found for the moment was to parse the JSON inside the validation predicate in params.validation. I also had to open a PR in Granite::ORM to add more support for JSON types. https://github.com/amberframework/granite-orm/pull/82

Ideally for the params we should create something similar to JSON.mapping so we can map values pre compilation and properly cast the values for each params. The benefits of this is that it will have strict type params hence stronger validations and speed since it does not need to go through the JSON.parse.

To have this branch pass it is necessary to merge this PR https://github.com/amberframework/granite-orm/pull/82

spec/spec_helper.cr
```diff
@@ -1,12 +1,13 @@
...
-TEST_PATH    = "spec/support/sample"
-PUBLIC_PATH  = TEST_PATH + "/public"
-VIEWS_PATH   = TEST_PATH + "/views"
-TESTING_APP  = "./tmp/test_app"
-APP_TPL_PATH = "./src/amber/cli/templates/app"
-CURRENT_DIR  = Dir.current
+TEST_PATH     = "spec/support/sample"
+PUBLIC_PATH   = TEST_PATH + "/public"
+VIEWS_PATH    = TEST_PATH + "/views"
+TEST_APP_NAME = "test_app"
```
elorest: Why was this changed?

src/amber/cli/templates/auth.cr
elorest: @visible_fields is being used everywhere forms or controllers are being generated.

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -19,7 +19,7 @@ class <%= @name.capitalize %>Controller < ApplicationController
   end
 
   def create
-    <%= @name %> = <%= @name.capitalize %>.new(params.to_h.select(<%= @visible_fields %>))
+  <%= @name %> = <%= @name.capitalize %>.new(<%= @name.downcase%>_params.validate!)
```
elorest: Does `validate!` take into affect visible fields?

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -63,4 +63,26 @@ class <%= @name.capitalize %>Controller < ApplicationController
     end
     redirect_to "/<%= @name %>s"
   end
+
+  def <%= @name %>_params
+    params.validation do
+      <%- @fields.each do |field| -%>
+      <%- unless %(created_at updated_at).includes? field.name -%>
```
elorest: This conditional doesn't seem to have a body to execute if true.

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -63,4 +63,26 @@ class <%= @name.capitalize %>Controller < ApplicationController
...
+      <%- if %w(Int32 Int64).includes? field.cr_type -%>
+      required(:<%= field.name %>) { |p| !JSON.parse(p.as(String)).as_i.nil? }
+      <%- elsif %w(Bool).includes? field.cr_type -%>
+      required(:<%= field.name %>) { |p| JSON.parse(p.as(String)).as_bool }
```
elorest: Why is everything being parsed as JSON? Most params won't be JSON.

eliasjpr:
@elorest I would love to pair on this. I got burned out over the weekend on this. 

All params are now a Hash of ParamsType but most of them will be strings when parsed from the request, This doesn't allow to cast to the correct type. For instance `params["name"].as(Int32)` where name is a String, this throws an error because at the moment it was populated it was set as string. Using JSON.parse allows to cast to the correct type. 

> It would be ideal to be able to map the params to their correct type so when the request gets parsed there is no need for casting.

I understand that this does not look pretty `JSON.parse(p.as(String)).as_f.nil?` but is the only way to cast between types to the correct right now in Crystal.  

The PR looks big, not intended. I hope there is a better shorter way to do type casting in crystal but it seems that is not supported by the language at the moment.

watzon:
Looks like I really opened a can of worms didn't I? Haha. That's one of the things that I feel like Crystal really needs, a [martialing system[(https://golang.org/pkg/encoding/json/) like that of golang.

src/amber/router/params_parser.cr
```diff
@@ -1,28 +1,29 @@
 require "json"
 
 module Amber::Router
-  # The Params module will parse parameters from a URL, a form post or a JSON
+  alias ParamsType = Float32 | Int32 | Time | Bytes | Nil | String | Int64 | Float64 | Bool | Hash(String, JSON::Type) | Array(JSON::Type)
```
elorest: alias ParamsType = (String | JSON::Type)

src/amber/router/params_parser.cr
```diff
@@ -90,10 +91,10 @@ module Amber::Router
           case json = JSON.parse(body).raw
           when Hash
             json.each do |key, value|
-              params[key.as(String)] = value.to_s
+              params[key] = value.as(Amber::Router::ParamsType)
```
elorest:
```cr
    def parse_json
      if body = request.body.not_nil!.gets_to_end
        if body.size > 2
          case json = JSON.parse(body).raw
          when Hash
            json.each do |key, value|
              case value.raw.class
              when String
                params[key.as(String)] = value.to_s
              else
                params[key.as(String)] = value
              end
            end
          when Array
            params["_json"] = json.to_s
          end
        end
      end
    end
```

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
eliasjpr: I will be removing this `{ |p| !JSON.parse(p.as(String)).as_i.nil? }` to this `{ |p| !p.nil? }`

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
eliasjpr: The intention is not to render the `created_at` and `updated_at` for params validations

spec/spec_helper.cr
eliasjpr: We talked offline. 

This is to avoid the `split('/').last` on the test. See https://github.com/amberframework/amber/pull/315/files#diff-3269a1cd7605965238570f1b2aafc0a7L16

src/amber/router/params_parser.cr
```diff
@@ -90,10 +91,10 @@ module Amber::Router
           case json = JSON.parse(body).raw
           when Hash
             json.each do |key, value|
-              params[key.as(String)] = value.to_s
+              params[key.as(String)] = value
```
elorest: If JSONType is a string I think we should set it as a string otherwise just set it to value. WDYT?

elorest:
```cr
          when Hash
            json.each do |key, value|
              case value.raw.class
              when String
                params[key.as(String)] = value.to_s
              else
                params[key.as(String)] = value
              end
            end
          when Array
            params["_json"] = json.to_s
          end
```

eliasjpr: JSON::Parse returns always a  JSON::Any

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
drujensen: Seems like we would want this indented 2 spaces

src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -63,4 +63,14 @@ class <%= @name.capitalize %>Controller < ApplicationController
     end
     redirect_to "/<%= @name %>s"
   end
+
+  def <%= @name %>_params
+    params.validation do
+      <%- @fields.each do |field| -%>
+      <%- unless %(created_at updated_at).includes? field.name -%>
+      required(:<%= field.name %>) { |p| p.to_s.empty? }
```
drujensen: This is failing for checkboxes that may not send a value in the params.

```bash
amber g scaffold post title:string body:text published:bool
```

When trying to create a new blog post, I get an error:
```
ERROR: Validation failed. [Amber::Validators::Error(@param="title", @value="This is a test", @message="Field title is required"), Amber::Validators::Error(@param="body", @value="Testing", @message="Field body is required"), Amber::Validators::Error(@param="published", @value="0", @message="Field published is required")] (Amber::Exceptions::Validator::ValidationFailed)
```

src/amber/router/params_parser.cr
elorest: Yeah. That's why I'm saying that if it's a string instead of a hash we should just use the string value as the value instead of json::any. If it's anything except the string then just set it to the Json any value. See above.

eliasjpr: screen shot 2017-10-25 at 11 11 13 am

eliasjpr:
@drujensen and @elorest I have tested this generating a new test app with auth and model. This were the steps taken. This updates the Auth templates as well

1. make (Compile amber cli from Params branch to ensure is using the correct code)
2. amber n test_app
3. cd test_app
4. amber g scaffold Post title:string body:text published:bool
5. amber g auth user
6. Edit database.yml 
7. amber m up
8. shards install
9. amber w
10. Visit http://0.0.0.0:3000
11. Create account
12. Sign In
13. Create Post
14. Edit Post

src/amber/router/params_parser.cr
```diff
@@ -106,7 +107,7 @@ module Amber::Router
         if filename.is_a?(String) && !filename.empty?
           files[upload.name] = Files::File.new(upload: upload)
```
eliasjpr: For a separate PR this should be move to the params scope beware that the filename can collide with a parameter found in the body or url. Also `ParamsType` would need to include the union `Files::File`

elorest: `upload.name` is just an a file input name. So sure it could collide but so could any other input field.

spec/amber/amber_spec.cr
```diff
@@ -55,8 +54,6 @@ describe Amber do
       end
 
       it "retains environment.yml settings that haven't been overwritten" do # NOTE: Any changes to settings here remain for all specs run afterwards.
-        # This is a problem.
```
elorest: Until settings is change to an instances instead of singleton this is still an "a problem".

src/amber/router/params_parser.cr
elorest: How is value an Array(JSON::Type) for you? The array should still be JSON::Any. In my tests it is.

src/amber/router/params_parser.cr
```diff
@@ -1,28 +1,29 @@
 require "json"
 
 module Amber::Router
-  # The Params module will parse parameters from a URL, a form post or a JSON
+  alias ParamsType = String | JSON::Type
```
elorest: Should this be JSON::Any?

src/amber/router/params_parser.cr
eliasjpr: We should pair on that one, to give you more context. But in a nutshell JSON.parse will return a JSON::Type and since we are not ‘unboxing’ to a specific type it reads as the first type `maybe` (I can be completely wrong here). 

This might work `value.as(String).raw.is_a? String` now this part of the params parser only gets trigger if the request is Application/JSON, this works fine as it is. So I think we can leave it

src/amber/router/params_parser.cr
eliasjpr: See https://crystal-lang.org/api/0.23.1/JSON/Type.html

eliasjpr: it will be spending  a lot of time instantiating a JSON::Any for each json param, instead of doing this ahead of time without knowing what params will be use in the request life-cycle the user can determine when he wants to parse out the JSON to the correct type or simply wrap it in a JSON::Any object. JSON::Any is a convenience wrapper to traverse the JSON.

elorest: But JSON.parse returns JSON::Any not JSON::Type right? So isn't it already doing that work?

src/amber/router/params_parser.cr
elorest: `.raw` only exists on unions and abstracts though. So if it's an Array(JSON::Any) .raw won't exist. In my tests values from JSON.parse are always JSON::Any though. Let me know when you have time to pair on this. Hopefully we can get it merged today.

src/amber/router/params_parser.cr
eliasjpr: JSON.parse return JSON::Type See https://github.com/crystal-lang/crystal/blob/e2a1389e8165fb097c785524337d7bb7b550a086/src/json.cr#L83

eliasjpr: Disregard the my comment above since I was looking at the wrong method 😕

src/amber/router/params_parser.cr
elorest: Ok so I didn't know you could do parse_raw. I was basing this off of the old implementation. So this is what I should have said.

```cr
          when Hash
            json.each do |key, value|
             if value.is_a? String
                params[key.as(String)] = value.to_s
             else
                params[key.as(String)] = value
             end
            end
          when Array
            params["_json"] = json
          end
```

src/amber/router/params_parser.cr
elorest: Got it thanks.

elorest:
@eliasjpr Why is this closed. I thought we were going to merge it?

watzon:
@elorest I was wondering the same thing

PR #318, opened 2017-10-20
Adds Amber Task Runner

eliasjpr:
As a developer, I want to be able to perform tasks using the framework.

- Adds Perform command to execute tasks
- Tasks should be inherited from **Amber::Tasks::Task**
- The task should declare a `perform` method. This is the method that gets
called.

✋ @aarongodin

### DIFF
--- a/spec/amber/cli/commands/perform_spec.cr
+++ b/spec/amber/cli/commands/perform_spec.cr
@@ -1,0 +1,36 @@
+require "../../../spec_helper"
+require "../../tasks/**"
+
+class FakeTask < Amber::Tasks::Task
+  def description
+    "Fake command task"
+  end
+
+  def perform
+    "Fake task completed!"
+  end
+end
+
+module Amber::CLI
+  describe MainCommand::Perform do
+    it "performs a tasks via command line" do
+      runner = MainCommand.run ["perform", "faketask"]
+
+      runner.should eq "Fake task completed!"
+    end
+
+    it "performs a tasks via command line alias" do
+      runner = MainCommand.run ["p", "faketask"]
+
+      runner.should eq "Fake task completed!"
+    end
+
+    it "shows task description" do
+      runner = MainCommand.run ["p", "-l", "faketask"]
+
+      runner.should eq(
+        %(FirstFakeTask\t\t #First fake task\nSecond::FakeTask\t\t #Second fake task\nFakeTask\t\t #Fake command task\n)
+      )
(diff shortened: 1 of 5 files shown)


imdrasil:
I've just took a quick look some missed a lot of things but what about [sam](https://github.com/imdrasil/sam.cr)?

eliasjpr:
Hey @imdrasil 

While I personally like `sam` a lot I find it that it will add a lot of complexity when we only need a simple way to define tasks it has to be ran separately from Amber CLI and it has an extensive DSL which adds a learning curve.

We are more than happy to feature SAM in Amber. What do you think of creating a Recipe using Amber with SAM?

imdrasil:
@eliasjpr nice idea, will take a look

eliasjpr:
@drujensen Will add the missing generator.

@imdrasil Open a PR here https://github.com/amberframework/online-docs and I will merge. Thank you this will be great!

drujensen:
@elorest good catch!  I'm not a big fan of how rails does this with the `task => :environment` syntax.  Should we always include the `config/**` so the environment loads or should this be up to the user to include?

eliasjpr:
@elorest I forgot to mention about this in the PR description. This was by design. it not load the amber environment. This is to make ver flexible and is going to be part of the documentation. The user just needs to require amber on the task to have it load.  

This is pretty much how Rails does it for Rails runner

eliasjpr:
Tasks generators will come in a separate PR

lady-elorest:
@eliasjpr @drujensen if we're not loading the amber environment how is this an advantage over just creating a crystal script that can be run with `crystal your_task.cr`

elorest:
@eliasjpr @drujensen if we're not loading the amber environment how is this an advantage over just creating a crystal script that can be run with `crystal your_task.cr`

drujensen:
@elorest I agree.  I rarely create a rake task that doesn't load the environment.  Rails provides both options.  Either we document how to do this or we provide a way to do it.  

It's pretty simple:
```
require "../config/**"
```
at the top of your rake task in order to include the environment, but I don't think its intuitive enough.

Maybe we allow the generator to add it and then the user can simply remove it if they don't want the environment added.

We can comment in the code saying `# comment the next line out if you don't want to load the environment` or something like that.

WDYT?

drujensen:
or we can default to not including the environment and say `# uncomment the next line to include the environment`

PR #323, opened 2017-10-22
Enable Teeplate interactive mode by default

adam-paterson:
This ensures users are prompted before their application
is overwritten with `amber new`

### Description of the Change
* Enables Teeplate's "interactive" mode to allow the user to chose whether their application is replaced during `amber new <name>`

### Alternate Designs
* Prompt the user that the target directory is not empty and request confirmation before application skeleton is generated

I disregarded this option as Teemplate interactive mode would allow users to pick and choose the files which are generated when executing `amber new`

### Benefits
* Users won't accidentally replace their application

### Possible Drawbacks
* Unaware of any currently


adam-paterson:
@elorest Must have mis-read the output when I tested this. The command currently fails. I'll look into and provide a fix for this later today.

screen shot 2017-10-23 at 09 18 32

adam-paterson:
@elorest looks like this is because of #327

elorest:
The reason for this error is that the .amber_secret_key has been replaced. So you can't decrypt. We should probably handle this error more gracefully though.

elorest:
There's a couple options:
Delete the old config that we can no longer decrypt since the key was just deleted.
Don't overwrite the secret_key and this won't happen.

adam-paterson:
I vote for not overwriting the secret_key. I’ll update the PR later.

eliasjpr:
Any updates on this?

adam-paterson:
@eliasjpr Sorry, been a majorly busy week. I'm working on this as we speak and should have something up soon.

I've gone with the approach of skipping `.amber_secret_key` during template generation if the file already exists. All other files are at the users discretion to overwrite, diff or skip.

eliasjpr:
@adam-paterson sweet would like to get this merge as soon as you are ready

elorest:
Might just want to skip all hidden files.

elorest:
Do we want to allow them to overwrite any configuration?

elorest:
In my opinion we shouldn't be overwriting any configuration files. All dotfiles and files inside of config should be exempt.

elorest:
It might be better just to return with an error if the folder exists instead of overwriting anything like `git` does. Or your solution is fine too.

marksiemers:
@elorest - This looks uncontroversial to me in its current state (raising an exception if the directory is not empty).

The only think I could think might improve it is a more eloquent error delivery - the exception raising will give stack-trace and everything won't it?

marksiemers:
Will re-open after repo cleaning. I have a copy of this branch, and it still exists in @adam-paterson fork. I believe @eliasjpr also has a local copy.

Issue #324, opened 2017-10-23
Unable to render layout in directory


adam-paterson:
### Description

Defining a layout in a sub-directory of `src/views/layouts` causes compilation error:

```
expanding macro
in macro 'embed' /Users/adampaterson/Code/garnet-app/lib/kilt/src/kilt.cr:12, line 4:

   1.
   2.
   3.
>  4.       Slang.embed("layouts/new_app/new_application.slang", "__kilt_io__", )
   5.
   6.

expanding macro
in macro 'embed' /Users/adampaterson/Code/garnet-app/lib/slang/src/slang/macros.cr:2, line 1:

>  1.     {{ run("slang/slang/process", "layouts/new_app/new_application.slang", "__kilt_io__") }}
   2.

Error executing run: slang/slang/process "layouts/new_app/new_application.slang" "__kilt_io__"

Got:
```

### Steps to Reproduce

1. Create controller
**src/controllers/new_controller.cr**
```ruby
class NewController < ApplicationController
  LAYOUT = "new_app/new_application.slang"

  def index
    render("new_app/index.slang")
  end
end
```
2. Add route
**config/routes.yml**
```yaml
routes :web do
  get "/new", NewController, :index
end
```
3. Create layout
**src/views/layouts/new_app/new_application.slang**
```slim
body
  == content
```
4. Create view
**src/views/new_app/index.slang**
```slim
h1 New Application
```

5. Build (or amber watch)

**Expected behavior:**
- Layout is rendered with content from view

**Actual behavior:**
- Compilation error

**Reproduces how often:**
- 100%

### Versions
- Crystal 0.23.1 (2017-10-12) LLVM 4.0.1
- Amber CMD (amberframework.org) - v0.3.0
- macOS High Sierra 10.13.1 Beta (17B35a)

### Additional Information
- I've tried to debug this but only have `lldb` setup and have no access to vars in my IDE.

- Replacing `render()` with what the macro should be generating works as expected
```ruby
content = Kilt.render("src/views/new_app/index.slang")
Kilt.render("src/views/layouts/new_app/new_application.slang")
```
- I've tried a combination of `render` including defining all params in my call
- `src/view/layouts/new_application.slang` works when in the parent `layouts` directory

faustinoaq:
I can reproduce this issue

Amber render wrapper is [here](https://github.com/amberframework/amber/blob/aa4f941ee06454dadfc958393603c4dc99dc02a8/src/amber/controller/render.cr#L5)

[...]

PR #328, opened 2017-10-24
Fixed an issue I found with render if we use namespaces.

elorest:
### Description of the Change

While thinking about https://github.com/amberframework/amber/issues/324 I realized we would have an issue if we we use namespaces. So I've modified render to allow for this. Layouts an also be defined in folders now. `LAYOUT = "some/folder/application.slang"`


adam-paterson:
@eliasjpr Line for line it hasn't changed. Perhaps this can be raised as a separate issue?

elorest:
@eliasjpr I'd like to pair with you on cleaning this up someday. I've played with it a bit more but since there isn't any duplication there's nothing to dry up. I could always extract parts out into other macros but since it's only called once it seems a little pointless and only slows down compilation. WDYT? Anyway I'm going to merge this for now.

Issue #316, opened 2017-10-19
Controller generator is not great with multi-word names


aarongodin:
### Description

The generator for controllers has trouble with multi-word names. Creating a multi-word controller name results in incorrect file names or incorrect class names, depending on the name you pass.

### Steps to Reproduce

Test case 1:  `amber g controller PostComments`
Test case 2: `amber g controller post_comments`

**Expected behavior:**

- I see a file called `post_comments_controller_spec.cr` was created.
- I see a file called `post_comments_controller.cr` was created.
- Both the spec and the controller use the class name `PostCommentsController`

**Actual behavior:** [What actually happens]

Test case 1:

- class name is `PostcommentsController`
- file names are `postcomments_controller`, rather than an underscore between the words.

Test case 2:

- class name is `Post_commentsController`
- file names are correct!

**Reproduces how often:**

Always

### Versions

0.3.0

aarongodin:
After looking further, this seems to be pretty generic as to how all file names and class names are created. I'm seeing the same issue when generating models as well.

Perhaps the generators should have a shared piece of logic for taken a user-given name and discerning its CamelCase and snake_case names.

eliasjpr:
@aarongodin How do you feel about opening a PR addressing this issue? 😄  We welcome the help.

aarongodin:
@eliasjpr not a problem! I would love to help out.

marksiemers:
@aarongodin - If you don't object, I would like to take a shot at this.

aarongodin:
@marksiemers Looks nice!

marksiemers:
@aarongodin - Is this fixed for you on the master branch? Can we close the issue?

aarongodin:
@marksiemers It's fixed from my perspective. Closing!

Issue #329, opened 2017-10-24
App names with dashes (-) create invalid postgresql database names


marksiemers:
### Description

When creating an amber application containing a dash (e.g. `my-blog`) the database name will not be valid for postgresql (e.g. `my-blog_development`).

It may be as simple as adding `@name.gsub('-','_')` to the `database.yml` generator, but I also know there is talk about moving the url to a different file, and I'm not yet familiar enough with the codebase to know all the places the database name is required.

### Steps to Reproduce

```
$ amber new my-blog
$ cd my-blog
$ shards install
$ amber db create
syntax error at or near "-"
```

**Expected behavior:**
Expect the database to be created as `my_blog_development` (or another postgresql compatible name)

**Actual behavior:**
An error is reported:
`syntax error at or near "-"`

**Reproduces how often:**
It happens with any app name that has a dash in it. The `database.yml` file is easy enough to manually edit to fix this, so it isn't a very high priority, but maybe low-hanging fruit to fix.

### Versions

- Amber CMD (amberframework.org) - v0.3.0
- PostgreSQL 9.6.5 on x86_64-apple-darwin16.7.0, compiled by Apple LLVM version 8.1.0 (clang-802.0.42), 64-bit
- OS X Sierra
   - ProductName: Mac OS X
   - ProductVersion: 10.12.6
   - BuildVersion: 16G29

### Additional Information

This may be a separate issue, but in my experience with postgresql, the username `postgres` is the default root user, rather than `root`. Amber's default of `root` for postgresql also requires a manual change to `database.yml` if the local version of postgresql has the default `postgres` root user.

Issue #294, opened 2017-10-07
Error pages generator


drujensen:
### Description

When first creating a new application, an Error Controller is provided but it isn't hooked up and doesn't do anything without some work.

Create a new generator that will provide the 3 views, the Error Controller, and the registry of the handler in the pipeline.

marksiemers:
Are there two steps to this issue?
1. Create fully functioning `error` generator (probably take cues from `auth` generator)
1. Remove the error controller from being generated with `amber new`

I could probably take this one.

marksiemers:
I have some WIP stashed for this, but waiting on https://github.com/amberframework/amber/pull/330 to get merged in.

@drujensen - Can you confirm my question about the the steps for this issue above?

drujensen:
@marksiemers that is correct.  There are a couple helper methods to inject the pipeline and add the routes.  You can find those used in the Auth generator as well.  Thanks for taking this on.

eliasjpr:
Being addressed in https://github.com/amberframework/amber/pull/345

marksiemers:
Can this be moved to 0.4.0 milestone?

marksiemers:
#415

PR #330, opened 2017-10-26
Have generators handle multi-word names (#316)

marksiemers:
### Description of the Change

This changes how the generators handle file names and class names. It is related to #316 .
That issue was originally about controllers, but it applies to all the generators, so this fix addresses all of them.

I ended up added a lot of LOC to the test suite. Part of this is due to the `before_each` and `after_each` hooks in crystal being global (so I couldn't use them here), another reason is that the generators create a lot of files, checking each one takes a bit of code.

That said, I'm opening this PR for initial feedback to see if this approach is acceptable - both to testing and coding this fix. If there is a better way to structure the test suite, I'm all ears.

### Alternate Designs

I thought about DRY'ing the code up a bit so that `camelcase` wasn't used in so many places, but it would have required adding at least an instance variable and a getter to the `Template` - in the end that may be the best solution, in case the generators need to deal with other naming considerations (singular and plural).

I went with this solution to get something working, and provided a (hopefully) comprehensive set of tests if a refactor happens in the future.

### Benefits

It fixes issue #316 

### Possible Drawbacks

This change touches a lot of files and has an effect on every generator. Based on the test suite, everything is still generating as normal, but it is possible something breaks in the actual generation of files.

On my machine, one of the pre-existing tests is either not passing, or hanging indefinitely. I checked out the master branch and got the same results. 

It hangs here: https://github.com/amberframework/amber/blob/master/spec/amber/cli/commands/generator_spec.cr#L14

Or fails here: https://github.com/amberframework/amber/blob/master/spec/amber/cli/commands/generator_spec.cr#L16

### DIFF
--- a/spec/amber/cli/commands/generator_spec.cr
+++ b/spec/amber/cli/commands/generator_spec.cr
@@ -17,13 +17,88 @@ module Amber::CLI
 
           Amber::CLI::Spec.cleanup
         end
+
+        it "follows naming conventions for all files and class names" do
+          ENV["AMBER_ENV"] = "test"
+          Amber::CLI::MainCommand.run ["new", TESTING_APP]
+          Dir.cd(TESTING_APP)
+
+          camel_case = "PostComment"
+          snake_case = "post_comment"
+          class_definition_prefix = "class #{camel_case}"
+          spec_definition_prefix = "describe #{camel_case}"
+
+          [camel_case, snake_case].each do |arg|
+            MainCommand.run ["generate", "scaffold", arg, "name:string"]
+
+            File.exists?("./spec/models/#{snake_case}_spec.cr").should be_true
+            File.exists?("./src/models/#{snake_case}.cr").should be_true
+            File.exists?("./spec/controllers/#{snake_case}_controller_spec.cr").should be_true
+            File.exists?("./src/controllers/#{snake_case}_controller.cr").should be_true
+            File.exists?("./src/views/#{snake_case}/_form.slang").should be_true
+            File.exists?("./src/views/#{snake_case}/edit.slang").should be_true
+            File.exists?("./src/views/#{snake_case}/index.slang").should be_true
+            File.exists?("./src/views/#{snake_case}/new.slang").should be_true
+            File.exists?("./src/views/#{snake_case}/show.slang").should be_true
+
+            File.read("./spec/models/#{snake_case}_spec.cr").should contain spec_definition_prefix
+            File.read("./src/models/#{snake_case}.cr").should contain class_definition_prefix
+            File.read("./spec/controllers/#{snake_case}_controller_spec.cr").should contain spec_definition_prefix
+            File.read("./src/controllers/#{snake_case}_controller.cr").should contain class_definition_prefix
+            File.read("./src/views/#{snake_case}/_form.slang").should contain snake_case
+            File.read("./src/views/#{snake_case}/edit.slang").should contain camel_case
+            File.read("./src/views/#{snake_case}/index.slang").should contain camel_case
+            File.read("./src/views/#{snake_case}/new.slang").should contain camel_case
+            File.read("./src/views/#{snake_case}/show.slang").should contain snake_case
+
+            File.delete("./spec/models/#{snake_case}_spec.cr")
+            File.delete("./src/models/#{snake_case}.cr")
+            File.delete("./spec/controllers/#{snake_case}_controller_spec.cr")
+            File.delete("./src/controllers/#{snake_case}_controller.cr")
+            File.delete("./src/views/#{snake_case}/_form.slang")
+            File.delete("./src/views/#{snake_case}/edit.slang")
+            File.delete("./src/views/#{snake_case}/index.slang")
+            File.delete("./src/views/#{snake_case}/new.slang")
+            File.delete("./src/views/#{snake_case}/show.slang")
+          end
+          Amber::CLI::Spec.cleanup
+        end
       end
 
-      context "controllers" do
+      context "model" do
+        ENV["AMBER_ENV"] = "test"
+        Amber::CLI::MainCommand.run ["new", TESTING_APP]
+        Dir.cd(TESTING_APP)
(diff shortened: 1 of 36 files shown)


spec/amber/cli/commands/generator_spec.cr
```diff
@@ -16,13 +16,88 @@ module Amber::CLI
...
+
+            File.delete("./spec/models/#{snake_case}_spec.cr")
+            File.delete("./src/models/#{snake_case}.cr")
+            File.delete("./spec/controllers/#{snake_case}_controller_spec.cr")
+            File.delete("./src/controllers/#{snake_case}_controller.cr")
+            File.delete("./src/views/#{snake_case}/_form.slang")
+            File.delete("./src/views/#{snake_case}/edit.slang")
+            File.delete("./src/views/#{snake_case}/index.slang")
+            File.delete("./src/views/#{snake_case}/new.slang")
+            File.delete("./src/views/#{snake_case}/show.slang")
```
veelenga: If one expectation above fails these files will not be deleted. Or presence of this files are guaranteed by `Amber::CLI::Spec.cleanup` ? If not, you may need to use `try ... ensure` block for this:

```crystal
try
  # expectations
ensure
  # delete files
end
```

marksiemers: The calls to `Amber::CLI::Spec.cleanup` should ensure cleanup, but admittedly, I didn't explicitly test for or look for it.

I did run a bunch of failing test suites though, and there were never any leftovers (according to git).

The entire spec in this file is wrapped in a `begin ... ensure ... end` that executes `Amber::CLI::Spec.cleanup` which is more comprehensive than my file deletion.

I can rerun some failing tests and explicitly look for leftover files (as they may be gitignored).

marksiemers: I made it fail one of the `should` tests on generating a scaffold, then did an `ls tmp` and it was empty. It looks like we are good to count on `Amber::CLI::Spec.cleanup`

veelenga: So you don't need to delete it manually, right ?

marksiemers: When the test suite is complete, `Amber::CLI::Spec.cleanup` makes sure everything is cleaned up.

The test itself tests two cases of possible CLI input (in a loop) - `PostComment` and `post_comment`

The reason for the manual delete is to avoid false green tests with the second iteration of the loop.

veelenga: Ah, sorry, i haven't noticed the loop above. Thanks for explanation!

marksiemers:
The conflicts that showed up - 3 of them are not in files that I changed.
What's the preferred course of action here? Resolve the conflicts here, or rebase (off master) my changes locally and push with the new commits?

eliasjpr:
Rebase off master

marksiemers:
Rebased, refactored, and add a new concept of `display_name`
LMKWYT

marksiemers:
Hold off on this, I found a bug while manually testing - I didn't touch the routes generator.

marksiemers:
Manually tested with both Crecto and Granite, and added spec tests for both as well.

LMK if anything else is needed for this PR.

drujensen:
@marksiemers I found a couple issues testing the branch:

1. The name of the project having dashes:
```
amber new my-cool-blog
cd my-cool-blog
amber g scaffold post title:string
amber db drop create migrate
```
produces:
`syntax error at or near "-"`

The `config/database.yml` needs to be changes to replace the dashes in the db name.
```
pg:
  database: postgres://root:@localhost:5432/my-cool-blog_development
```

2. The field name having dashes:
```
amber g scaffold post my-title:string
```
`An error ocurred executing migration 20171028085330. Error message is: syntax error at or near "-"`

marksiemers:
@drujensen - The issue with the app name having dashes is a pre-existing bug
Reported here: https://github.com/amberframework/amber/issues/329
Pending fix here: https://github.com/amberframework/amber/pull/332

marksiemers:
Any chance of this getting merged soon? The work I'm doing on https://github.com/amberframework/amber/issues/294 requires code that will need to be reworked after this change is incorporated.

spec/amber/cli/commands/generator_spec.cr
```diff
@@ -17,13 +17,122 @@ module Amber::CLI
...
+            File.read("./src/views/#{snake_case}/_form.slang").should contain snake_case
+            File.read("./src/views/#{snake_case}/edit.slang").should contain display
+            File.read("./src/views/#{snake_case}/index.slang").should contain display
+            File.read("./src/views/#{snake_case}/new.slang").should contain display
+            File.read("./src/views/#{snake_case}/show.slang").should contain snake_case
+
+            File.read("./config/routes.cr").should contain "#{camel_case}Controller"
+            File.read("./config/routes.cr").should_not contain "#{incorrect_case}Controller"
+
+            File.delete("./spec/models/#{snake_case}_spec.cr")
```
elorest: Why are we deleting all of these one by one? We already delete he folder after the test. At least we should be.

elorest: Looks like you sort of answered my question below. Even if you do have a reason for deleting these we shouldn't be deleting them one by one.

marksiemers: I can clean it up, the two (not mutually exclusive) ideas that come to mind first are:
- Use `rm -rf ./spec/*` `rm -rf ./src/*`
- Refactor to put the file handling in a module

elorest: The first option seems clean and easy to read.

marksiemers: For the scaffold, I changed it so that a new app is created for each iteration of the loop, and the spec `cleanup` is also run. This is because the routes file is touched and that can't be deleted safely without generating a new app.

When applicable, I switched to `rm -rf ./[dir]/*`

PR #331, opened 2017-10-26
Better Request Parsing Parameters

drujensen:
Edited by @eliasjpr
### Description of the Change

The framework should be able to parse JSON request body appropriately by
returning a `JSON::Any` object.

Currently, there is a bug when trying to parse JSON parameters from the
body of a request where nested JSON payload is not being parsed correctly

```crystal
{ "test": "test", "address": { "city": "New York" }
params["address"].as(Hash)["city"]
```
Params is now a custom Hash of the following signature
`Hash(String | Symbol, Amber::Router::ParamsType)` or
`Amber::Router::ParamsHash.new`

To give more context all parameters where being parsed using the
**HTTP::Params.parse** method which returns an object of type
**Hash(String, Array(String))** this caused issues when parsing
JSON params from the body, It turns out that nested JSON keys where
being parsed as a string. Basically, all `params` values are currently
returning as a string.

With the changes presented in this PR users can now define model objects
with JSON mapping that should be evaluated correctly based on the mapping
defined. 

**Controller**
The controller now generates the app using params.validation similar to
Rails strong params. It adds a method to the Controller and overloads the
Granite initializer and set_attributes methods.

```crystal
private def animal_params
  params.validation do
    required(:name) { |v| v.str? & !v.empty? }
    required(:last_name, "Last name is required")
    optional(:phone)  
  end
end

# Validate! returns a hash of only the specified params above.
animal = Animal.new(animal_params.validate!)
```

### Alternate Designs

It would be nice to specify a type that can be cast automatically

```crystal
required(:last_name, :string, "Last name is required") do
...validation code here...
end
```

### Benefits
- Fixes Issue: https://github.com/amberframework/amber/issues/310
- Parses JSON body as JSON::Type. with this type user can convert to the correct type either using 
`.to_i`, `.to_f`, `.to_s` or traverse the JSON using `json = JSON.parse(params["some_key"])`
- Uses params.validation to validate request params before any processing. Inspired by http://hanamirb.org/guides/1.0/actions/parameters/#validations-amp-coercion
- Updates Auth, scaffold generators to use the params.validation

### Possible Drawbacks

-Fear something can break

### Example app using this branch 
https://github.com/eliasjpr/params-test-app

### DIFF
--- a/src/amber/cli/templates/auth.cr
+++ b/src/amber/cli/templates/auth.cr
@@ -25,7 +25,7 @@ module Amber::CLI
       @timestamp = Time.now.to_s("%Y%m%d%H%M%S")
       @primary_key = primary_key
       @visible_fields = @fields.reject(&.hidden).map(&.name)
-      
+
       add_routes :web, <<-ROUTES
         get "/signin", SessionController, :new
           post "/session", SessionController, :create
@@ -37,7 +37,7 @@ module Amber::CLI
       add_plugs :web, <<-PLUGS
         plug Authenticate.new
       PLUGS
-      
+
       add_dependencies <<-DEPENDENCY
       require "../src/models/**"
       require "../src/handlers/**"
--- a/src/amber/cli/templates/auth.cr
+++ b/src/amber/cli/templates/auth.cr
@@ -28,10 +28,10 @@ module Amber::CLI
 
       add_routes :web, <<-ROUTES
         get "/signin", SessionController, :new
-          post "/session", SessionController, :create
-          get "/signout", SessionController, :delete
-          get "/signup", RegistrationController, :new
-          post "/registration", RegistrationController, :create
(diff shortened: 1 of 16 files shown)


eliasjpr:
Example app to test integration with Granite https://github.com/eliasjpr/params-test-app

src/amber/cli/templates/auth.cr
```diff
@@ -25,7 +25,7 @@ module Amber::CLI
       @timestamp = Time.now.to_s("%Y%m%d%H%M%S")
       @primary_key = primary_key
       @visible_fields = @fields.reject(&.hidden).map(&.name)
-      
+
       add_routes :web, <<-ROUTES
         get "/signin", SessionController, :new
           post "/session", SessionController, :create
```
faustinoaq: Indentation issue

PR #335, opened 2017-10-28
Fix missing require for task runner

eliasjpr:
### Description of the Change

Currently Amber is not compiling on master because of missing requires.
This adds require to the correct places.
Amber compiles successfully after.


drujensen:
@elorest I'm going to merge to get master fixed.  If you have feedback, still provide it.

elorest:
So I left that message above. Not used to there being someone else logged into github on my livingroom computer.

PR #340, opened 2017-10-30
Use table name as part of database index name (#336)

marksiemers:
### Description of the Change

Fixes https://github.com/amberframework/amber/issues/336

### Alternate Designs

The lines that fix this issue exist in two ECR files. The code can likely be refactored into the Migration template class.

### Benefits

Fixes https://github.com/amberframework/amber/issues/336

### Possible Drawbacks

None known.

PR #344, opened 2017-10-30
Clean Up and Dry Tests

eliasjpr:
### Description of the Change

As the framework grows having a clean and expressive test suite is very important for those who are currently working on the project and new people that want to contribute. 

This PR is a WORK IN PROGRESS to clean up and structure the test using the Arrange, Act, Assert Annihilate pattern with only one action and one assertion.

### Alternate Designs

- Currently the test suite is just testing each class part of the Amber Framework maybe it would be a good idea to create 2 directories for integrations, unit tests

### Benefits

With the work done it will make it very easy to read and maintain the test suite

### Possible Drawbacks

- Maybe there is a better design outther


spec/support/fixtures/cli_fixtures.cr
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def remove
+    render("remove.slang")
+  end
+end
+
+CONT
+  end
+end
```
marksiemers: Crystal will create whitespace based on the closing `CONT`, so you can do:
```crystal
module CLIFixtures
  def expected_controller
    <<-CONT
    class AnimalController < ApplicationController
      def add
        render("add.slang")
      end
      def list
        render("list.slang")
      end
      def remove
        render("remove.slang")
      end
    end
    CONT
  end
end
```

eliasjpr: Yeah I was confused about this, fixing it

marksiemers:
@eliasjpr Thanks for your work on this.

I like a lot of the refactoring so far (particularly `scaffold_app`)

My main concern is moving `should` statements into helpers. Crystal's `Spec` definitely has taken cues from RSpec, and there, the convention is to have `should` statements directly in an `it` block. (Well in RSpec it is `expect` now).

spec/support/fixtures/cli_fixtures.cr
marksiemers: The closing `CONT` should align with `class` and the last `end`

eliasjpr:
I agree when having 1 assertion line, but when there is multiple line attempting to assert related expectations we should have them in a composed assert method, this makes it very reusable, readable and follow the one assert rule. IMHO

An example of this is when asserting if the encryption files exists after encrypting.
```crystal
File.exists?("config/environments/#{enc_file}").should be_true
File.read(enc_key).size.should eq 44
```

marksiemers:
I think "Assert" in the AAA pattern is roughly equal to `should` in Spec, if you move two `should`s into a helper, IMHO it is just hiding that there are two asserts.

I think the idea behind the one assertion per test, is to make each test (`it`) statement as granular as possible.

Using this [spec file](https://github.com/amberframework/amber/blob/3f408a4fd6e422159f5c0303cb69f5783ee9d308/spec/amber/cli/commands/encrypt_spec.cr) as a starting point
```crystal
describe MainCommand::Encrypt do
  context "application structure" do
    it "creates amber directory structure" do
      scaffold_app(TESTING_APP)
      MainCommand.run ["encrypt", "test"]
      assert_encrypted_files_exists?(".test.enc", ".amber_secret_key")
      cleanup
    end
  end
end
```
The `it` statement seems broad, and does not match what the assertion is actually checking.

I would like to see that our goal is to have each `it` only execute one `should` (directly or indirectly). For example, I find the following code more readable and I have a better understanding of what `amber encrypt test` should do:
```crystal
describe MainCommand::Encrypt do
  begin
    scaffold_app(TESTING_APP)
    arg = "test"
    MainCommand.run ["encrypt", arg]

    it "should create an hidden .[arg].enc file" do
      File.exists?("config/environments/.#{arg}.enc").should be_true
    end

    it "should create a 44 character secret key in .amber_secret_key" do
      File.read(".amber_secret_key").size.should eq 44
    end

  ensure
    cleanup
  end
end
```

I know it isn't always practical to have a 1:1 `it` to `should` ratio, but I think using it as a guide can lead to more declarative and descriptive.

eliasjpr:
Thats what I am looking for! Thanks @marksiemers

marksiemers:
It isn't bad practice to setup descriptions for all your known requirements right away:
```crystal
module Amber::CLI
  describe MainCommand::Perform do
    it "performs a tasks via command line" {} # `it` requires a block

    it "performs a tasks via command line alias" {}

    it "shows task description" {}
  end
end
```

spec/amber/cli/commands/generator_spec.cr
```diff
@@ -6,32 +6,39 @@ include CLIHelper
...
-          MainCommand.run ["generate", "scaffold", "Animal", "name:string"]
-          assert_app_compiled?(TEST_APP_NAME)
-          cleanup
-        end
+  describe MainCommand::Generate do
+    ENV["AMBER_ENV"] = "test"
+
+    context "when scaffolding an Amber app" do
+      scaffold_app(TESTING_APP)
+      MainCommand.run ["generate", "scaffold", "Animal", "name:string"]
```
marksiemers: This line shouldn't be needed to test a `shards build` - did we just want something in there?

If we want a more complete test, it should be structured something like this:
```
context "scaffolding an Amber app"  
  assert that it builds
  test scaffold generator (in a describe block)
  test controller generator (in a describe block)
  test model generator (in a describe block)
  ...test each generator...
  assert that it still builds
  cleanup
end
```

spec/amber/cli/commands/generator_spec.cr
```diff
@@ -6,32 +6,39 @@ include CLIHelper
...
-      context "controllers" do
-        it "generates controller with correct verbs and actions" do
-          ENV["AMBER_ENV"] = "test"
-          controller = "Animal"
-          options = %w(add:post list:get remove:delete)
-          scaffold_app(TESTING_APP)
+      cleanup
+    end
+
+    context "when generating controllers" do
```
marksiemers: Semantically, I would say we are in the context of "scaffolding and amber app" and this should be `describe "amber generate controller"`

spec/amber/cli/commands/generator_spec.cr
```diff
@@ -6,32 +6,39 @@ include CLIHelper
...
+      controller = "Animal"
+      options = %w(add:post list:get remove:delete)
+      route_file = File.read("./config/routes.cr")
+      MainCommand.run ["generate", "controller", controller] | options
+
+      it "generates controller with correct verbs and actions" do
+        File.read("./src/controllers/#{controller.downcase}_controller.cr").should eq expected_controller
```
marksiemers: This line and line 25 are coupling this code with wherever `expected_controller` is defined. I would say choose one place and have both `controller = "Animal"` and whatever is in the definition of `expected_controller`. I think the helper or the fixture is the better place.

marksiemers: For an example of how I approached it, see here: https://github.com/amberframework/amber/pull/340/commits/3220d1312608f53ca486b8f5a668219870c19b57

And the helper here: https://github.com/marksiemers/amber/blob/3220d1312608f53ca486b8f5a668219870c19b57/spec/amber/cli/templates/migration_spec_helper.cr

All the decisions about names are in the helper - so they are easier to keep in sync.

spec/amber/cli/commands/generator_spec.cr
marksiemers: Practically speaking, we may not want two builds, until the shard downloading thing is optimized one way or another.

It might be better to test the shard building at the end - will it build after all the generators have been run, or did any of them introduce a compile time error?

spec/support/fixtures/render_fixtures.cr
```diff
@@ -0,0 +1,50 @@
...
+        <textarea class="form-control" rows="10" name="content" placeholder="Content">out there in the cold</textarea>
+      </div>
+      <button class="btn btn-primary btn-xs" type="submit">Submit</button>
+      <a class="btn btn-default btn-xs" href="/posts">back</a>
+    </form>
+    HTML
+  end
+
+  def layout_with_template
+    %(<html>\n  <body>\n    <h1>Hello World</h1>\n<p>I am glad you came</p>\n  </body>\n</html>)
```
marksiemers: Any reason for this not to use a heredoc like the others?

spec/amber/cli/commands/init_spec.cr
```diff
@@ -1,44 +1,62 @@
...
-          Amber::CLI::Spec.db_yml["pg"].should_not be_nil
-          shard_yaml = Amber::CLI::Spec.shard_yml
-          shard_yaml["dependencies"]["pg"].should_not be_nil
-          shard_yaml["dependencies"]["amber"]["version"].should eq Amber::VERSION
-          Amber::CLI::Spec.amber_yml["language"].should eq "slang"
-          File.read("#{TESTING_APP}/.amber_secret_key").size.should eq 44
-          Amber::CLI::Spec.cleanup
-        end
+  describe MainCommand::New do
+    context "application structure" do
```
marksiemers: I would call this `describe "amber new #{TESTING_APP}"`

spec/amber/cli/commands/init_spec.cr
```diff
@@ -1,44 +1,62 @@
...
+        it "sets PG shards dependencies" do
+          shard_yml["dependencies"]["pg"].should_not be_nil
         end
+
+        cleanup
+      end
+    end
+
+    context "template" do
+      cleanup
```
marksiemers: Should this be necessary?

spec/amber/controller/base_spec.cr
```diff
@@ -26,107 +30,53 @@ module Amber::Controller
...
-          <div class="form-group">
-            <textarea class="form-control" rows="10" name="content" placeholder="Content">out there in the cold</textarea>
-          </div>
-          <button class="btn btn-primary btn-xs" type="submit">Submit</button>
-          <a class="btn btn-default btn-xs" href="/posts">back</a>
-        </form>
-        HTML
-
-        TestController.new(context).render_with_csrf.should eq html_output
+        TestController.new(context).render_with_csrf.should eq csrf_form
```
marksiemers: All of this looks so much better with the refactor, nice work.

spec/support/fixtures/cli_fixtures.cr
```diff
@@ -0,0 +1,20 @@
...
+  def list
+    render("list.slang")
+  end
+
+  def remove
+    render("remove.slang")
+  end
+end
+
+CONT
```
marksiemers: As long as you move the closing `CONT` over, you can safely indent the string.

spec/support/helpers.cr
```diff
@@ -1,70 +1,13 @@
...
+require "./helpers/router_helper"
+require "./helpers/websockets_helper"
+require "./helpers/validations_helper"
+
+module Helpers
+  include ControllerHelper
+  include CookieHelper
+  include RouterHelper
+  include WebsocketsHelper
+  include ValidationsHelper
```
marksiemers: I really like this organization.

spec/amber/cli/commands/encrypt_spec.cr
```diff
@@ -1,21 +1,23 @@
...
-          MainCommand.run ["new", TESTING_APP]
-          Dir.exists?(TESTING_APP).should be_true
-          Dir.cd(TESTING_APP)
-          MainCommand.run ["encrypt", "test"]
-          File.exists?("config/environments/.test.enc").should be_true
-          File.read(".amber_secret_key").size.should eq 44
-          Amber::CLI::Spec.cleanup
-        end
-      end
+  describe MainCommand::Encrypt do
```
marksiemers: Any opinion on having this be `describe "amber encrypt" do` ?

spec/amber/cli/commands/generator_spec.cr
```diff
@@ -1,36 +1,82 @@
...
-          MainCommand.run ["new", TESTING_APP]
-          Dir.cd(TESTING_APP)
-          MainCommand.run ["generate", "scaffold", "Animal", "name:string"]
-          Amber::CLI::Spec.prepare_yaml(Dir.current)
+include CLIHelper
+include CLIFixtures
 
-          `shards build`
+module Amber::CLI
+  describe MainCommand::Generate do
```
marksiemers: Same thing here - we could have `describe "amber generate" do` and remove the context below.

spec/amber/cli/commands/init_spec.cr
```diff
@@ -1,88 +1,130 @@
...
-          Dir.exists?(TESTING_APP).should be_true
-          Amber::CLI::Spec.dirs(TESTING_APP).sort.should eq Amber::CLI::Spec.dirs(APP_TPL_PATH).sort
-          Amber::CLI::Spec.db_yml["pg"].should_not be_nil
-          shard_yaml = Amber::CLI::Spec.shard_yml
-          shard_yaml["dependencies"]["pg"].should_not be_nil
-          shard_yaml["dependencies"]["amber"]["version"].should eq Amber::VERSION
-          Amber::CLI::Spec.amber_yml["language"].should eq "slang"
-          File.read("#{TESTING_APP}/.amber_secret_key").size.should eq 44
-          Amber::CLI::Spec.cleanup
+  describe MainCommand::New do
```
marksiemers: Same thing here, `describe "amber new #{TESTING_APP}" do` and remove the context below.

spec/amber/cli/commands/init_spec.cr
```diff
@@ -1,88 +1,130 @@
...
+      incorrect_case = "Post_comment"
+      display = "Post Comment"
+      class_definition_prefix = "class #{camel_case}"
+      spec_definition_prefix = "describe #{camel_case}"
+
+      it "generates amber directory structure" do
+        dirs(TESTING_APP).sort.should eq dirs(APP_TEMPLATE_PATH).sort
+      end
+
+      describe "-m granite (Granite ORM)" do
```
marksiemers: I would call this `context` but it isn't a deal-breaker.

spec/amber/cli/commands/init_spec.cr
```diff
@@ -1,88 +1,130 @@
...
 
-      context "database" do
-        context "postgres" do
-          MainCommand.run ["new", TESTING_APP]
-          it "creates app with correct settings" do
-            Amber::CLI::Spec.shard_yml["dependencies"]["pg"].should_not be_nil
-            Amber::CLI::Spec.db_yml["pg"].should_not be_nil
-            db_url = Amber::CLI::Spec.db_yml["pg"]["database"].as_s
-            db_url.should_not be_nil
+      describe "-m crecto (Crecto Repo)" do
```
marksiemers: Same thing, `context`

spec/support/helpers/cli_helper.cr
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+  def assert_route_generated?(route_file, route)
+    route_file.includes?(route).should be_true
+  end
+
+  def assert_encrypted_files_exists?(enc_file, enc_key)
+    File.exists?("config/environments/#{enc_file}").should be_true
+    File.read(enc_key).size.should eq 44
+  end
```
marksiemers: Are these still needed?

Issue #342, opened 2017-10-30
Test suite takes too long to run


marksiemers:
### Description

When running the test suite locally, it takes 2-5 minutes to complete. This is mostly due to a `shards build` on [this line](https://github.com/amberframework/amber/blob/master/spec/amber/cli/commands/generator_spec.cr#L14)

### Steps to Reproduce

1. `crystal spec`
2. wait ...
3. wait ...

**Expected behavior:** [What you expect to happen]

The full test suite to run more quickly on local development machine (under a minute, hopefully under 10 seconds)

**Actual behavior:** [What actually happens]

The full test suite takes several minutes to run.

**Reproduces how often:** [What percentage of the time does it reproduce?]

Every time you run `crystal spec`

### Versions

amber v0.3.0

### Additional Information

In gitter we discussed taking the following approach:
1. Local development, by default should NOT run shards build
1. Everywhere else (mostly Travis) it should run shards build by default
1. It should be easy to opt in on local development to running the shards build test from the command line

Either using an environment variable or option flag.

marksiemers:
@eliasjpr @elorest - Please weigh in

elorest:
That test proves the generated code actually works which I feel is important. Possibly having a flag to turn it off and on is a good idea. Tests take 1:28 minutes to complete as they are on my development box. Usually for development though I just run specific tests while developing and then the complete suite before pushing.

marksiemers:
I think it depends on your internet connection:

With build
```bash
Finished in 2:25 minutes
336 examples, 0 failures, 0 errors, 0 pending

real	2m38.068s
user	1m28.255s
sys	0m13.785s
```

Without build:
```bash
Finished in 574.5 milliseconds
336 examples, 0 failures, 0 errors, 0 pending

real	0m19.855s
user	0m36.337s
sys	0m3.854s
```

marksiemers:
I got some info from the Crystal gitter about how to possibly do this:
`{% if flag?(:slow_test) %} require "./slow/*" {% end %}`
and it sounds like `-D` allows you to pass option flags.

As an extension to this issue, I think isolating a `build` test file that is separate and isolated is a good idea. We can have it be a large integration test for the CLI - generate a new app, generate at least one of everything possible to generate, then build.

That way, if anything introduces a compile time error, it should show up.

eliasjpr:
[...]

PR #345, opened 2017-10-31
[WIP] Implement error generator (#294)

marksiemers:
### Description of the Change

Add a new CLI generator for an error controller and views. Requested in #294 

### Alternate Designs

I used the scaffold generator as a guideline for file structure and class design decisions in `src`

For `spec` - I created a fixture in an attempt to match what @eliasjpr has been doing in https://github.com/amberframework/amber/pull/344

### Benefits

A user will be able to generate an error controller and views.

### Possible Drawbacks

This particular generator should not require a name - I think enforcing that it is called "error" is a good choice. Given that, we'll have to either ignore any name given, or change the logic for making `name` a required argument for generators.

This is a WIP because I have some questions, which I will include in the comments.


spec/amber/cli/commands/generate/error_template_spec.cr
```diff
@@ -0,0 +1,35 @@
+require "../../../../spec_helper"
+require "../../../../support/fixtures/cli/generators/error/*"
```
marksiemers: What do you think of this file structure choice for the fixture?

spec/amber/cli/commands/generate/error_template_spec.cr
```diff
@@ -0,0 +1,35 @@
...
+
+          it "generates a controller spec file" do
+            File.exists?("./spec/controllers/error_controller_spec.cr").should be_true
+          end
+
+          it "generates view files" do
+            ["forbidden", "not_found", "internal_server_error"].each do |view|
+              File.exists?("./src/views/error/#{view}.slang").should be_true
+            end
+          end
```
marksiemers: I didn't test the contents of the view files, it seemed unnecessary.

spec/amber/cli/commands/generate/error_template_spec.cr
```diff
@@ -0,0 +1,35 @@
...
+          MainCommand.run %w(generate error a)
+
+          it "generates expected controller class" do
+            expected = CLIFixtures::ErrorControllerFixture.expected_error_controller
+            File.read("./src/controllers/error_controller.cr").should eq expected
+          end
+
+          it "generates a controller spec file" do
+            File.exists?("./spec/controllers/error_controller_spec.cr").should be_true
+          end
```
marksiemers: I didn't test the contents of the spec file, would that be valuable?

spec/support/fixtures/cli/generators/error/error_controller_fixture.cr
```diff
@@ -0,0 +1,27 @@
+module CLIFixtures
+  module ErrorControllerFixture
```
marksiemers: Is this name redundant - based on the namespacing?

src/amber/cli/templates/app/src/controllers/error_controller.cr.ecr
```diff
@@ -1,18 +0,0 @@
...
-  # end
-
-  # def not_found
-  #   render("404.<%= @language %>")
-  # end
-
-  # def internal_server_error
-  #   render("500.<%= @language %>")
-  # end
-end
```
marksiemers: We do not want this generated by default for `amber new app` anymore, correct?

src/amber/cli/templates/error/controller.cr
```diff
@@ -0,0 +1,21 @@
...
+    include Amber::CLI::Helpers
+    directory "#{__DIR__}/controller"
+
+    @name : String
+    @actions : Hash(String, String)
+    @language : String
+
+    def initialize(@name, actions)
+      @language = fetch_language
+      @actions = actions.map{|action| [action, "get"] }.to_h
```
marksiemers: I don't allow any verbs other than "get". Does that seem reasonable?

src/amber/cli/templates/error/controller.cr
```diff
@@ -0,0 +1,21 @@
...
+      add_plugs
+    end
+
+    private def add_plugs
+      add_plugs :web, "plug Amber::Pipe::Error.new"
```
marksiemers: Is this idempotent? I assumed it was, but haven't doublechecked yet.

src/amber/cli/templates/error/controller/src/controllers/error_controller.cr.ecr
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def not_found
+    render("not_found.<%= @language %>")
+  end
+
+  def internal_server_error
+    render("internal_server_error.<%= @language %>")
+  end
+end
```
marksiemers: I went back and forth between using that status codes as file names and this - any thoughts?

src/amber/cli/templates/helpers.cr
```diff
@@ -34,4 +34,21 @@ module Amber::CLI::Helpers
...
+  end
+
+  private def fetch_language : String
+    language = "slang"
+    if File.exists?(AMBER_YML) && (yaml = YAML.parse(File.read AMBER_YML))
+      language = yaml["language"]?.to_s if yaml["language"]?
+    end
+    return language
+  end
+
```
marksiemers: These two looked general enough to extract to this helper - though I didn't remove the methods from any pre-existing files.

src/amber/cli/templates/template.cr
```diff
@@ -96,6 +97,11 @@ module Amber::CLI
           raise "Auth not supported for crecto yet"
         end
         Auth.new(name, fields).render(directory, list: true, color: true)
+      when "error"
+        puts "Rendering Error Template"
+        actions = ["forbidden", "not_found", "internal_server_error"]
```
marksiemers: The initial implementation, there is no customization allowed - just these 3 actions will be generated.

marksiemers:
In regards to naming, even with a namespace, it felt weird to call something `Error` - so I chose `ErrorTemplate` but the directory name is still `error`.

Is it ok as is, or should it be consistent one way or the other?

faustinoaq:
@marksiemers Thans for this change, I used `error` following rails pattern, What other alternative name we can use ?

src/amber/cli/templates/error/controller/src/controllers/error_controller.cr.ecr
faustinoaq: Yeah, I think status codes are too implicit, error name is more readable

faustinoaq:
Also, just to remember:

phoenix

Something like Phoenix error pages will be amazing, we already have the back trace info, just need to add some interactivity. It will be a whole PR to achieve that.

spec/amber/cli/commands/generate/error_template_spec.cr
```diff
@@ -0,0 +1,61 @@
...
+              end
+
+              it "has no errors" do
+                result.should_not contain "Error in line"
+              end
+
+              it "has no failures" do
+                result.should_not contain "Failures"
+              end
+            end
```
marksiemers: Inception

marksiemers: I'm not sure if this should be the permanent place for this - but it was a good thing to learn - both that you can use a macro to skip certain tests with a flag, and you can actually test if a generated test suite works.

I would say it is best to just have one build that incorporates every generator, and that can make sure that the whole test suite can run.

@eliasjpr @elorest - What do you think about that approach? It could be done in another PR - one that resolves https://github.com/amberframework/amber/issues/342

spec/amber/cli/commands/generate/error_template_spec.cr
```diff
@@ -0,0 +1,61 @@
...
+            File.read("./spec/controllers/error_controller_spec.cr").should contain expected
+          end
+
+          it "generates view files" do
+            ["forbidden", "not_found", "internal_server_error"].each do |view|
+              File.exists?("./src/views/error/#{view}.slang").should be_true
+            end
+          end
+
+          {% if flag?(:run_builds) || flag?(:run_test_app_builds)%}
```
marksiemers: Probably need better names here - maybe `run_build_tests` - assuming we move this to a general area for one build to happen

spec/amber/cli/commands/generate/error_template_spec.cr
drujensen: what about using the same file structure as the spec but using `_fixture` extension?

`spec/support/fixtures/cli/commands/generate/*`

spec/amber/cli/commands/generate/error_template_spec.cr
drujensen: I think verifying the file exists is fine.  It will become brittle if we check for contents IMO.

spec/support/fixtures/cli/generators/error/error_controller_fixture.cr
drujensen: I like modules to follow the file structure. What about something like `Fixtures::CLI::Generate`?

src/amber/cli/templates/app/src/controllers/error_controller.cr.ecr
drujensen: correct.  This is to remove that since it isn't fully functional.

src/amber/cli/templates/error/controller.cr
drujensen: yup, seems reasonable.

src/amber/cli/templates/error/controller.cr
drujensen: yes, it should only add it once.

src/amber/cli/templates/error/controller/src/controllers/error_controller.cr.ecr
drujensen: I prefer the names as well

spec/amber/cli/commands/generate/error_template_spec.cr
marksiemers: That sounds reasonable.

@eliasjpr - Can you weigh in? After spending the time that you did re-organizing the tests I definitely want your thoughts on how to structure this.

With your change, will I even need line 2, or will the spec_helper take care of the fixtures being required?

drujensen:
Can we eliminate the need for a `name` when we run `amber g error`?

spec/support/fixtures/cli/generators/error/error_controller_fixture.cr
marksiemers: I like that convention.

I was following what was done here: https://github.com/amberframework/amber/pull/344/files#diff-38c1446038d633c0ece1152b8988e578R1

Maybe a different PR to consistently change the naming everywhere.

marksiemers:
@drujensen - I think eliminating `name` for the error generator is a good idea. It is the exception though as it should still be required for everything else. I wasn't sure of a good way to do that (still getting familiar with the code base, and the best way to do things in crystal).

This whole PR is going to need a rebase most likely. Are we targeting this for the release on Saturday?

drujensen:
When I ran the generator from your branch, it doesn't seem to kick in. 

Is the plan to add a redirect to the `Amber::Pipe:Error` handler or will this change create a new handler and perform the redirects?

https://github.com/amberframework/amber/blob/master/src/amber/router/pipe/error.cr#L9

marksiemers:
@drujensen - This PR is still very much a WIP. I hadn't thought through the design of handling the pipes. Do you have an opinion on it?
- Overwrite the existing handler when someone generates error?
- Create a different pipe that gets placed in front of the existing error pipe?

src/amber/cli/templates/error/view/src/views/error/internal_server_error.ecr.ecr
```diff
@@ -0,0 +1 @@
+<h1>500 - Internal Server Error</h1>
```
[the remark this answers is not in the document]
marksiemers: Do we have something like `Amber.env` or `Amber.env.development?`

I don't want to dump the backtrace unless it is the right environment.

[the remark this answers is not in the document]
faustinoaq: Yeah, we have `Amber.env.development?` see https://github.com/amberframework/amber/pull/302`

spec/amber/cli/commands/generate/error_template_spec.cr
eliasjpr: This is not require `require "../../../../support/fixtures/cli/generators/error/*"` as long it is being required by the fixtures file

spec/amber/cli/commands/generate/error_template_spec.cr
eliasjpr: This is fine, since it should be captured by the `master build spec` if the contents are valid

spec/amber/cli/commands/generate/error_template_spec.cr
eliasjpr: This can now be removed since we have the `master build`

spec/support/fixtures/cli/generators/error/error_controller_fixture.cr
eliasjpr: `Fixtures::CLI::Generate` 👍

src/amber/cli/templates/error/controller.cr
eliasjpr: Do we need to specify `plug` to `add_plugs`? 😕

src/amber/cli/templates/error/controller/src/controllers/error_controller.cr.ecr
eliasjpr: maybe we should loop thru these names to tidy up, WDYT?

marksiemers:
Will re-open after repo cleanup.

drujensen:
@marksiemers looking forward to seeing this re-opened and merged.

PR #350, opened 2017-11-01
merge docker files

drujensen:
### Description of the Change

This merges the docker files in the root and spec folders into one set in the root folder so we don't maintain 2 sets.

### Alternate Designs

maintain 2 sets

### Benefits

only need to maintain 1 set

### Possible Drawbacks

none that i can think of


faustinoaq:
Do we should merge docker files on app template too?

`Dockerfile.ecr` and `docker-compose.yml.ecr`

https://github.com/amberframework/amber/tree/master/src/amber/cli/templates/app

drujensen:
@elorest are you good with this?  would like to merge.

faustinoaq:
@drujensen Do you think we should merge `Dockerfile.ecr` and `docker-compose.yml.ecr` on app template too ? https://github.com/amberframework/amber/tree/master/src/amber/cli/templates/app

PR #351, opened 2017-11-02
cli: implement `db create/drop` for sqlite3

nathanj:
When running `amber db create` for an sqlite3 database, amber displays
the error:

  could not determine database name

This can be confusing to the user who doesn't know that `amber db
create` does not need to be run for sqlite3 databases. Display a
friendlier message.

Also implement `amber db drop` to delete the database.

### DIFF
--- a/spec/amber/cli/commands/generator_spec.cr
+++ b/spec/amber/cli/commands/generator_spec.cr
@@ -6,6 +6,34 @@ include CLIHelper
 include CLIFixtures
 
 module Amber::CLI
+  context "sqlite" do
+    cleanup
+    MainCommand.run ["new", TESTING_APP, "-d", "sqlite"]
+
+    describe "database" do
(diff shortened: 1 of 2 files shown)


faustinoaq:
Yeah, I can reproduce this, A workaround is using `amber migrate up`

faustinoaq:
@nathanj Yeah, please add some test for `amber db drop` and `amber db create`(_sqlite_) 😅

nathanj:
I added a test to `init_spec.cr` since I didn't know where else to put it. Let me know if there is a better place

eliasjpr:
@nathanj you can add the specs in `spec/amber/cli/commands/generator_spec.cr`. Please keep the tests clean and readable follow the pattern

nathanj:
Moved the tests and made them a little more spec-like I think (this is my first time using spec).

eliasjpr:
@nathanj Thank you we were just talking about your PR  🎉

Issue #339, opened 2017-10-29
Scaffold generated code gives a CSRF error on delete


satb:
### Description

Very new to Crystal and Amber. Was going through the guides and generated a basic project with one User model. The app started and everything was good but "delete" of the user isn't working. It fails with a "CSRF check failed." error upon delete.

Also delete musn't be a http get request but a http post. The network request on delete is a GET with out of the box scaffolding and the URL looks like this - "http://localhost:3000/users/1?_method=delete"

### Steps to Reproduce

1. amber new myapp
2. amber g scaffold User first_name:string last_name:string
3. shards install (took me a while to figure out why amber was complaining that "require "amber"" is failing. Looks like crystal shards now install to a local lib folder
4. amber w
5. open http://localhost:3000
6. Create a new user and try to delete the user

**Expected behavior:** http post request with _method of delete and no CSRF error

**Actual behavior:** http get request with an error that says "CSRF check failed."

**Reproduces how often:** Everytime

### Versions

❯ amber --version
Amber CMD (amberframework.org) - v0.3.0

### Additional Information

Just an out of the box scaffolding

faustinoaq:
I can reroduce it too, try commenting `Amber::Pipe::CSRF` on `:web` pipeline

```crystal
Amber::Server.configure do |app|
  pipeline :web do
    plug Amber::Pipe::Error.new
    plug Amber::Pipe::Logger.new
    plug Amber::Pipe::Session.new
    plug Amber::Pipe::Flash.new
    # plug Amber::Pipe::CSRF.new
  end

  routes :web do
    resources "/posts", PostController
    get "/", HomeController, :index
  end
end
```

elorest:
@faustinoaq Commenting out CSRF pipe will definitely work. But it doesn't explain why it stopped working. It might have something to do with https://github.com/amberframework/amber/pull/299

elorest:
@drujensen @faustinoaq @eliasjpr This brings up this issue once again. Should DELETE have to be a POST like it is in rails? 

Personally I don't like having to make it either a form or rely on a js library to make delete links work, so would vote for allowing DELETE through GET.

In the case that it's a GET though do we want to pass the CSRF through in a query param?

satb:
[...]

PR #356, opened 2017-11-02
Fixed csrf delete for generated scaffolding

elorest:
### Description of the Change

Fixes https://github.com/amberframework/amber/issues/339. 

Also uses link helpers instead of `<a href...>`

### Alternate Designs

At some point we might want to use a helper to generate a form that posts or a JS hack that does it onclick.

### Benefits

Protects from cross site scripting, so that javascript in other tabs can't post valid requests to domains associated with your cookies.

### DIFF
--- a/src/amber/cli/templates/scaffold/view/src/views/{{name}}/index.ecr.ecr
+++ b/src/amber/cli/templates/scaffold/view/src/views/{{name}}/index.ecr.ecr
@@ -24,9 +24,9 @@
 <% end -%>
         <td>
           <span>
-            <a class="btn btn-default btn-xs" href="/<%= @name %>s/<%="<"%>%= <%= @name %>.id %>">read</a>
-            <a class="btn btn-success btn-xs" href="/<%= @name %>s/<%="<"%>%= <%= @name %>.id %>/edit">edit</a>
-            <a class="btn btn-danger btn-xs" href="/<%= @name %>s/<%="<"%>%= <%= @name %>.id %>?_method=delete" onclick="return confirm('Are you sure?');">delete</a>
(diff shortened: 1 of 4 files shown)


faustinoaq:
@elorest Can Amber have something like `link "Home", to: HomeController, :index` ?

With params `link "Home", to: HomeController, :index, with: {id: id}`

Instead of using direct links ?

elorest:
@faustinoaq it should be able to. @eliasjpr was working on `url_for` a couple of months ago but it sort of ended up going in a different direction. This will also fix redirect. That information is all available in the router so it wouldn't be that hard to do if you wanted to take a stab at it. It should be in amber core instead of jasper_helpers though since it would be needed by both jasper and amber redirect.

elorest:
`link "Home", to: HomeController, :index, with: {id: id}` syntactically won't work since it has a named params `to:` and then a none named param for `:index`

elorest:
We can improve that later though. This just fixes a bug and should be merged asap in my opinion.

elorest:
Whether is should override GET with DELETE or use POST is a conversation we can have another day. However this is not a security concern since CSRF isn't concerned with man in the middle attacks or copy and pasting links. Please see conversation on issue #339 or read up on CSRF mitigation.
https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)

marksiemers:
@elorest - I agree this should get merged in ASAP. If this gets refactored / more work is done with CSRF and DELETE, we should write some tests around it. AFAIK, this was happening in production and the test suite didn't catch it.

Is that waiting on the ability to do capybara type testing?

PR #357, opened 2017-11-03
Organizes Controllers Helper Methods

eliasjpr:
- Moves redirect into the Helpers subdirectory in controller
- Updates require path
- Renames Helper::Tags to Helpers::CSRF


src/amber/controller/helpers/tags.cr
```diff
@@ -1,5 +1,5 @@
 module Amber::Controller::Helpers
-  module Tag
+  module CSRF
```
elorest: Probably name the file csrf.cr as well if you're changing the name of the module.

PR #362, opened 2017-11-05
[Controller] Adds Route Helper Methods to Controller

eliasjpr:
> As a developer I want to have access to the current matched
route information such as the Action, Controller, Resource
and Scope accessed.

This adds a Route helper file
- Includes the following methods route_action, route_controller,
route_resource, route_scope in the controller

### Alternate Design

I thought of using a different name such as action_name, controller_name
but it seemed more explicit to have `route_` prefix to hint the user where is
this information is coming from.

### DIFF
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -27,10 +27,9 @@ module Amber::Controller
       controller.responds_to?(:client_ip).should eq true
       controller.responds_to?(:request).should eq true
       controller.responds_to?(:response).should eq true
-      controller.responds_to?(:route_action).should eq true
+      controller.responds_to?(:action_name).should eq true
       controller.responds_to?(:route_resource).should eq true
       controller.responds_to?(:route_scope).should eq true
-      controller.responds_to?(:route_controller).should eq true
       controller.responds_to?(:controller_name).should eq true
     end
 
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -27,6 +27,11 @@ module Amber::Controller
       controller.responds_to?(:client_ip).should eq true
       controller.responds_to?(:request).should eq true
       controller.responds_to?(:response).should eq true
+      controller.responds_to?(:route_action).should eq true
+      controller.responds_to?(:route_resource).should eq true
+      controller.responds_to?(:route_scope).should eq true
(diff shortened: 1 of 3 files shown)


src/amber/controller/helpers/route.cr
```diff
@@ -0,0 +1,23 @@
...
+    def route_scope
+      context.request_handler.scope
+    end
+
+    def route_controller
+      context.request_handler.controller
+    end
+
+    def controller_name
+      self.class.name.gsub(/Controller/i, "")
```
marksiemers: For this regex, I would anchor to the end. And does the `i` make it case-insensitive, is that necessary?

https://play.crystal-lang.org/#/r/31bu

marksiemers: Would people want the controller name downcased?

spec/amber/controller/base_spec.cr
```diff
@@ -27,6 +27,11 @@ module Amber::Controller
       controller.responds_to?(:client_ip).should eq true
       controller.responds_to?(:request).should eq true
       controller.responds_to?(:response).should eq true
+      controller.responds_to?(:route_action).should eq true
+      controller.responds_to?(:route_resource).should eq true
+      controller.responds_to?(:route_scope).should eq true
+      controller.responds_to?(:route_controller).should eq true
```
marksiemers: Do all of these need to be prefixed with `route_`?

src/amber/controller/helpers/route.cr
faustinoaq: Yeah, I think we should use `/Controller$/`

eliasjpr: You guys are correct I will also fix the regex. I looked at the rails implementation 

https://apidock.com/rails/ActionController/Metal/controller_name/class

spec/amber/controller/base_spec.cr
eliasjpr: I prefixed to avoid name collisions with methods define by the user. I am open tomsuggestions

eliasjpr: I prefixed to avoid name collisions with methods define by the user. I am open tomsuggestions

src/amber/controller/helpers/route.cr
elorest: Use sub instead of gsub since we only want to replace once.

Issue #353, opened 2017-11-02
redirect_to raise Nil assertion failed


faustinoaq:
### Description

Trying documentation examples here: https://amberframework.org/guides/getting-started/controllers/redirecting.md#controllers-redirecting

```
redirect_to controller: :HomeController, action: :index
```

Output:

```
ERROR: Nil assertion failed (Exception)
```

### Steps to Reproduce

1. create new amber project
2. try to create a redirection using `redirect_to` and symbols following documentation
3. Test redirection

**Expected behavior:** successful redirection

**Actual behavior:** `Error: Nil Assertion failed`

**Reproduces how often:** 100%

### Versions

Master `0.3.0`

### Additional Information

Full error output:

```
ERROR: Nil assertion failed (Exception)
0x6fd3ff2a10: not_nil! at /usr/lib/crystal/class.cr 65:0
0x6fd40e2748: payload at /home/main/Projects/reddit_clone/lib/radix/src/radix/node.cr 0:5
0x6fd40dfdc2: match_by_controller_action at /home/main/Projects/reddit_clone/lib/amber/src/amber/router/router.cr 60:44
0x6fd40dfd2a: match_by_controller_action at /home/main/Projects/reddit_clone/lib/amber/src/amber/router/router.cr 54:7
0x6fd41588b3: from_controller_action at /home/main/Projects/reddit_clone/lib/amber/src/amber/controller/redirect.cr 32:7
0x6fd410fef4: redirect_to:controller:action at /home/main/Projects/reddit_clone/lib/amber/src/amber/controller/redirect.cr 12:7
0x6fd410efaf: create at /home/main/Projects/reddit_clone/src/controllers/post_controller.cr 26:7
0x6fd410782c: call at /home/main/Projects/reddit_clone/lib/amber/src/amber/router/route.cr 255:3
0x6fd411cb06: process_request at /home/main/Projects/reddit_clone/lib/amber/src/amber/router/context.cr 122:5
0x6fd3fab706: ~procProc(HTTP::Server::Context, Nil) at /usr/lib/crystal/concurrent.cr 29:3
0x6fd4102e24: call_next at /usr/lib/crystal/http/server/handler.cr 255:3
0x6fd4102518: call at /home/main/Projects/reddit_clone/src/handlers/authenticate.cr 10:7
0x6fd4101dde: call_next at /usr/lib/crystal/http/server/handler.cr 24:7
[...]

PR #363, opened 2017-11-05
[Redirect] Allows for Class and Symbol Controller name

eliasjpr:
Issue: https://github.com/amberframework/amber/issues/353

> As a developer I would like to pass and full controller name to a
redirect as a class or symbol and allow for the redirection to happen

Example

`redirect_to controller: :HomeController, action: :index`

- Would be nice to accept HelloController::Index or
Admin::UserController:Index

### DIFF
--- a/src/amber/router/route.cr
+++ b/src/amber/router/route.cr
@@ -48,13 +48,5 @@ module Amber
       end
       {result, params}
     end
-
-    def match?(controller : Class, action : Symbol)
-      self.controller.downcase == controller.to_s.downcase && self.action == action
-    end
-
-    def match?(controller : Symbol, action : Symbol)
-      self.controller.downcase == "#{controller}controller" && self.action == action
-    end
   end
 end
--- a/src/amber/router/route.cr
+++ b/src/amber/router/route.cr
@@ -49,7 +49,11 @@ module Amber
       {result, params}
     end
 
-    def match?(controller, action)
+    def match?(controller : Class, action : Symbol)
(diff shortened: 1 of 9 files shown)


eliasjpr:
That would be a great idea. We should also have the ‘link_to’ jasper helpers user class name

elorest:
We need to create url_for and have link_to and redirect_to translate links with that like in rails.

eliasjpr:
A couple of changes introduced in the last 2 commits. I have decided to simplified the redirect api. 

There is only 3 possible ways to redirect now:
- Redirecting to full URL will be redirect_to "http://google.com", params. String based redirect are assumed that the user is passing the full path or url to redirect.
- Redirecting controller action `redirect_to PostController, :index`, `redirect_to :post, :index`
- Redirecting withing same controller `redirect_to :index` whichever controller this is called from it will be implicitly added.

Why?

- This keeps things simple and coherent and simple and short.
- Supporting `redirect_to :PostController, :index` or `redirect_to "PostController", :index` makes things messy and confusing
- People can pass controller name as symbols without the "controller" part, keeps it short.

src/amber/router/route.cr
```diff
@@ -49,7 +49,11 @@ module Amber
       {result, params}
     end
 
-    def match?(controller, action)
+    def match?(controller : Class, action : Symbol)
```
eliasjpr: This is separated into 2 method so we don't have to perform checks or substitutions for `/Controller$/`

PR #364, opened 2017-11-05
Improved settings and fixed tests

elorest:
### Description of the Change

Fixed Settings so that it is initialized. This allows it to be tested without previous tests effecting it's state randomly. Also conforms to best practices.

### Benefits

Allows for better tests. A few of the settings tests had to be commented out because settings were being overwritten by previous tests. It also makes more sense to instantiate settings when server is started instead of having a preexisting pseudo singleton.

### Possible Drawbacks

None.

### DIFF
--- a/spec/amber/amber_spec.cr
+++ b/spec/amber/amber_spec.cr
@@ -18,7 +18,7 @@ describe Amber do
   describe Amber::Server do
     describe ".configure" do
       it "loads environment settings from test.yml" do
-        settings = Amber::Server.settings
+        settings = Amber::Settings.new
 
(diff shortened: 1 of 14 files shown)


spec/amber/amber_spec.cr
```diff
@@ -37,6 +36,7 @@ describe Amber do
       end
```
marksiemers: It seems like this should be a test for `Amber::Settings` rather than `Amber::Server.configure`

elorest: Done.

PR #367, opened 2017-11-06
Respond_with uri extensions path.[json|xml|text|txt|html]

elorest:
### Description of the Change

A few small changes to allow extensions to work for choosing content type. I've also modified the response priority so that it will return a 406 if you only ask for one content type and it's not defined. If you ask for more it will return the first match or the first defined response. 

### Youtube Example

https://www.youtube.com/watch?v=6KNjWDRUo_c

### DIFF
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -84,7 +84,6 @@ module Amber::Controller
         ResponsesController.new(context).index.should eq expected_result
       end
 
-
       it "responds with json for path.json" do
         expected_result = %({"type":"json","name":"Amberator"})
         context.request.path = "/response/1.json"
@@ -116,6 +115,12 @@ module Amber::Controller
         context.response.status_code.should eq 406
       end
 
+      it "respond_with default or accept header request if extension is invalid" do
+        expected_result = "<html><body><h1>Elorest <3 Amber</h1></body></html>"
+        context.request.path = "/response/1.texas"
+        context.request.headers["Accept"] = "text/html"
+        ResponsesController.new(context).index.should eq expected_result
+      end
     end
 
     describe "#render" do
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -58,54 +58,72 @@ module Amber::Controller
       it "respond_with html as default option" do
         expected_result = "<html><body><h1>Elorest <3 Amber</h1></body></html>"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "text/html"
+        context.response.status_code.should eq 200 
       end
 
       it "respond_with html" do
         expected_result = "<html><body><h1>Elorest <3 Amber</h1></body></html>"
         context.request.headers["Accept"] = "text/html"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "text/html"
+        context.response.status_code.should eq 200 
       end
 
       it "responds with json" do
         expected_result = %({"type":"json","name":"Amberator"})
         context.request.headers["Accept"] = "application/json"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "application/json"
+        context.response.status_code.should eq 200 
       end
 
       it "responds with xml" do
         expected_result = "<xml><body><h1>Sort of xml</h1></body></xml>"
         context.request.headers["Accept"] = "application/xml"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "application/xml"
+        context.response.status_code.should eq 200 
       end
 
       it "responds with text" do
         expected_result = "Hello I'm text!"
         context.request.headers["Accept"] = "text/plain"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "text/plain"
+        context.response.status_code.should eq 200 
       end
 
       it "responds with json for path.json" do
         expected_result = %({"type":"json","name":"Amberator"})
         context.request.path = "/response/1.json"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "application/json"
+        context.response.status_code.should eq 200 
       end
 
       it "responds with xml for path.xml" do
         expected_result = "<xml><body><h1>Sort of xml</h1></body></xml>"
         context.request.path = "/response/1.xml"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "application/xml"
+        context.response.status_code.should eq 200 
       end
 
       it "responds with text for path.txt" do
         expected_result = "Hello I'm text!"
         context.request.path = "/response/1.txt"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "text/plain"
+        context.response.status_code.should eq 200 
       end
 
       it "responds with text for path.text" do
         expected_result = "Hello I'm text!"
         context.request.path = "/response/1.text"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "text/plain"
+        context.response.status_code.should eq 200 
       end
 
       it "responds with 406 for path.text when text hasn't been defined" do
@@ -115,11 +133,22 @@ module Amber::Controller
         context.response.status_code.should eq 406
       end
 
-      it "respond_with default or accept header request if extension is invalid" do
+      it "respond with default if extension is invalid and accepts isn't defined" do
         expected_result = "<html><body><h1>Elorest <3 Amber</h1></body></html>"
         context.request.path = "/response/1.texas"
         context.request.headers["Accept"] = "text/html"
         ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "text/html"
+        context.response.status_code.should eq 200 
+      end
+
+      it "responds with or accept header request if extension is invalid" do
+        expected_result = %({"type":"json","name":"Amberator"})
+        context.request.headers["Accept"] = "application/json"
+        context.request.path = "/response/1.texas"
+        ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "application/json"
+        context.response.status_code.should eq 200 
       end
     end
 
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -109,7 +109,7 @@ module Amber::Controller
       end
 
       it "responds with 406 for path.text when text hasn't been defined" do
-        expected_result = "Response unexceptable."
+        expected_result = "Response Not Acceptable."
         context.request.path = "/response/1.text"
         ResponsesController.new(context).show.should eq expected_result
         context.response.status_code.should eq 406
--- a/spec/amber/controller/base_spec.cr
+++ b/spec/amber/controller/base_spec.cr
@@ -83,6 +83,39 @@ module Amber::Controller
         context.request.headers["Accept"] = "text/plain"
         ResponsesController.new(context).index.should eq expected_result
       end
+
+
+      it "responds with json for path.json" do
+        expected_result = %({"type":"json","name":"Amberator"})
+        context.request.path = "/response/1.json"
+        ResponsesController.new(context).index.should eq expected_result
+      end
+
+      it "responds with xml for path.xml" do
+        expected_result = "<xml><body><h1>Sort of xml</h1></body></xml>"
+        context.request.path = "/response/1.xml"
+        ResponsesController.new(context).index.should eq expected_result
+      end
+
+      it "responds with text for path.txt" do
+        expected_result = "Hello I'm text!"
+        context.request.path = "/response/1.txt"
+        ResponsesController.new(context).index.should eq expected_result
+      end
+
+      it "responds with text for path.text" do
+        expected_result = "Hello I'm text!"
+        context.request.path = "/response/1.text"
+        ResponsesController.new(context).index.should eq expected_result
+      end
+
+      it "responds with 406 for path.text when text hasn't been defined" do
+        expected_result = "Response unexceptable."
+        context.request.path = "/response/1.text"
+        ResponsesController.new(context).show.should eq expected_result
+        context.response.status_code.should eq 406
+      end
+
     end
 
     describe "#render" do
--- a/src/amber/controller/helpers/responders.cr
+++ b/src/amber/controller/helpers/responders.cr
@@ -9,8 +9,8 @@ module Amber::Controller::Helpers
         xml:  "application/xml",
       }
 
-      REGEX                     = /\.(#{TYPE.keys.join("|")})$/
-      COMMA_OR_COMMASPACE_REGEX = /,|,\s/
+      TYPE_EXT_REGEX         = /\.(#{TYPE.keys.join("|")})$/
+      ACCEPT_SEPARATOR_REGEX = /,|,\s/
 
       @requested_responses : Array(String)
       @available_responses = Hash(String, String).new
@@ -41,7 +41,7 @@ module Amber::Controller::Helpers
       end
 
       def type
-        select_type.to_s
+        (@type ||= select_type).to_s
       end
 
       def body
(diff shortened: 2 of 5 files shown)


src/amber/controller/helpers/responders.cr
```diff
@@ -51,24 +61,25 @@ module Amber::Controller::Helpers
...
     end
 
     protected def respond_with(&block)
-      content = with Content.new(requested_responses) yield
+      content = with Response.new(requested_responses) yield
       if content.body
         set_response(body: content.body, status_code: 200, content_type: content.type)
       else
-        set_response(body: "Response unexceptable.", status_code: 406, content_type: Content::TYPE[:text])
+        set_response(body: "Response unexceptable.", status_code: 406, content_type: Response::TYPE[:text])
```
marksiemers: Should this be - "Response Not Acceptable" ?

src/amber/router/router.cr
```diff
@@ -81,7 +81,12 @@ module Amber
       end
 
       def match(http_verb, resource) : Radix::Result(Amber::Route)
-        @routes.find build_node(http_verb, resource)
+        result = @routes.find build_node(http_verb, resource)
+        if result.found?
+          result
+        else
+          @routes.find build_node(http_verb, resource.sub(/\.[^$\/]+$/, ""))
```
marksiemers: Can this regex be named with a constant to indicate what it is for? It is meant to chop off the extension, right? .json or .text, etc.

marksiemers:
I know this isn't part of this change, but https://github.com/amberframework/amber/pull/367/files#diff-e43f8575da93c21cf741fdd422c2b60cR78
Should be changed from "Response unexceptable." to "Response Not Acceptable."

src/amber/controller/helpers/responders.cr
```diff
@@ -51,8 +61,9 @@ module Amber::Controller::Helpers
 
     private def requested_responses
       req_responses = Array(String).new
-
-      if (accept = context.request.headers["Accept"]?) && !accept.empty?
+      if (ext = request.path.match(Content::REGEX).try(&.[1]))
```
eliasjpr: What about using `.last` instead of `&.[1]` since is more readable?

src/amber/controller/helpers/responders.cr
```diff
@@ -51,8 +61,9 @@ module Amber::Controller::Helpers
...
+        req_responses << Content::TYPE[ext]
+      elsif (accept = context.request.headers["Accept"]?) && !accept.empty?
         accepts = accept.split(";").first?.try(&.split(/,|,\s/))
```
eliasjpr: Make this regex `/,|,\s/` a constant that shows intent

spec/amber/controller/base_spec.cr
```diff
@@ -83,6 +83,39 @@ module Amber::Controller
         context.request.headers["Accept"] = "text/plain"
         ResponsesController.new(context).index.should eq expected_result
       end
+
```
eliasjpr: Can you add a test for an invalid (unsupported) extension?

src/amber/router/params_parser.cr
```diff
@@ -78,7 +78,13 @@ module Amber::Router
     end
 
     def merge_route_params
-      route_params.each { |k, v| params[k] = v }
+      route_params.each do |k, v|
+        if k == route_params.keys.last
```
eliasjpr: This 2 lines should probably be in its own method, called `format`. I forgot to tell you in the previous PR that we have already define a similar method in the context https://github.com/amberframework/amber/blob/master/src/amber/router/context.cr#L72

src/amber/controller/helpers/responders.cr
elorest: A regex match returns a hash instead of an array, 1 is just a key for the first capture group. Similar to $1 in perl.

src/amber/controller/helpers/responders.cr
elorest: To me 4 regex characters are easier to read than tracking down the definition of a constant, but sure.

src/amber/router/params_parser.cr
elorest: I `format` currently being used for something? My need is just to remove 5 ext's from the end of the last path_param. I could extract it to a method but I don't think moving it to context makes sense.

elorest: If it don't remove these then paths like `/pets/1.json` try looking up an id of `1.json` in Granite which fails. This isn't where I'm detecting the format, since it will only sometimes be on a param I have to parse it from the path, for responses.

marksiemers:
Per discussion with @elorest - invoking the methods without blocks is fast. If it becomes a problem, we can revisit using blocks/procs. For this PR, eager loaded strings are good.

spec/amber/controller/base_spec.cr
elorest: Yeah. I'll commit that in a second. I already have a test for a supported but undefined extension.

src/amber/controller/helpers/responders.cr
marksiemers: Does it break with `request.path.match(Content::REGEX)[1]?` ?

src/amber/router/params_parser.cr
```diff
@@ -29,7 +29,7 @@ module Amber::Router
     def parse_params
       parse_part(request.query)
       if content_type = request.headers["Content-Type"]?
-        parse_multipart if content_type.try(&.starts_with?(MULTIPART_FORM))
+          parse_multipart if content_type.try(&.starts_with?(MULTIPART_FORM))
```
eliasjpr: remove extra identation

src/amber/router/params_parser.cr
```diff
@@ -78,13 +78,18 @@ module Amber::Router
...
-        else
-          params[k] = v
-        end
+      route_params_without_ext.each do |k, v|
+        params[k] = v
+      end
+    end
+
+    def route_params_without_ext
+      rparams = route.params
```
eliasjpr: you can use `_params`or `route_params` instead of `rparams`, the latter can be confusing. 

Thanks for splitting it into 2 separate methods.

src/amber/controller/helpers/responders.cr
```diff
@@ -61,10 +62,11 @@ module Amber::Controller::Helpers
 
     private def requested_responses
       req_responses = Array(String).new
-      if (ext = request.path.match(Content::REGEX).try(&.[1]))
-        req_responses << Content::TYPE[ext]
+      path_ext = request.path.match(Content::REGEX).try(&.[1])
```
eliasjpr: Non blocking comment: This would probably read better in a private method called format.

```crystal
def requested_format
  request.path.match(Content::REGEX).try(&.[1])
end
```

elorest: Possibly but it's only one source of format. The private method that it's in is already tasks with finding the requested response types.

src/amber/router/params_parser.cr
elorest: I just wanted something short and temporary what would overwrite anything else. To me rparams looked better than _params. WDYT?

src/amber/controller/helpers/responders.cr
```diff
@@ -35,13 +45,14 @@ module Amber::Controller::Helpers
       end
```
marksiemers: Can we move `@type ||=` into this method? rather than needing the begin statement down below.
```crystal
def type
  @type ||= select_type.to_s
end
```
I like the convention of 
```crystal
def property_name
  @property_name ||= calculate_property_name
end
```
Of course, 'calculate' can be 'select' or 'generate' - what ever verb makes sense.

src/amber/controller/helpers/responders.cr
```diff
@@ -35,13 +45,14 @@ module Amber::Controller::Helpers
...
-        @available_responses[type]
+        @available_responses[type]?
       end
 
       private def select_type
         @type ||= begin
           raise "You must define at least one response_type." if @available_responses.empty?
-          @requested_responses << @available_responses.keys.first
+          # NOTE: If only one response is requested don't return something else.
+          @requested_responses << @available_responses.keys.first if @requested_responses.size != 1
```
marksiemers: So, this is "non-strict" mode - where even if they didn't explicitly request the type we return the first available response (if they had more than one thing in the 'accept' header), correct?

src/amber/controller/helpers/responders.cr
```diff
@@ -1,7 +1,17 @@
...
+      TYPE = {
+        html: "text/html",
+        json: "application/json",
+        txt:  "text/plain",
+        text: "text/plain",
+        xml:  "application/xml",
+      }
+
+      REGEX                     = /\.(#{TYPE.keys.join("|")})$/
+      COMMA_OR_COMMASPACE_REGEX = /,|,\s/
```
marksiemers: `ACCEPT_SEPARATOR_REGEX`

src/amber/controller/helpers/responders.cr
```diff
@@ -1,7 +1,17 @@
...
-      TYPE = {html: "text/html", json: "application/json", text: "text/plain", xml: "application/xml"}
+      TYPE = {
+        html: "text/html",
+        json: "application/json",
+        txt:  "text/plain",
+        text: "text/plain",
+        xml:  "application/xml",
+      }
+
+      REGEX                     = /\.(#{TYPE.keys.join("|")})$/
```
marksiemers: `CONTENT_TYPE_REGEX`

src/amber/controller/helpers/responders.cr
```diff
@@ -51,9 +62,11 @@ module Amber::Controller::Helpers
...
     private def requested_responses
       req_responses = Array(String).new
-
-      if (accept = context.request.headers["Accept"]?) && !accept.empty?
-        accepts = accept.split(";").first?.try(&.split(/,|,\s/))
+      path_ext = request.path.match(Content::REGEX).try(&.[1])
+      if (path_ext)
+        req_responses << Content::TYPE[path_ext]
+      elsif (accept = context.request.headers["Accept"]?) && !accept.empty?
+        accepts = accept.split(";").first?.try(&.split(Content::COMMA_OR_COMMASPACE_REGEX))
```
marksiemers: This is my inner ruby-ist thinking out loud.

Take it or leave it.

```crystal
private def type_from_path_extension : String | Nil
  return @type_from_path_extension if @type_from_path_extension_checked
  @type_from_path_extension_checked = true
  @type_from_path_extension = request.path.match(Content::REGEX).try(&.[1])
end

private def accepts_from_headers : Array(String)
  @accepts_from_headers ||= parse_accepts_from_headers
end

private def parse_accepts_from_headers : Array(String)
  accept_header = context.request.headers["Accept"]?
  if accept_header
    accept_header.split(";").first?.try(&.split(Content::COMMA_OR_COMMASPACE_REGEX))
  else
    [] of String
  end
end

private def requested_responses : Array(String)
  return [type_from_path_extension] unless type_from_path_extension.nil?
  accepts_from_headers
end
```

spec/amber/controller/base_spec.cr
elorest: Done.

spec/amber/controller/base_spec.cr
```diff
@@ -83,6 +83,44 @@ module Amber::Controller
...
+      end
+
+      it "responds with text for path.text" do
+        expected_result = "Hello I'm text!"
+        context.request.path = "/response/1.text"
+        ResponsesController.new(context).index.should eq expected_result
+      end
+
+      it "responds with 406 for path.text when text hasn't been defined" do
+        expected_result = "Response unexceptable."
```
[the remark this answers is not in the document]
elorest: Done.

spec/amber/controller/base_spec.cr
```diff
@@ -83,6 +83,44 @@ module Amber::Controller
...
+        ResponsesController.new(context).show.should eq expected_result
+        context.response.status_code.should eq 406
+      end
+
+      it "respond_with default or accept header request if extension is invalid" do
+        expected_result = "<html><body><h1>Elorest <3 Amber</h1></body></html>"
+        context.request.path = "/response/1.texas"
+        context.request.headers["Accept"] = "text/html"
+        ResponsesController.new(context).index.should eq expected_result
+      end
```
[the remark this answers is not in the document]
elorest: I think so. I'll check it I guess. Since they're all set at the some time I didn't think it was important to set that since the response was right.

src/amber/controller/helpers/responders.cr
elorest: Done

src/amber/controller/helpers/responders.cr
elorest: Done.

src/amber/controller/helpers/responders.cr
elorest: I would have done that but I needed to return @type.to_s so that @type wouldn't have the possibility of being Nil.

src/amber/controller/helpers/responders.cr
elorest: While I feel that if json is the only thing available it should be return even if the browser is asking for html, it seems like a different case if someone asks for `domain.com/cats.json`. In that case it will return 406 if json doesn't exist.

spec/amber/controller/base_spec.cr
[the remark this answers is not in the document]
elorest: For some reason I'm getting an error that `content_type` doesn't exist on response. When I try to test it's value.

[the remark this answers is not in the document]
elorest: I can set it though.

src/amber/controller/helpers/responders.cr
elorest: Is this easier to read to you? I do like the idea of not defining the Array in the beginning.

elorest: I guess the main thing for me here is avoiding as many branches and memory allocations as I can.

marksiemers: Maybe this:
```crystal
private def type_from_path_extension : String | Nil
  request.path.match(Content::REGEX).try(&.[1])
end

private def parse_accepts_from_headers : Array(String)
  accept_header = context.request.headers["Accept"]?
  if accept_header
    accept_header.split(";").first?.try(&.split(Content::COMMA_OR_COMMASPACE_REGEX))
  else
    [] of String
  end
end

private def requested_responses : Array(String)
  if (type_from_path = type_from_path_extension)
    [type_from_path]
  else
    parse_accepts_from_headers
  end
end
```

spec/amber/controller/base_spec.cr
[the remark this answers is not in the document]
eliasjpr: In base_spec we should be verifying the context `context.response` as well. This is `ResponsesController.new(context).index.should eq expected_result` totally valid, at least confirm the content-type and status code are correct. For instance 

```crystal
context.response.body.should
context.response.status_code.should
context.response.headers["Content-Type"].should
```
I think it is important that is done this way because something else could be changing the response. This is how I found the issue with the `redirect` :)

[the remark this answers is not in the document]
eliasjpr: Unless you are covering testing status code and content-type in the unit test for Responders::Content context.response.status_code.should
context.response.headers["Content-Type"].should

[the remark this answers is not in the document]
elorest: Good point. I was put off by the fact that there was no getter for content_type. Makes sense that it's getting written to the header though. I'll test that.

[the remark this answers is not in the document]
marksiemers: `context.response.headers["Content-Type"]?` will report a failure with helpful info rather than an error when `nil`

src/amber/controller/helpers/responders.cr
elorest: I did come up with a solution here that I think we'll both like.

src/amber/controller/helpers/responders.cr
```diff
@@ -31,40 +41,46 @@ module Amber::Controller::Helpers
...
+      extension_request_type || accepts_request_type || [] of String
     end
 
     protected def respond_with(&block)
       content = with Content.new(requested_responses) yield
       if content.body
         set_response(body: content.body, status_code: 200, content_type: content.type)
       else
-        set_response(body: "Response unexceptable.", status_code: 406, content_type: Content::TYPE[:text])
+        set_response(body: "Response Not Acceptable.", status_code: 406, content_type: Content::TYPE[:text])
```
faustinoaq: Hi @elorest I didn't see it 😅 

Why not `raise Amber::Exception::NoAcceptable` so we can use `Amber::Pipe::Error` here 👉 https://github.com/amberframework/amber/blob/master/src/amber/router/pipe/error.cr#L19 and be hable to customize the 406 reponse. WDYT?

elorest: I was under the impression that we didn't want to raise exceptions because it was slower or something. I remember @eliasjpr saying that was the reason he made halt instead of just raising and exception to handle later. Also since I'm setting a response anyone it sort of seems natural to set the other response and message.

PR #369, opened 2017-11-08
Created params object that can parse json when asked too.

elorest:
### Description of the Change

Added a `Amber::Router::Params`.

This gives us a `#json(key)` method as well as allowing keys to be either symbols or strings.

Params work fine but validations seem messed up. I played with them a bit but not much progress

### DIFF
--- a/src/amber/validations/params.cr
+++ b/src/amber/validations/params.cr
@@ -1,14 +1,14 @@
 module Amber::Validators
   # Holds a validation error message
-  record Error, param : String, value : Amber::Router::Params, message : String
+  record Error, param : String, value : String?, message : String
 
   # This struct holds the validation rules to be performed
   class BaseRule
-    getter predicate : (Amber::Router::Params -> Bool)
+    getter predicate : (String -> Bool)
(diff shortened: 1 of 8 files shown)


src/amber/validations/params.cr
```diff
@@ -1,19 +1,19 @@
...
-  record Error, param : String, value : Amber::Router::ParamsType, message : String
+  record Error, param : String, value : Amber::Router::Params, message : String
 
   # This struct holds the validation rules to be performed
   class BaseRule
-    getter predicate : (Amber::Router::ParamsType -> Bool)
+    getter predicate : (Amber::Router::Params -> Bool)
     getter field : String
-    getter value : Amber::Router::ParamsType
+    getter value = Amber::Router::Params.new
```
elorest: Before I changed this I was getting an error about it not being set by initialize.

elorest:
All of my Real Life tests pass.

eliasjpr:
It wsan't that bad :)

elorest:
Thanks.

PR #377, opened 2017-11-10
Set ENV["PORT"] to be used by default if defined.

elorest:
### Description of the Change

Fixed small bug where `ENV["PORT"]` wasn't being used by default. I thought that it would be good to leave this more configurable, but agree that this might be confusing to users.

### DIFF
--- a/src/amber/scripts/environment_loader.cr
+++ b/src/amber/scripts/environment_loader.cr
@@ -41,7 +41,7 @@ str = String.build do |s|
   #s.puts %(@log.level = #{settings["log_level"]? || "::Logger::INFO"})
   s.puts %(@color = #{settings["color"]? == nil ? true : settings["color"]})
   s.puts %(@redis_url = "#{settings["redis_url"]? || "redis://localhost:6379"}")
-  s.puts %(@port = #{settings["port"]? || 3000})
+  s.puts %(@port = #{ENV["PORT"]? || settings["port"]? || 3000})
   s.puts %(@host = "#{settings["host"]? || "127.0.0.1"}")
   s.puts %(@secret_key_base = "#{settings["secret_key_base"]? || SecureRandom.urlsafe_base64(32)}")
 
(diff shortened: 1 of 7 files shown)


src/amber/scripts/environment_loader.cr
```diff
@@ -41,7 +41,7 @@ str = String.build do |s|
   #s.puts %(@log.level = #{settings["log_level"]? || "::Logger::INFO"})
   s.puts %(@color = #{settings["color"]? == nil ? true : settings["color"]})
   s.puts %(@redis_url = "#{settings["redis_url"]? || "redis://localhost:6379"}")
-  s.puts %(@port = #{settings["port"]? || 3000})
+  s.puts %(@port = #{ENV["PORT"]? || settings["port"]? || 3000})
```
eliasjpr: This should be `(ENV["PORT"] ||= "3000").to_i`

You can just do this on each of the environment YAMLs

```yaml
secret_key_base: <%= SecureRandom.urlsafe_base64(32) %>
port:  <%= (ENV["PORT"] ||= "3000").to_i %>
name: <%= @name %>
log: "::Logger.new(STDOUT)"
log_level: "::Logger::INFO"
color: true
...
```

elorest: In this file it shouldn't be `(ENV["PORT"] ||= "3000").to_i` Just in the environment files. Do you prefer having it in each of the env yaml files? That was my original opinion but started second guessing that yesterday after reading your comments on it. 

Would it maybe be better to let port be overridden by ENV["PORT"] regardless of what's in the env.yml?

sam0x17: @elorest I would expect the environment variable to override env.yml personally

sam0x17:
This would have saved me some headaches with docker two weeks ago lol

src/amber/scripts/environment_loader.cr
elorest: @sam0x17 We'll it will be setup that way by default in the env.yml's but it gives you the option to change it to a different ENV VAR if you wanted. The default will still have PORT override though it will just be configured in development.yml, prod, test etc.

elorest: Either way this will be fixed for you.

marksiemers:
@elorest - Is this waiting on anything?

PR #386, opened 2017-11-13
Ep/move properties to server

eliasjpr:
- [Server] Moves Router and Handler out off Settings 
  - Moves the Pipeline instance from Amber::Settings into the Server instance
  - Moves the Router instance from Amber::Settings into the Server instance
- [Server] Moves WebSocketAdapter and Removes Settings class method
- [Server] Move websockets from Configuration to Server
  - Replace settings.handler with just handler

### DIFF
--- a/src/amber.cr
+++ b/src/amber.cr
@@ -19,6 +19,8 @@ require "./amber/websockets/**"
 module Amber
   class Server
     include Amber::Configuration
+    getter handler = Pipe::Pipeline.new
+    getter router = Router::Router.new
 
     def initialize
       settings.log.level = ::Logger::INFO
--- a/src/amber.cr
+++ b/src/amber.cr
@@ -18,7 +18,10 @@ require "./amber/websockets/**"
 
 module Amber
   class Server
+    include Amber::DSL::Server
     include Amber::Configuration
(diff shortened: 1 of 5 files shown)


src/amber.cr
```diff
@@ -18,7 +18,13 @@ require "./amber/websockets/**"
 
 module Amber
   class Server
+    include Amber::DSL::Server
     include Amber::Configuration
+    alias WebSocketAdapter = WebSockets::Adapters::RedisAdapter.class | WebSockets::Adapters::MemoryAdapter.class
+    property pubsub_adapter : WebSocketAdapter = WebSockets::Adapters::MemoryAdapter
+    property settings = Settings.new
```
elorest: I'm fine with this being defined here but as long as we use configuration.cr doesn't it make more sense to define this there? In WIP settings branch you can call Settings.fetch(env) to get the env you need. Was thinking it would be easy to have `configuration#settings` return that..

marksiemers: Agreed, it won't cause any issues to leave this line in `configuration.cr` will it? Should be functionally the same.

eliasjpr: I'll move this back to the confirguration.cr. Now bare in mind that this is a  Server property that is used within the server instance.

Issue #388, opened 2017-11-14
Timestamped migrations need to have timestamps with milliseconds


marksiemers:
### Description

This is probably true of all timestamped file names - for consistency.

If you scaffold a model, then generate auth, you get a migration error - due to the difference in timestamps in the filenames.

### Steps to Reproduce

1. `amber new test-app`
2. `amber g scaffold post title body:string`
3. `amber db create migrate`
4. `amber g auth User`  
5. ` amber db migrate`  

**Expected behavior:** [What you expect to happen]

I expect the migration in step 5 to run successfully.

**Actual behavior:** [What actually happens]

Migration reports an error:
```
The following migrations haven't been applied but have a timestamp older then the current version:
    20171113185807_create_user.sql

Micrate will not run these migrations because they may have been written with an older database model in mind.
You should probably check if they need to be updated and rename them so they are considered a newer version.
```
**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions

amber master branch

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.
See the discrepancy in timestamps:
```
20171113181707047_create_post.sql
20171113185807_create_user.sql
```

elorest:
After how much time I wasted on Saturday I can second that milliseconds should probably be used for everything.

eliasjpr:
Milliseconds was added to some not all generated migrations in #381

eliasjpr:
Resolved with #390

PR #390, opened 2017-11-14
Add ms to auth generated migration timestamped file name

marksiemers:
Addresses #388


marksiemers:
I tried writing tests around this, but capturing the stdout from micrate proved difficult.

marksiemers:
@elorest This should go in before the patch release, as I would consider that issue a bug.

PR #393, opened 2017-11-15
Bumped version number to v0.3.5.

elorest:
### Description of the Change

Updated version to v0.3.5

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,5 +1,5 @@
 name: amber
-version: 0.3.0
+version: 0.3.5
 
 authors:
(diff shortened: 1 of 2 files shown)


faustinoaq:
@elorest Do you think we should keep [stable](https://github.com/amberframework/amber/tree/stable) branch or just recommend the use of latest tag?

elorest:
This serves two purposes.
1. It gives us a location to build and push releases from.
2. It gives us a moving stable head to point `install from source` instructions to. Otherwise we have to update the docs everytime we released.

PR #394, opened 2017-11-15
Add app binary to generated .gitignore file

katafrakt:
### Description of the Change

When I was creating my first new application with `amber`, it was natural for me to perform following steps (with a little help of the docs, of course):

```
amber new my-app
cd my-app
shards install
amber w
git init .
git add .
git commit
```

As a result of this operation, a binary file `my-app` was commited. I haven't noticed that in some time, then I was pretty surprised. This is problematic, because:

a) File is more than 4MB large, which is quite a lot for a repository
b) It changes every time I recompile the app, so I would commit more and more versions, increasing repository size.

It would seem natural to have a binary (which name is known) included in autogenerated `.gitignore`.

### Alternate Designs

Maybe having a directory such as `build` where binaries (there might be more than one, actually) would be kept and this whole directory could be gitignored. However if someone has more binaries, he can adjust their `.gitignore` properly.

### Benefits

Less accidental commits including binary file.

### Possible Drawbacks

I don't see any. Perhaps someone does want to keep binaries in version control, but it can be easily achieved by removing appropriate line from autogenerated `.gitignore` file.


faustinoaq:
@drujensen Do you think we can add the ability to compile an amber project using `shards build`? should we use the `bin/` directory instead?

drujensen:
@faustinoaq yeah, we should consider moving the binary to `/bin` or `/build` instead of the root directory.

katafrakt:
@faustinoaq Is there a reason why `amber run` generates binary called `app` and `amber watch` a binary with project name? Maybe it would be better to have one convention.

drujensen:
@katafrakt great question.  It seems like they should both generate the same binary and it would be placed them in the `/bin` directory.

@katafrakt do you feel comfortable making that change?  If not, we can do that in a different PR.

katafrakt:
@drujensen I'll look into it and I will let you know if I feel that it's a bit too much for me. But I think it's not too difficult.

katafrakt:
This is what I came up with.

It would probably be better to extract `Sentry::SentryCommand::Options.get_name` as more general helper method, but to be honest I don't know where to put it.

src/amber/cli/sentry_command_helper.cr
```diff
@@ -15,8 +15,8 @@ module Sentry
         name = Options.get_name
         {
           name:         name,
-          process_name: "./#{name}",
-          build:        "crystal build ./src/#{name}.cr",
+          process_name: "./bin/#{name}",
+          build:        "crystal build ./src/#{name}.cr -o bin/#{name}",
```
faustinoaq: I think we can use `"mkdir -p bin && crystal build ./src/#{name}.cr -o bin/#{name}"` to avoid error if we execute `amber watch` before `amber run`

src/amber/cli/commands/run.cr
```diff
@@ -19,13 +19,15 @@ module Amber::CLI
       end
 
       def run
+        name = Sentry::SentryCommand::Options.get_name
         puts colorize("💎  Crystalizing...", :dark_gray)
-        compile_command = "crystal build $(ls ./src/*.cr | sort -n | head -1) -o app"
+        Dir.mkdir_p("bin")
```
elorest: Check if it exists before creating it?

katafrakt: But `mkdir_p` does essentially the same (checks if exists and creates if not)

faustinoaq: Yes, it does 👍

PR #396, opened 2017-11-16
Added Amber::Pipe::Last to replace current proc.

elorest:
### Description of the Change

Procs end up being slower than method calls.

```sh
call function 238.56M (  4.19ns) (± 8.56%)       fastest
call proc 182.02M (  5.49ns) (± 8.23%)  1.31× slower
```

So I added last Amber::Pipe::Last to replace pipeline proc and made small change to pipeline handler. Worked first time in existing test app.

### DIFF
--- a/src/amber/router/pipe/controller.cr
+++ b/src/amber/router/pipe/controller.cr
@@ -1,8 +1,8 @@
 module Amber
   module Pipe
-    # The session handler provides a cookie based session.  The handler will
-    # encode and decode the cookie and provide the hash in the context that can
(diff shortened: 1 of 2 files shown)


src/amber/router/pipe/controller.cr
```diff
@@ -0,0 +1,12 @@
+module Amber
+  module Pipe
+    # The session handler provides a cookie based session.  The handler will
```
eliasjpr: This comment does not belong here.

PR #397, opened 2017-11-17
[CLI] Removes Amber Console

eliasjpr:
Currently the `amber console` command does not behave as people would
expect. One of the main issues with `amber console` is that already
executed code gets re-executed again and again when a new command is
entered.

- This change is necessary to avoid ambiguity and confusion of the
existence of the two commands `amber console` and `amber exec`
- With the existence of `amber exec` it replaces `amber console` with
better user experience.

- Removes `amber console` in favor of `amber exec`

* This is the most important question to answer, as it can point out problems
where you are making too many changes in one commit or branch. One or two bullet
points for related changes may be okay, but five or six are likely indicators of
a commit that is doing too many things.


elorest:
As soon as the tests pass.

faustinoaq:
@eliasjpr So, We don't need ICR dependency, right? 
https://github.com/amberframework/amber/blob/master/shard.yml#L48 😅

PR #399, opened 2017-11-17
Ep/more db spec

eliasjpr:
## [CLI] Move DB Spec into Correct Spec File

Currently the Database spec is residing inside the generator spec
breaking our current spec design conventions.

This moves the `db` command spec to the correct file providing better
organization for the test suite.

### DIFF
--- a/spec/amber/cli/commands/database_spec.cr
+++ b/spec/amber/cli/commands/database_spec.cr
@@ -9,8 +9,8 @@ module Amber::CLI
   describe "database" do
     describe "sqlite" do
       cleanup
-			scaffold_app(TESTING_APP, "-d", "sqlite")
-			db_filename = db_yml("./")["sqlite"]["database"].to_s.sub("sqlite3:", "")
+      scaffold_app(TESTING_APP, "-d", "sqlite")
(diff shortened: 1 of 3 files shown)


spec/amber/cli/commands/database_spec.cr
```diff
@@ -0,0 +1,33 @@
...
+
+include CLIHelper
+include CLIFixtures
+
+module Amber::CLI
+  describe "database" do
+    describe "sqlite" do
+      cleanup
+			scaffold_app(TESTING_APP, "-d", "sqlite")
+			db_filename = db_yml("./")["sqlite"]["database"].to_s.sub("sqlite3:", "")
```
marksiemers: spaces, not tabs here.

spec/amber/cli/commands/database_spec.cr
```diff
@@ -0,0 +1,33 @@
...
+
+      it "does not create the database when `db create`" do
+        MainCommand.run ["db", "create"]
+        File.exists?(db_filename).should be_false
+      end
```
marksiemers: This is expected? Do weird things happen with sqlite with a blank db?

eliasjpr: Sqlite file gets created with the first migration.

Issue #293, opened 2017-10-07
Move the database url config from the database.yml to the environment.yml


drujensen:
### Description

Currently both Granite and Crecto are using the `config/database.yml` file to configure the database.  They also both support DATABASE_URL environment variable override.

This request is to add an initializer for Granite and Crecto that will look for a `database_url` in the environment instead. 

This will provide the ability to specify different databases for development, test and production.  It should continue to honor the `DATABASE_URL` override environment variable.

elorest:
I'd like to pair with you on this.

eliasjpr:
Can we start work on this? This seems to be fairly straightforward

elorest:
Complete!

PR #402, opened 2017-11-18
[CLI] Moves DB settings to config/environments

eliasjpr:
[CLI] Moves DB settings to config/environments  …
Currently database settings reside in the config/database.yml this is
redundant since Amber has settings per environment.

This PR removes the `config/database.yml` and uses the environment yamls
files to load the database settings.

There is a couple of benefits for doing this:
1. Database settings are loaded per environment.
2. It is more intuitive to know where to put DB settings.

[ENV LOADER] Add database settings to environment_loader

- Adds database property to Amber::Settings
- Adds database property to spec test.yml
- Load database settings from environment yml
- Updates settings specs to check for database settings
- Updates environment loader specs to check for database settings

[CLI] Updates init spec file

- Updates amber spec file to check for database key
- Tidy init_spec connection string validation
- Add `expected_db_url(db_key, env)` method to cleanup specs

Issue https://github.com/amberframework/amber/issues/293

### DIFF
--- a/spec/amber/amber_spec.cr
+++ b/spec/amber/amber_spec.cr
@@ -56,9 +56,9 @@ describe Amber do
         expected_session = {:key => "amber.session", :store => :signed_cookie, :expires => 0}
         settings.session.should eq expected_session
         settings.port.should_not eq 3000
+        settings.database.should eq "mysql://root@localhost:3306/amber_test_app_test"
         expected_secrets = {
           description: "Store your test secrets credentials and settings here.",
-          database:    "mysql://root@localhost:3306/amber_test_app_test",
         }
         settings.secrets.should eq expected_secrets
         settings.secret_key_base.should eq "mV6kTmG3k1yVFh-fPYpugSn0wbZveDvrvfQuv88DPF8"
--- a/spec/amber/cli/commands/database_spec.cr
+++ b/spec/amber/cli/commands/database_spec.cr
@@ -8,25 +8,47 @@ include CLIFixtures
 module Amber::CLI
   describe "database" do
     describe "sqlite" do
+      env = ENV["AMBER_ENV"]
       cleanup
       scaffold_app(TESTING_APP, "-d", "sqlite")
-      db_filename = db_yml("./")["sqlite"]["database"].to_s.sub("sqlite3:", "")
 
-      it "does not create the database when `db create`" do
-        MainCommand.run ["db", "create"]
-        File.exists?(db_filename).should be_false
+      context env do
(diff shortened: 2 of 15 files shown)


spec/amber/amber_spec.cr
```diff
@@ -56,9 +56,9 @@ describe Amber do
         expected_session = {:key => "amber.session", :store => :signed_cookie, :expires => 0}
         settings.session.should eq expected_session
         settings.port.should_not eq 3000
+        settings.database.should eq "mysql://root@localhost:3306/amber_test_app_test"
```
elorest: I liked this being in secrets more because as a framework we're not opinionated on which ORM to use or even if you have to use one at all. 

Someone might choose to use mongo instead of a relational database or choose to have 2 or 3 databases per environment. As such I feel like it makes more sense to leave this in secrets. 

WDYGT? @drujensen @marksiemers @faustinoaq @eliasjpr

drujensen:
@elorest I think the database is embedded in Amber and feel like the `database_url` and `redis_url` should be on par with each other.  All of our scaffolding generators are focused on a full MVC experience.  Just like `redis_url` for websockets, it feels more proper to put the `database_url` as a first class citizen.  People can ignore the setting if they decide to use a different backend.  

Does that make sense?

spec/amber/cli/commands/database_spec.cr
```diff
@@ -8,25 +8,47 @@ include CLIFixtures
...
+
+        it "does not create the database when `db create`" do
+          MainCommand.run ["db", "create"]
+          File.exists?(db_filename).should be_false
+        end
+
+        it "does create the database when `db migrate`" do
+          MainCommand.run ["generate", "model", "Post"]
+          MainCommand.run ["db", "migrate"]
+          p db_filename
```
robacarp: leftover debugging puts?

spec/amber/amber_spec.cr
marksiemers: One way or another, the database url for non-development environments needs to be protected, as it has a username and password in it.

Either encrypting it, or reading from an environment variable.

Is there a certain one that amber has been preferring to date?

eliasjpr:
I have not made any changes to this PR since @drujensen and will pair to merge the 2 branches of work #403

eliasjpr:
@drujensen and @amberframework/core-team please review and approve

faustinoaq:
We will use `database_url`, right?
Still looks good to me, just check the `p db_filename` line, is this needed? https://github.com/amberframework/amber/pull/402#discussion_r152048039

drujensen:
closing this in favor of #406 which changes the name from `database` to `database_url`

PR #404, opened 2017-11-20
Binary signing of encryption.

elorest:
### Description of the Change

1. Creates signature from bytes and appends onto IO.
  * Reduces string operation count to zero in some cases and one in other cases.
  * When saving encryption to files there is no need to cast to string or base64 encode.
  * Signing encrypted file data adds security and prerequisite checks.
  * Makes encrypted cookies harder to gain any useful data from.
  
2. Uses SHA256 instead of compromised SHA1.

### Benefits

1. Roughly 4% faster for encrypted cookies/session. 
2. More secure due to sha256 instead of sha1.

### DIFF
--- a/src/amber/support/message_encryptor.cr
+++ b/src/amber/support/message_encryptor.cr
@@ -15,22 +15,8 @@ module Amber::Support
       @signature_size = 32
     end
 
-    # Encrypt and sign a message. We need to sign the message in order to avoid
-    # padding attacks. Reference: http://www.limited-entropy.com/padding-oracle-attacks.
-    def encrypt_and_sign(value : Slice(UInt8)) : String
-      verifier.generate(encrypt(value))
-    end
-
-    def encrypt_and_sign(value : String) : String
-      encrypt_and_sign(value.to_slice)
-    end
-
     # Verify and Decrypt a message. We need to verify the message in order to
     # avoid padding attacks. Reference: http://www.limited-entropy.com/padding-oracle-attacks.
-    def verify_and_decrypt(value : String) : Bytes
-      verify_and_decrypt(Base64.decode(value))
-    end
-
     def verify_and_decrypt(value : Bytes) : Bytes
       signature = value[value.size - @signature_size, @signature_size]
       data_iv = value[0, value.size - @signature_size]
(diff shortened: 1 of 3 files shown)


src/amber/support/message_encryptor.cr
```diff
@@ -24,22 +28,30 @@ module Amber::Support
     # Verify and Decrypt a message. We need to verify the message in order to
     # avoid padding attacks. Reference: http://www.limited-entropy.com/padding-oracle-attacks.
     def verify_and_decrypt(value : String) : Bytes
-      decrypt(verifier.verify_raw(value))
+      verify_and_decrypt(Base64.decode(value))
```
robacarp: The base64 de-marshaling here doesn't match the marshaling that is happening [when the cookie is written](https://github.com/amberframework/amber/pull/404/files#diff-e135a6eb6d4cfcdee2e8019a8f4a005aR257), does it make sense to have this class handle both Base64 encoding and decoding?

robacarp:
Is there a simple way we can fingerprint the encryption method to make this more upgrade-able in the future?

elorest:
@robacarp Not sure it understand you're meaning on this. This isn't going to change all over the place. It will be a small breaking change with the 0.4.0 release but that's sort of the point of minor and major version releases.

One will just have to decrypt and save their production.yml before upgrading, and reencrypting.

robacarp:
@elorest I think it will also invalidate all cookies signed with the previous algorithm, but I'm not familiar enough with all the code at play to say that with confidence. I'm thinking more in terms of when sha256 is broken enough that it needs to be replaced, is there a way the cookies can be marked as sha256 so that they can be decrypted when this code is updated again.

Its not something necessary to put a ton of time into now, but simply prefixing the signed cookie with `sha256:` after it's been encrypted would give some future developer the opportunity to provide a seamless transition period.

src/amber/support/message_encryptor.cr
robacarp: I'm only trying to point out that by my reading, the code flow for decrypting a message is:

other->MessageEncryptor->base64.decode

and the code flow for encrypting is:

other->base64.encode->MessageEncryptor

I'd rather MessageEncryptor handle both base64 directions or neither.

elorest:
Generally cookies aren't important enough that if they become invalid every year or two anyone worries about it. Rails has changed the way cookies worked many times in the last 12 years.
If they become invalid people will have to login again. Same as if they used a different browser or computer.

src/amber/support/message_encryptor.cr
elorest: Neither makes more sense, since method overloading doesn't work with return values, we'd have to create another method called encrypt_and_base64_encode or something.

I think removing it makes more sense.

robacarp:
@elorest thanks, that's good background and context

PR #405, opened 2017-11-20
Allows amber exec to work in more situations.

elorest:
### Description of the Change

If `amber exec` is called from outside of an amber project it doesn't attempt to load amber config.

This is really useful since crystal doesn't have IRB or PRY. 

I made this change last night while working within amber core to quickly test results.

### Benefits

Amber exec is even more useful.

### Possible Drawbacks

None what-so-ever.

### DIFF
--- a/spec/amber/cli/commands/exec_spec.cr
+++ b/spec/amber/cli/commands/exec_spec.cr
@@ -12,8 +12,8 @@ module Amber::CLI
       `shards`
 
       it "executes one-liners from the first command-line argument" do
-        expected_result = "[:a, :b, :c, :d]\n"
-        MainCommand.run(["exec", "[:a, :b, :c] + [:d]"])
+        expected_result = "3000\n"
+        MainCommand.run(["exec", "Amber::Server.settings.port"])
         logs = `ls tmp/*_console_result.log`.strip.split(/\s/).sort
         File.read(logs.last?.to_s).should eq expected_result
       end
@@ -43,13 +43,18 @@ module Amber::CLI
     end
 
     context "outside of project" do
-      it "complains if not in the root of a project" do
-        expected_result = "Error: 'amber exec' can only be used from the root of a valid amber project"
+      it "executes outside of project but without including project" do
+        expected_result = ":hello\n"
         MainCommand.run(["exec", ":hello"])
         logs = `ls tmp/*_console_result.log`.strip.split(/\s/).sort
         File.read(logs.last?.to_s).should eq expected_result
       end
 
+      it "errors outside of project if referencing amber specific code" do
+        MainCommand.run(["exec", "Amber::Server.settings"])
+        logs = `ls tmp/*_console_result.log`.strip.split(/\s/).sort
+        File.read(logs.last?.to_s).should contain "undefined constant Amber::Server"
+      end
       cleanup
     end
   end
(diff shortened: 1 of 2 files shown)


spec/amber/cli/commands/exec_spec.cr
```diff
@@ -43,13 +43,18 @@ module Amber::CLI
     end
 
     context "outside of project" do
-      it "complains if not in the root of a project" do
-        expected_result = "Error: 'amber exec' can only be used from the root of a valid amber project"
+      it "executes outside of project but without including project" do
```
robacarp: Perhaps this is not easy to do, but how is this test verifying that a project is not getting included?

src/amber/cli/commands/exec.cr
```diff
@@ -39,10 +39,11 @@ module Amber::CLI
         end
 
         result = ""
-        result = `crystal eval 'require "./config/*"; require "#{@filename}"'` if File.exists?(@filename)
-
-        if result.includes?("while requiring \"./config/*\": can't find file './config/*' relative to '.'")
-          result = "Error: 'amber exec' can only be used from the root of a valid amber project"
+        if File.exists?(@filename)
+          eval_cmd = Array(String).new
+          eval_cmd << %(require "./config/*";) if Dir.exists?("config")
```
robacarp: Out of the scope of this pull, but it was not immediately obvious to me that the config directory would be the point of entry for an application because so much require-ing happens in src/application.cr.  Requiring the src/application.cr file isn't appropriate here because starting the server isn't desired, but when the config directory is required, how does that translate to requiring the rest of the application? 

I've run into this a few times when trying to extend some base class for my application and Amber commands start to fail because application.cr is where I loaded my extensions.

elorest: That is a valid point that we should probably discuss again.

spec/amber/cli/commands/exec_spec.cr
elorest: The the next test verifies that part. This test just verifies that it still runs.

src/amber/cli/commands/exec.cr
drujensen: Once we load the `src/application.cr`, the application will start.  We don't want the application to start so we are only loading its dependencies which should all be loaded from the `config/application.cr`

Extensions should be loaded in the `config/initializers/*`

Maybe we can use `require config/application.cr` instead of `./config/*` and then make sure all the of the rest of the stuff is loaded from there.  Does that sound reasonable?

robacarp: My current project is a little out of date with the latest amber release, but there is no reference to initializers in config/application.cr or elsewhere. Perhaps I deleted it! 🎃

drujensen: @robacarp right, this is something fairly new (copied from rails).  We generate the directory in some cases but I think we should consider documenting this and then adding require `./config/initializers/*` in the `config/application.cr` so this becomes the entry point and hopefully help establish `initializers` as the best place to configure your extensions.

robacarp: If we are in process of implementing this, I’d like to toss out there that the initializers folder in big Rails projects frequently becomes a catch all bucket of nonsense for lots of things that aren’t initializer type activities, including but not limited to monkeypatching. I don’t have a recommendation for a better name though

elorest: I think that initializers is fine. People abuse every convention in every language. There are countless things in rails that people are apposed to due to the fact that they worked on a project where someone used way too many callbacks for the wrong things or put the wrong things in initializers. I think the concept of initializers is still valid.

PR #407, opened 2017-11-21
Update auth generator

marksiemers:
### Description of the Change

Updates to the code generated by auth generator. The main motivation for this was that any update to the auth resource (e.g. User) would require a change to the password as well. This is for two reasons:
1. There is a password length validation that happens every time the user is saved
2. The password is stored as a hash, making it impossible to verify the length of the password unless the user provides it.

### Alternate Designs

The validation could be changed to only happen on create, rather than save. That aspect of the design should be considered if a `before_create` validation is available in all ORMs that amber supports using.

### Benefits

A user can change their part of their profile in a generated app, without needing to change their password.

### Possible Drawbacks

This may be scope creep for the auth generator, not sure where to draw the line between including bcrypt and recreating devise (or clearance).


src/amber/cli/templates/auth.cr
```diff
@@ -18,7 +18,7 @@ module Amber::CLI
       @language = language
       @fields = fields.map { |field| Field.new(field, database: @database) }
       @fields << Field.new("email:string", hidden: false, database: @database)
-      @fields << Field.new("encrypted_password:password", hidden: false, database: @database)
+      @fields << Field.new("hashed_password:password", hidden: false, database: @database)
```
marksiemers: This is more semantic, but if we want to keep it consistent with previous versions of amber, we can.

src/amber/cli/templates/auth.cr
```diff
@@ -27,6 +27,9 @@ module Amber::CLI
       @visible_fields = @fields.reject(&.hidden).map(&.name)
 
       add_routes :web, <<-ROUTES
+        get "/profile", #{class_name}Controller, :show
+        get "/profile/edit", #{class_name}Controller, :edit
+        patch "/profile", #{class_name}Controller, :update
```
marksiemers: This is where we might be getting scope creep - having a few profile actions.

src/amber/cli/templates/auth/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -0,0 +1,45 @@
...
+      flash[:danger] = "Could not update Profile!"
+      render("edit.slang")
+    end
+  end
+
+  private def <%= @name %>_params
+    params.validation do
+      required(:email) {|f| !f.nil? && !f.empty? }
+    end
+  end
```
marksiemers: This probably needs to be revisited. I noticed that if I called `user_params.validate!` it gave a runtime exception, so I need to know more about how these should be written and used.

src/amber/cli/templates/auth.cr
drujensen: no, no.  this is great!  I have been wanting to add something like this.  Also, I have been wanting to do `forget password` and generate a mailer.

src/amber/cli/templates/auth.cr
drujensen: yup, that is fine.  love it!

src/amber/cli/templates/auth/src/controllers/{{name}}_controller.cr.ecr
drujensen: I don't know how this works as well.  @eliasjpr ?

src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
```diff
@@ -22,16 +12,34 @@ class <%= class_name %> < Granite::ORM::Base
...
+    <%= @name %>.password_changed? ? <%= @name %>.valid_password_size? : true
+  end
+
+  def password=(password)
+    @new_password = password
+    @hashed_password = Bcrypt::Password.create(password, cost: 10).to_s
+  end
+
+  def password
+    @password ||= (hash = hashed_password) ? Bcrypt::Password.new(hash) : nil
```
drujensen: Will the `||=` cause any issues if they change their password?  Should this reset when they set a new password?

src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
```diff
@@ -22,16 +12,34 @@ class <%= class_name %> < Granite::ORM::Base
...
+  validate :email, "is required", -> (<%= @name %> : <%= class_name %>) do
+    (email = <%= @name %>.email) ? !email.empty? : false
+  end
+
+  validate :password, "is too short", -> (<%= @name %> : <%= class_name %>) do
+    <%= @name %>.password_changed? ? <%= @name %>.valid_password_size? : true
+  end
+
+  def password=(password)
+    @new_password = password
```
drujensen: should this just be a flag instead?  `@new_password = true`?

src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
```diff
@@ -22,16 +12,34 @@ class <%= class_name %> < Granite::ORM::Base
...
+  def password
+    @password ||= (hash = hashed_password) ? Bcrypt::Password.new(hash) : nil
+  end
+
+  def password_changed?
+    new_password ? true : false
+  end
+
+  def valid_password_size?
+    (pass = new_password) ? pass.size >= 8 : false
```
drujensen: nm, just saw this being used here.  right, that makes sense.

src/amber/cli/templates/auth.cr
marksiemers: Is the existing mailer example a good starting point? That should probably be another PR (forgot password functionality).

I'm assuming `SecureRandom.urlsafe_base64(16)` will be sufficient for generating a token.

drujensen: yeah, the `forget password` would be a different PR.

src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
marksiemers: That's a good point.

Normally with the flow through the program, `password=` is called before `password` but there isn't anything stopping it from happening the other way around.

I don't think you have the blowfish overhead with `Bcrypt::Password.new(hash)` only on `Bcrypt::Password.create` and `Bcrypt::Password#==`

I was thinking of ways to still memoize `@password` but I'm not sure it is even worth it.

src/amber/cli/templates/auth/src/controllers/{{name}}_controller.cr.ecr
marksiemers: It looks like calling `params.valid?` in combination with `params.to_h` may be the way to go. @eliasjpr can you comment?

eliasjpr: In a nutshell this is how it works.

```crystal
# This defines a validation to perform
params.validation do
  required(:email, "Custom error message here") {|f| !f.nil? && !f.empty? }
end

# Returns boolean
params.valid?     

# Success: returns a hash with only the fields defined in the validation
# Lets assume token, password and email exists in the request params 
# but only email is returned since password and token are not defined in the validation
params.validate! # => { :email => "some@email.com" } 

# Failure: returns exception and `params.errors` will have the failed fields
params.validate! # => `params.errors => [ { :email => "Custom error message here"}]`
```

marksiemers:
@drujensen and @eliasjpr - I think a bit more thought needs to go into the flows here before finalizing this.

What features should be included with this PR?
- Allowing updating of a profile?
- Requiring input of password to update a profile?
- Doing a `sudo` style requiring of password (similar to github) where your password input lasts for a little while (I'm not sure how this is done, perhaps a timestamp in a cookie)

Also, see https://github.com/marksiemers/auth-test/pull/1 for some thoughts on using before_action, and generally DRYing up that controller.

If it isn't obvious, I try to follow Sandi Metz's 5 rules, in particular, [this one](https://robots.thoughtbot.com/sandi-metz-rules-for-developers#five-lines-per-method)

drujensen:
@marksiemers yes, its obvious.  ;-)

elorest:
I'm liking this!

src/amber/cli/templates/auth/src/controllers/{{name}}_controller.cr.ecr
marksiemers: @eliasjpr - Can you take a look at this file: https://github.com/marksiemers/auth-test/blob/7cc94319054a7a1971c4d269cbf6bacf1cc6759c/src/controllers/user_controller.cr

Overall, does this look good? The use of `before_all`, the way params validations work, the overall design of the controller and actions.

src/amber/cli/templates/auth/src/views/layouts/+_nav.slang.ecr
```diff
@@ -1,8 +1,9 @@
...
-  a class="nav-item pull-right" href="/signout" Sign Out
-  div.nav-item.pull-right == context.current_<%= @name %>.not_nil!.email
+- if (current_<%= @name %> = context.current_<%= @name %>)
+  a.nav-item.pull-right[href="/signout"] Sign Out
+  - active = context.request.path == "/profile" ? "active" : ""
+  a[class="nav-item #{active} pull-right" href="/profile"] == current_<%= @name %>.email
 - else
   - active = context.request.path == "/signup" ? "active" : ""
-  a class="nav-item #{active} pull-right" href="/signup" Sign Up
+  a[class="nav-item #{active} pull-right" href="/signup"] Sign Up
```
marksiemers: Any thoughts on this syntax for `slang` I like it because it more explicitly separates attributes from content, but I could see the argument that it doesn't look as clean with the extra characters.

src/amber/cli/helpers/migration.cr
```diff
@@ -0,0 +1,63 @@
...
+      #{create_table_fields_sql}
+    );
+    SQL
+  end
+
+  def database
+    if File.exists?(AMBER_YML) &&
+       (yaml = YAML.parse(File.read AMBER_YML)) &&
+       (database = yaml["database"]?)
+      database.to_s
```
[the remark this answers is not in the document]
drujensen: no, this is the `.amber.yml` and this is specifying the type of database.  `pg`, `mysql` or `sqlite`.  Should we rename this to `database_type`?

src/amber/cli/templates/auth/src/views/layouts/+_nav.slang.ecr
drujensen: because of the dynamic class of `active`, we cannot not use the more explicit class separator.

marksiemers: Let me clarify

This (with square brackets):
```
a[class="nav-item #{active} pull-right" href="/signup"] Sign Up
```
vs this (without):
```
a class="nav-item #{active} pull-right" href="/signup" Sign Up
```

drujensen: ahhh, gotcha.  yes, go for it and remove them.

marksiemers: I like the square brackets, maybe we put it to a vote.
👍  - for square brackets, 👎  - for white space

src/amber/cli/templates/auth/src/controllers/{{name}}_controller.cr.ecr
eliasjpr: After tracing the `user` argument it seems that it would always be present by the time is performing the update. 

This how I might approach it personally. Maybe you can get an idea or two from it.
```crystal
class UserController < ApplicationController
  # Since the context already exists you can assign the context to the user in the property
  property current_user : User? = context.current_user

  before_action do
    all { user_signed_in? }
  end

  def show
    render("show.slang") if (user = @current_user)
  end

  def edit
    render("edit.slang") if (user = @current_user)
  end

  def update
    if update_user?
      flash[:success] = "Updated Profile successfully."
      redirect_to "/profile"
    else
      flash[:danger] = "Could not update Profile!"
      render("edit.slang")
    end
  end

  private def update_user(user)
    current_user.set_attributes(user_params.validate!)
    # Granite::ORM It would be nice to be able to do something like current_user.save!  where `save!` runs the validations.
    current_user.valid? && current_user.save
  end

  private def user_params
    params.validation do
      required(:email) {|f| !f.nil? && !f.empty? }
      optional(:password) {|f| !f.nil? && !f.empty? }
    end
  end
  
  # Previously this method was had two responsability assign the current user and redirect
  # Now it just checks if the current user is signed in.
  private def user_signed_in?
    unless current_user
      flash[:info] = "Must be logged in"
      redirect_to "/signin"
    end
  end
end
```

marksiemers: > After tracing the user argument it seems that it would always be present by the time is performing the update.

This is true, I don't know if the compiler can track it though. I can put this syntax in and see what happens. 

The same is true for `edit` and `show` (@current_user will always be something), but with the current setup, the compiler can't track that.

eliasjpr: It also seems that people could benefit from having a user_signed_in? and current_user in application controller. Since people would exercise these methods is other places as well.

eliasjpr: This two methods should definitely be added to application_controller.cr
```crystal
def current_user
  @current_user ||= context.current_user
end

private def redirect_signed_out_user 
    unless current_user
      flash[:info] = "Must be logged in"
      redirect_to "/signin"
    end
end
```

src/amber/cli/templates/auth/src/views/layouts/+_nav.slang.ecr
marksiemers: Fine, we'll treat attributes and content the same, like barbarians 😉 
```slim
/ So clean and CSS-like
a[href="/signup"] Sign Up

/ But one less character tho
a href="/signup" Sign Up
```

faustinoaq: Well, we can use indentation too:

```slim
a href="/signup"
  | Sign Up
```

marksiemers:
I ran into some crystal compiler issues trying to get `current_user`, etc. into the `ApplicationController`

It appears to have to do with downcasting, which is a known bug: https://github.com/crystal-lang/crystal/issues/5190

marksiemers:
@amberframework/core-team - Many changes since the last approvals. LMKWYT.

elorest:
I could help with this if you're still working on it?

> I ran into some crystal compiler issues trying to get current_user, etc. into the ApplicationController

marksiemers:
@elorest - Thank you. I did get it to compile. I think the auth generator will continue to be a WIP beyond this PR. Hopefully we can get some feedback from the community if this is put into a release.

There is a question of scope - somewhere between what is here and the full functionality of devise.

Unless there are any outstanding bugs or coding style issues to be resolved, this can be merged. It is working.

See this comment https://github.com/crystal-lang/crystal/issues/5190#issuecomment-348117600 if you're curious about the line that seems to be causing the compiler issue.

drujensen:
@marksiemers I don't see any outstanding concerns.  I am going to merge this.

PR #410, opened 2017-11-22
[CLI] Updates Router Template

eliasjpr:
Updates config/router.cr to use Amber::Pipe:Static instead of Crystals
Static handler. Users are receiving the incorrect content type for
static files served. 

Using the Amber::Pipe::Static results in returning the correct content type for documents served by Amber Server.

### Benefits

It serves static documents and sets the correct content type for the response header.

### DIFF
--- a/spec/amber/router/pipe/static_spec.cr
+++ b/spec/amber/router/pipe/static_spec.cr
@@ -5,7 +5,7 @@ include RouterHelper
 
 module Amber
   module Pipe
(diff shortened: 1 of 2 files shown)


elorest:
If you've tested this and know it works it makes sense to change it, since it apparently handles content type better than the default one.

eliasjpr:
I added a test to verify this is true

robacarp:
I happened on this bug just yesterday, good timing!

Issue #346, opened 2017-10-31
Improve amber logs


faustinoaq:
### Description

Currently Amber shows request and some response information on logs.

```
 | DELETE  "/languages/1"  | 14.55ms
Params: _method=delete&id=1
200  | GET  "/languages"  | 1.96ms
Params: 
 | DELETE  "/languages/2"  | 27.49ms
Params: _method=delete&id=2
200  | GET  "/languages"  | 1.66ms
Params: 
 | DELETE  "/languages/3"  | 24.58ms
Params: _method=delete&id=3
200  | GET  "/languages"  | 2.28ms
Params: 
 | DELETE  "/languages/4"  | 24.58ms
Params: _method=delete&id=4
200  | GET  "/languages"  | 1.68ms
Params: 
 | DELETE  "/languages/5"  | 27.02ms
Params: _method=delete&id=5
 | DELETE  "/languages/5"  | 397.0µs
Params: _method=delete&id=5
200  | GET  "/languages"  | 1.45ms
```

Would be useful to show query data on logs too, just like Rails and Phoenix,

Also adding time and aligned/tabulated logs would be useful and more readable.

### Steps to Reproduce

1. create a new app
2. run the app
3. see the logs

**Expected behavior:** aligned and well spaced logs, also timestamp 

**Actual behavior:** logs spreaded in terminal, using multiple lines and empty `Params: ` without timestamps.

**Reproduces how often:** 100%

### Versions

100%

### Additional Information

We can follow Lucky/Kemal logs, or even better logs, using Rails/Phoenix pattern

I personally liked the Lucky logs 👇 **However I think we can improve them**, Lucky logs still lacks good tabulation and spacing.

![screenshot_20171031_143928](https://user-images.githubusercontent.com/3067335/32252202-681590d8-be62-11e7-9997-6d22450c0cac.png)

![screenshot_20171031_145419](https://user-images.githubusercontent.com/3067335/32252203-682da8ee-be62-11e7-9946-8b247badc6ed.png)

@amberframework/contributors WDYT?

Reference: https://github.com/amberframework/granite-orm/issues/86

marksiemers:
@faustinoaq 
Improving the logs is a good idea. I think the lucky logs look a bit noisy (perhaps the lack of formatting / indentation).

I'm used to Rails logs, so that would be my bias. Should we start with a list of priorities that we want to see in the log, then think about the design of the output?

For a start:
- route information (verb, controller, action, pipeline, path)
- params
- http status code
[...]

PR #412, opened 2017-11-23
[ENV] Allows for environment settings to be loaded dynamically via a YAML settings file

eliasjpr:
### Description of the Change

Allows for environment settings to be loaded dynamically via a YAML settings file.

- Default environment path: `{project root}/config/environments`

### Usage
```crystal
Amber.env = :production
Amber.env = :test
Amber.env = :development

# Changing environment path
Amber.environment_path = "Some/Other/Path"

# Gets current environment settings
Amber.settings

# Get logger
Amber.logger

```

### Benefits

- Adds documentation to `config/application.cr` for each setting
- Conveniently easy to switch environments
- Easy to add more settings variables
- Easy to test
- All server settings are referenced via `Amber.settings` **Updated**
- All Amber logging is now done thru `Amber.logger` **Updated**
- Parallelizes Travis tests by component
- Command Exec uses `config/application.cr` since this is the main entry-point file 
- Adds helper methods `Amber.secret_key`

### Possible Drawbacks
- Cannot embed Dynamic code in yaml settings file.
- Must use `config/application.cr` for any dynamic settings including loading environment variables.


src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -1,16 +1,115 @@
...
 
-# NOTE: Settings should all be in config/environments/env.yml.
-# Anything here will overwrite all environments.
-Amber::Server.configure do |setting|
-  # Server options
-  # setting.name = "<%= display_name %> web application."
-  # setting.port = 80 # Port you wish your app to run
-  # setting.log = ::Logger.new(STDOUT)
-  # setting.log.level = ::Logger::INFO
+# About Application.cr File
```
drujensen: 🥇 excellent documentation!

src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -1,16 +1,116 @@
...
+# Amber::Server boot up.
+#
+# > We recommend to not modify the order of the require since the order will
+# affect the behavior of the application.
+#
+# With `Amber::Server.configure` block you can redefine the Server configuration
+# settings and use ENVIRONMENT variables and/or values evaluated at runtime.
+#
+#
+Amber::Server.configure do |settings|
```
elorest: I'm concerned that this will encourage them to configure their environments here instead of using `config/environments/env.yml`. 

Perhaps we should only show a couple examples that relate to ENV vars and show how to optionally override.

Most settings should be set in the env files. Not here!

src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -1,16 +1,116 @@
...
+  # .amber_secret_key in this order.
+  #
+  # settings.secret_key_base= ENV["SECRET_KEY_BASE"]
+  #
+  #
+  # Host: is the application server host address or ip address. Useful for when 
+  # deploying Amber to a PAAS and likely the assigned server IP is either 
+  # known or unknown. Defaults to an environment variable HOST
+  #
+  # settings.host = ENV["HOST"]
```
elorest: ENV overrides should probably be setup to over ride if ENV exists but otherwise use the default settings from env.ymls.

EXAMPLE: 

```cr
settings.host = ENV["HOST"] if ENV["HOST"]?
```

This way if HOST is empty it will keep the existing value instead of setting it to nil.

src/amber/cli/templates/app/config/environments/development.yml.ecr
```diff
@@ -18,9 +16,11 @@ database_url: postgres://postgres:@localhost:5432/<%= database_name_base %>_deve
...
-  store: :signed_cookie
-  expires: 0
-secrets:
-  description: Store your development secrets credentials and settings here.
+session: {
+  key: amber.session,
+  store: signed_cookie,
+  expires: 0,
+}
+secrets: {
```
elorest: Why does this have curly brackets? Shouldn't it just be a tab to turn into has with yaml?

src/amber/cli/templates/app/config/environments/production.yml.ecr
```diff
@@ -1,14 +1,12 @@
 secret_key_base: <%= SecureRandom.urlsafe_base64(32) %>
-port: <%= (ENV["PORT"] ||= "80").to_i %>
+port: 8080
 name: <%= @name %>
-log: "::Logger.new(STDOUT)"
```
elorest: How do we configure loggers per environment this way?

src/amber/scripts/environment_loader.cr
```diff
@@ -1,75 +0,0 @@
...
-        "env: #{environment}"
-      end
-
-settings = YAML.parse(yml)
-
-str = String.build do |s|
-  # Most of this logic can be cleaned up by just requiring environment files to contain valid params.
-  # For now I have this here so that tests can still pass without having to load env files although they should.
-  # This is a transistion.
-  s.puts %(@name = "#{settings["name"]? || "Amber_App"}")
```
elorest: So much red. I agree this is for the best but it makes me sad.

src/amber/server/server.cr
```diff
@@ -0,0 +1,102 @@
...
+    def self.start
+      instance.run
+    end
+
+    # Configure should probably be deprecated in favor of settings.
+    def self.configure
+      with self yield settings
+    end
+
+    def self.settings
```
elorest: So this is all back here now? I'm fine with it. Why did you move it out to `configuration.cr` just to move it back here though?

src/amber/server/settings.cr
```diff
@@ -1,26 +1,63 @@
...
-    property host : String
-    property ssl_key_file : String? = nil
-    property ssl_cert_file : String? = nil
-    property redis_url = ""
+    property log : ::Logger = Logger.new(STDOUT)
+
+    property colorize_logging : Bool = true
     property database_url = ""
-    property session : Hash(Symbol, Symbol | String | Int32)
+    property host : String = "localhost"
```
elorest: We shouldn't be defining defaults in both property definitions and yaml mappings. I'd vote for putting all defaults in yaml mappings.

src/amber/server/settings.cr
```diff
@@ -1,26 +1,63 @@
...
-    property color : Bool = true
-    property process_count : Int32
-    property secret_key_base : String
-    property port : Int32
-    property name : String
-    property host : String
-    property ssl_key_file : String? = nil
-    property ssl_cert_file : String? = nil
-    property redis_url = ""
+    property log : ::Logger = Logger.new(STDOUT)
```
elorest: I don't think we should have properties of settings that can't be set through environment.yms's.

src/amber/websockets/adapters/redis.cr
```diff
@@ -10,8 +10,8 @@ module Amber::WebSockets::Adapters
 
     # Establish subscribe and publish connections to Redis
     def initialize
-      @subscriber = Redis.new(url: Amber::Server.redis_url)
-      @publisher = Redis.new(url: Amber::Server.redis_url)
+      @subscriber = Redis.new(url: Amber::Server.settings.redis_url)
```
elorest: Weren't you planning on moving all of these straight to Amber like `Amber.settings.redis_url`.

src/amber/cli/templates/app/config/environments/production.yml.ecr
eliasjpr: Honestly defining a Logger should not be part of a setting. I am considering to move the logger to `Amber.logger`

src/amber/server/server.cr
eliasjpr: Removing this in favor of `Amber.settings`

src/amber/server/settings.cr
eliasjpr: I agree with you. But for now I am going to leave as is for now until we can come up with a better soloution. WDYT?

src/amber/server/settings.cr
eliasjpr: Completely agree.

src/amber/websockets/adapters/redis.cr
eliasjpr: Just did 😃

src/amber/cli/templates/app/config/application.cr.ecr
eliasjpr: I will add a better note that the YAMLs are first class citizen and only use this if completely necessary.

src/amber/scripts/environment_loader.cr
eliasjpr: Don't be, most of the inspiration came from your version 😄

src/amber/cli/templates/app/config/environments/development.yml.ecr
elorest: In my tests with yaml it seems to work just fine without them...

```yaml
secrets:
  name: "isaac"
  age: 34
  hash: true
```

src/amber/cli/templates/app/config/environments/production.yml.ecr
elorest: Would it be possible to completely configure the logger through the pipeline rather than here?

`plug Amber::Pipe::FastLogger.new(stdout: Amber.env.development?)`

elorest:
Thanks

eliasjpr:
@amberframework/core-team made some updates based on @elorest suggestions. The updates includes

- Amber.logger  - Gets the logger from the settings
- Amber.settings - Gets `current` environment settings
- Removes curly braces from YAML - it turns out that it just needed a break-line between the session and secrets 
- Travis.yml - Reduce the number of parallel test run and moves the build_spec test to the root of the root. This allows to run the latter in isolation.

What I did not do:

**Application.cr** @elorest brought up a good point around the environment variables in the application.cr file. He suggested to use `settings.host = ENV["HOST"] if ENV["HOST"]?`

> I decided to not add the if statement because the settings in this file are by default commented out and it explicitly states that if used it will override the settings value from the YAML for the environment loaded. It is up to the user to use define and use the environment variables accordingly.

faustinoaq:
Is this PR the first step to achieve #346?

elorest:
@eliasjpr 

Thanks for working on this over Thanksgiving.

I'd like to point out that my suggestion for environment variables is bigger than just functionality. If someone were to uncomment host and port. But ever not set one environment variable it will exception on index out of bounds and crash. This behavior will be unexpected since we provided them this code to uncomment. 

IMO the expected behavior is "Use overwrite default settings if ENV VAR exists", not "use ENV VAR or Crash if it doesn't exist."

```cr
settings.host = ENV["HOST"]
settings.port = ENV["PORT"]
```
We should provide a few example of using ENV VAR only if it exists. Otherwise we're going to have to explain this a lot of times in gitter.

WDYT?

eliasjpr:
@elorest IMHO think is a good thing that it would raise an exception if the **ENV** variable was never defined at least I know as a developer I forgot to set something up. Not sure if this is a blocking issue for this PR. I would like for use to raise a discussion around that in the core team channel and get feedback from everyone, and address the outcome in a separate PR.

elorest:
Ok. To me it's not that it raises an exception that has me concerned, as it will work in development if they define an ENV VAR. As soon as it's not set though on a different server it will error. This could result in an error that's hard to track down. 

Obviously it's fine if the users write code like this. I'm just concerned that we seem to be encouraging it. 

Within Rails and Phoenix every ENV override has default values that are used instead if they're not set.

If a PORT isn't set it goes to 3000, if an RAILS_ENV isn't set it goes to development, if DATABASE_URL isn't set it uses settings from database.yml.

I don't think that it's clear to users that uncommenting the code in `application.cr` will work differently.

PR #415, opened 2017-11-24
Add error generator to CLI

marksiemers:
### Description of the Change

Add an error generator to the CLI

### Alternate Designs

None considered.

### Benefits

Allows generating an error controller and views from the CLI for an amber app.

### Possible Drawbacks

Encourages creating a custom error controller instead of letting the framework handle them.


spec/amber/cli/build_spec.cr
```diff
@@ -14,6 +14,7 @@ module Amber::CLI
       options = ["user:reference", "name:string", "body:text", "age:integer", "published:bool"]
       temp_options = options - ["user:reference", "age:integer"]
       MainCommand.run ["generate", "auth", "User"] | (options - ["user:reference"])
+      MainCommand.run ["generate", "error", "error"]
```
marksiemers: I think, as is, you will have to give an argument (a name) which the generator will require, but also ignore.

All the other generators need a name. So, so we want to allow custom names, or figure out how to over-ride the requirement for this one?

marksiemers:
@faustinoaq - it is possible to do that now, it is just a question of how many do we want automatically included in the generator.

The 3 we have are pretty common, and shows a developer the structure of creating the controller action and view.

There are nearly 40 4xx and 5xx status codes I think.

marksiemers:
Hmm, should I add to the build tests - making sure that there are controller specs that check that these return the correct status codes?
@eliasjpr @elorest

spec/amber/cli/build_spec.cr
drujensen: yeah, I think we need to overwrite the requirement for this one.  not sure how hard that is.

elorest: Why is it `generate error error` instead of like `generate error 404` or something? error error seems strange semantically.

marksiemers:
@amberframework/core-team - Significant changes since the last review. Please take a look.
- The code for `Amber::Pipe::Error` gets moved into the app if the user does `amber g error`
- Reorganized the directory structure for the template (to match `auth` rather than `scaffold`) - it is simplified
- Re-wrote the generated spec file (includes a controller unit test and integration test)
- `amber g error` no longer requires a name, while the other `amber g`commands do - not sure if I did this in the best way

spec/build_spec.cr
```diff
@@ -65,7 +66,7 @@ module Amber::CLI
       end
     end
   ensure
-    cleanup
+    # cleanup
```
eliasjpr: Is this intentional or just for debugging?

src/amber/cli/templates/error/spec/controllers/error_controller_spec.cr.ecr
```diff
@@ -0,0 +1,57 @@
+require "./spec_helper"
+
+# Controller Unit Test
```
marksiemers: @eliasjpr - What do you think about the unit test in the code below?

With the other controller tests, those are testing the class directly. I think they are more helpful, but strictly speaking, they may fall under "feature" tests or some other category.

I've included both here - since we haven't really fleshed out how to test controllers and organize those tests.

marksiemers:
@drujensen - When you get a chance, take a look.

PR #416, opened 2017-11-25
RFC: Proposing a new url router for amber.

robacarp:
At its current state, this is an MVP spike. The code is as yet incomplete, but before filling out the edge cases and ancillary features I want to get some feedback from @amberframework/core. It may not be worth putting the extra time into it if there's not enough admin buy in to use the thing :] (I had fun building it either way).

### Description of the Change

This replaces the router code which decides which decides which `Amber::Route` should be executed for a given url. 

### Alternate Designs

The current shard, https://github.com/luislavena/radix . I decided to prototype a router for Amber after running into two weaknesses in the design of this router in a single day of work. I think the efficiency of radix can be improved upon as well.

Radix doesn't support similar paths with different parameter names:

`/a/b/:foo/c` and `/a/b/:bar/z` cannot coexist because radix enforces that the variable names match. This poses problems with routes like this:

```
GET /widgets/:id => WidgetsController
GET /widgets/:widget_id/components => ComponentsController
```

Radix doesn't deal well with route trees which produce false matches. Given the current generated Amber project produces a `public/dist` folder and wires up StaticController to render files in there, Routes sometimes cannot begin with a d__. Eg: `GET /domain/` prevents static files from being rendered. (sometimes: it depends on how radix builds out the character tree).

Documented on radix issues, globs are always greedy. There is no way for `/foo/*/bar` and `/foo/*/baz` routes to match as expected because `*` greedily gobbles up the rest of the route.

Perhaps most importantly, I think a library as important as this should be community maintained. I believe that either the Kemal project or Amber should own a dedicated repository and shard for this code. It can be shared from there if projects desire.

### Benefits

This prototype router:
- Navigates urls based off of strings, not characters. I haven't run a benchmark but I believe this would have performance impacts and I'm not aware of any drawbacks to this.
- Allows for both greedy globs: `GET /*` and restrained globs: `GET /foo/*/bar` by switching to an end-first search when a glob segment is matched.
- Allows for diverse variable names in similar positions: `GET /widgets/:id` and `GET /widgets/:widget_id/components`
- Most importantly, doesn't get hung up on trees which produce false traversals as in the example above.

### Possible Drawbacks

- It's a big change, and this is the type of code that will run a lot. Performance needs to be considered, as does feature scope.
- Community pushback from the broader crystal community? I considered submitting a pull to radix to address the bug I found, but the bug may be a modeling problem in the algorithm. I am new enough to the crystal community that I don't have a good pulse on this.

### TODO list:

Assuming the amber project wants to move forward with this change, this is a list of things still needed to move forward:

- [ ] write a bunch of test cases
  - [ ] short globs: `GET /*`
  - [ ] long globs: `GET /foo/*/bar`
  - [ ] etc. (feel free to add to this list)
- [ ] benchmarks
- [ ] verify a top-down order to routes?
- [ ] move to a separate repository and make it generically typed
- [ ] handle returning route parameters (currently not handled at all)
- [ ] Get `params` for the matched route. Something like `matched_route.params` (Updated by @eliasjpr)
- [ ] decide what to do at edge cases:
  - [ ] when more than one route matches a url
  - [ ] when routes are declared that are functionally identical
  - [ ] _perhaps others_
- [ ] _additions welcome_

### DIFF
--- a/src/amber/router/route_set.cr
+++ b/src/amber/router/route_set.cr
@@ -1,0 +1,285 @@
+module Amber
+  module Router
+    class RouteSet
+      @trunk : RouteSet?
+      @route : Route?
+
+      property segment : String
+      property segment_type = 0
+      property full_path : String?
+
+      ROOT = 1
+      FIXED = 2
+      VARIABLE = 4
+      GLOB = 8
+
+      # A tree data structure (recursive). The initial construction has an segment
+      # of "root" and no trunk. Subtrees must pass in these details upon creation.
+      def initialize(@segment = "#", @trunk = nil)
+        @branches = Array(RouteSet).new
+
+        if @trunk
+          @segment_type = FIXED
+
+          if @segment.starts_with? ':'
+            @segment_type = VARIABLE
+          end
+
+          if @segment.starts_with? '*'
+            @segment_type = GLOB
+          end
+
+        else
+          @segment_type = ROOT
+        end
+      end
+
+      def deep_clone : RouteSet
+        clone = {{@type}}.allocate
+        clone.initialize_copy(self)
+        clone
+      end
+
+      protected def initialize_copy(other) : Nil
+        @route = other.@route
+        @trunk = nil
+        @segment = other.@segment
+
+        @segment_type = other.@segment_type
+        @full_path = other.@full_path
+
+        @branches = other.@branches.map { |s| s.deep_clone.as(RouteSet) }
+      end
+
+      # Look for or create a subtree matching a given segment.
+      def find_subtree!(segment : String) : RouteSet
+        if subtree = find_subtree segment
+          subtree
+        else
+          RouteSet.new(segment, self).tap do |subtree|
+            @branches.push subtree
+          end
+        end
+      end
+
+      # Look for and return a subtree matching a given segment.
+      def find_subtree(segment : String) : RouteSet?
+        @branches.each do |subtree|
+          break subtree if subtree.segment_match? segment
+        end
+      end
+
+      def segment_match?(segment : String) : Bool
+        segment == @segment
+      end
+
+      def root? : Bool
+        @segment_type == ROOT
+      end
+
+      def fixed? : Bool
+        @segment_type == FIXED
+      end
+
+      def variable? : Bool
+        @segment_type == VARIABLE
+      end
+
+      def glob? : Bool
+        @segment_type == GLOB
+      end
+
+      def leaf? : Bool
+        @branches.size == 0
+      end
+
+      def routes? : Bool
+        @branches.any?
+      end
+
+      # Recursively count the number of discrete paths remaining in the tree.
+      def size
+        return 1 if leaf?
+
+        @branches.reduce 0 do |count, branch|
+          count += branch.size
+        end
+      end
+
+      # Recursively descend to find the attached application route.
+      # Weakness: assumes only one path remains in the tree.
+      def route
+        return @route if leaf?
+        @branches.first.route
+      end
+
+      # Recursively _prunes_ the route tree by matching segments
+      # against path segment strings.
+      #
+      # A destructive breadth first search.
+      #
+      # return true if any routes matched.
+      def select_routes!(path : String) : Bool
+        first_segment, remaining_path = split_path path
+
+        reverse_match = false
+
+        case
+        when root?
+          # select all branches that match the full path
+          remaining_path = path
+        when fixed?
+          # select branches only if this segment matches
+          segment_match? first_segment
+        when variable?
+          # always match
+        when glob?
+          reverse_match = true
+        end
+
+        if reverse_match
+          match, _ = reverse_select_routes! path
+          return match
+        else
+          @branches.select! do |subtree|
+            subtree.select_routes! remaining_path
+          end
+        end
+
+        @branches.any? || (leaf? && remaining_path == "")
+      end
+
+      # Recursively matches the right hand side of a glob segment.
+      # Allows for routes like /a/b/*/d/e and /a/b/*/f/g to coexist.
+      # This is a modified version of a destructive depth first search.
+      #
+      # Importantly, each subtree must pass back up the remaining part
+      # of the path so it can be matched against the parent, so this
+      # method somewhat awkwardly returns:
+      #
+      #   Tuple(subtree_match : Bool, path_for_trunk_to_match : String)
+      #
+      def reverse_select_routes!(path : String) : Tuple(Bool, String)
+        remnant_path = ""
+        was_leaf = leaf?
+
+        @branches.select! do |subtree|
+          match, remaining_path = subtree.reverse_select_routes! path
+          if match
+
+            if remnant_path != ""
+              raise "warning: overwriting remnant path"
+            end
+
+            remnant_path = remaining_path
+          end
+        end
+
+        # If this segment started as a leaf, no remant path exists. Match against the whole path.
+        remnant_path = path if was_leaf
+
+        # If this wasn't a leaf and there are no branches left, it's not a match.
+        return Tuple(Bool, String).new(false, "") unless @branches.any? || was_leaf
+
+        # If this node is the glob, at least one subtree matched (or there are none).
+        return Tuple(Bool, String).new(true, "") if glob?
+
+        remaining_path, last_segment = reverse_split_path remnant_path
+
+        matched = case
+        when fixed?
+          segment_match? last_segment
+        when variable?
+          true
+        else
+          false
+        end
+
+        Tuple(Bool, String).new(matched, remaining_path)
+      end
+
+      # Add a route to the tree.
+      def add(path, route : Route) : Nil
+        add(path, route, path)
+      end
+
+      # Recursively find or create subtrees matching a given path, and store the
+      # application route at the leaf.
+      protected def add(path : String, route : Route, full_path : String) : Nil
+        if path == ""
+          if @route.nil?
+            @route = route
+            @full_path = full_path
+            return
+          else
+            raise "Unable to store route: #{full_path}, route is already defined as #{@full_path}"
+          end
+        end
+
+        first_segment, remaining_path = split_path path
+
+        subtree = find_subtree! first_segment
+        subtree.add(remaining_path, route, full_path)
+      end
+
+      # Find a route which has been assigned to a matching path
+      # Weakness: assumes only one route will match the path query.
+      def find(path) : Amber::Route?
+        matches = deep_clone
+        matches.select_routes!(path)
+
+        if matches.size > 1
+          raise "Warning: matched multiple routes"
+        end
+
+        matches.route
+      end
+
+      # Produces a readable indented rendering of the tree, though
+      # not really compatible with the other components of a deep object inspection
+      def inspect(*, ts = 0)
+        tab = "  " * ts
+        @branches.reduce("#{@segment} :\n") do |s, subtree|
+          if subtree.routes?
+            s += "#{tab}- #{subtree.inspect(ts: ts + 1)}"
+          else
+            s += "#{tab}- #{subtree.segment} (#{subtree.full_path})\n"
+          end
+
+          s
+        end
+      end
+
+      # Split a path by slashes, and return the first segment and the rest untouched.
+      #
+      # E.g. split_path("/a/b/c/d") => "a", "b/c/d"
+      private def split_path(path : String) : Tuple(String, String)
+        first_segment = path.split("/").first
+
+        if path.size >= first_segment.size + 1
+          remaining_path = path[first_segment.size + 1..-1]
+        else
+          remaining_path = ""
+        end
+
+        Tuple(String, String).new(first_segment, remaining_path)
+      end
+
+      # Split a path by slashes, and return the last segment and the rest untouched.
+      #
+      # E.g. split_path("/a/b/c/d") => "a/b/c", "d"
+      private def reverse_split_path(path : String) : Tuple(String, String)
+        last_segment = path.split("/").last
+
+        if path.size >= last_segment.size + 1
+          remaining_path = path[0...last_segment.size * -1 - 1]
+        else
+          remaining_path = ""
+        end
+
+        Tuple(String, String).new(remaining_path, last_segment)
+      end
+
+    end
+  end
+end
--- a/src/amber/router/router.cr
+++ b/src/amber/router/router.cr
@@ -9,7 +9,7 @@ module Amber
       PATH_EXT_REGEX = /\.[^$\/]+$/
 
       def initialize
-        @routes = Radix::Tree(Route).new
+        @routes = Amber::Router::RouteSet.new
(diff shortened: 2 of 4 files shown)


src/amber/router/route_set.cr
```diff
@@ -0,0 +1,285 @@
...
+        end
+
+        # If this segment started as a leaf, no remant path exists. Match against the whole path.
+        remnant_path = path if was_leaf
+
+        # If this wasn't a leaf and there are no branches left, it's not a match.
+        return Tuple(Bool, String).new(false, "") unless @branches.any? || was_leaf
+
+        # If this node is the glob, at least one subtree matched (or there are none).
+        return Tuple(Bool, String).new(true, "") if glob?
```
robacarp: there has to be a better way to do this, right?

src/amber/router/route_set.cr
```diff
@@ -0,0 +1,285 @@
...
+      def reverse_select_routes!(path : String) : Tuple(Bool, String)
+        remnant_path = ""
+        was_leaf = leaf?
+
+        @branches.select! do |subtree|
+          match, remaining_path = subtree.reverse_select_routes! path
+          if match
+
+            if remnant_path != ""
+              raise "warning: overwriting remnant path"
```
robacarp: raising here is not the right thing to do, but I'm not sure what is

src/amber/router/route_set.cr
```diff
@@ -0,0 +1,285 @@
...
+      end
+
+      # Find a route which has been assigned to a matching path
+      # Weakness: assumes only one route will match the path query.
+      def find(path) : Amber::Route?
+        matches = deep_clone
+        matches.select_routes!(path)
+
+        if matches.size > 1
+          raise "Warning: matched multiple routes"
```
robacarp: raising may or may not be the right thing to do here

src/amber/router/router.cr
```diff
@@ -75,13 +77,10 @@ module Amber
         end
       end
 
-      def match(http_verb, resource) : Radix::Result(Amber::Route)
+      def match?(http_verb, resource) : Amber::Route?
```
robacarp: this will probably have to change when route parameters are actually handled by the router

src/amber/router/router.cr
```diff
@@ -34,24 +34,26 @@ module Amber
...
       end
 
       def socket_route_defined?(request)
         @socket_routes.map(&.[:path]).includes?(request.path)
       end
 
       def match_by_request(request)
-        match(request.method, request.path)
+        unless route = match?(request.method, request.path)
+          raise Exceptions::RouteNotFound.new(request)
```
robacarp: this isn't the right thing to do because it causes the server to abort the request, but it's sufficient for testing the routeset code.

elorest:
Would it maybe be a better idea just to fork radix and fix it. 

I've also considered just having routes macro's generate case statements. 

`get "/posts/:id/edit", PostsController, :edit`

Could translate to...

```cr
def match(path)
  case path
  when /\/posts\/([^\/]\/edit/
    drains[{{pipeline}}].run do 
      controller = {{controller}}.new(context)
      controller.before_filters
      controller.edit
      controller.after_filters
  else
   #static handler
  end
end
```

WDYGT @eliasjpr @drujensen

robacarp:
I’m not familiar enough with macros to say with confidence I’d be able to solve the problems. 

As for forking radix, I considered that but the character wise comparison it does felt like a big enough difference to url segment comparison that it would mean rewriting most of it.

eliasjpr:
@robacarp this is a great contribution. Honestly as you stated this code will run quite frequently and because of that I would like for you to submit some benchmarks comparisons against radix.  I am very curious to see how this performs. and if it does well I am willing to push it to the team as the default router for amber. We would also make a `Amber::Router` repo so people can use it outside of amber as you suggested. 

Please reply at your earliest convenience since I am eager to know more about your proposed solution

drujensen:
@robacarp 💯 🥇 It would be great if this were a drop in replacement for radix.  Since its a completely different algorithm, I agree that submitting this to the radix_tree project doesn't make sense.

I attempted something similar several years ago but my effort failed to match the performance that radix provides.  If we can match or come close to where the performance is "good enough", I think we should pursue this further.

robacarp:
> If we can match or come close to where the performance is "good enough", I think we should pursue this further.

@drujensen I agree. At this time, radix with Amber has some bugs, and some pain points. Efficiency is important for this kind of code, but I also think it needs to be balanced with providing functionality that matches what people expect.

I made some tweaks to the code and pushed it up to [robacarp/amber_router](https://github.com/robacarp/amber_router), along with a quick script to run a series of benchmarks against radix.

Here are the results. As you can see, radix is _fast_:

```
/get/
router: root 130.33k (  7.67µs) (± 5.65%) 34.47× slower
 radix: root   4.49M (222.59ns) (± 6.88%)       fastest

/get/books/23/chapters
router: deep 118.23k (  8.46µs) (± 5.14%)  8.46× slower
 radix: deep    1.0M ( 999.8ns) (± 9.02%)       fastest

/get/books/23/pages
router: wrong 128.18k (   7.8µs) (± 8.50%)  8.36× slower
 radix: wrong   1.07M (932.76ns) (±16.25%)       fastest

/get/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q/r/s/t/u/v/w/x/y/z
router: many segments  85.65k ( 11.68µs) (± 7.52%) 17.46× slower
 radix: many segments    1.5M (668.64ns) (± 2.99%)       fastest

/get/var/2/3/4/5/6/7/8/9/0/1/2/3/4/5/6/7/8/9/0/1/2/3/4/5/6
router: many variables  73.38k ( 13.63µs) (±14.64%)  3.41× slower
 radix: many variables 250.26k (   4.0µs) (± 7.06%)       fastest

/get/foobarbizfoobarbizfoobarbizfoobarbizfoobarbizbat/3
router: long_segments 122.76k (  8.15µs) (±16.26%)  5.61× slower
 radix: long_segments 688.98k (  1.45µs) (±20.05%)       fastest
```

One of the changes I made was to switch to an array based system, but I think it could be improved by using Tuples. However, I don't know of a way to pop/shift off of a Tuple without passing through an array. I think my lack of experience with Crystal is inhibiting me here.

I spent some time trying to use gdb to crack open the debug binary to see what the profiling looked like and there's some work to be done to get it to debug on macos and I didn't at the time have a VM handy.

I think there's a lot of speed to be gained here, but getting "my code" in or not isn't the point. I'm hoping to have a discussion about the best way for amber to route requests to controller/actions. I'm open to all suggestions 🍏

robacarp:
Update: I spent some time toying with variations on the tree algorithm and came up with something that is both [better organized code](https://github.com/robacarp/amber_router/tree/object_tree) and much faster. There is still some work to do on that branch to achieve full functional parity, but this at least puts the code in the ballpark.

```
/get/
router: root   2.38M ( 419.5ns) (± 6.55%)  1.89× slower
 radix: root   4.51M (221.65ns) (± 6.83%)       fastest


/get/books/23/chapters
router: deep   1.04M (961.16ns) (± 6.98%)       fastest
 radix: deep 993.34k (  1.01µs) (± 3.17%)  1.05× slower


/get/books/23/pages
router: wrong   1.33M (754.35ns) (± 5.07%)       fastest
 radix: wrong   1.13M (886.86ns) (± 3.20%)  1.18× slower


/get/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q/r/s/t/u/v/w/x/y/z
router: many segments 179.26k (  5.58µs) (± 9.40%)  8.21× slower
 radix: many segments   1.47M (679.53ns) (± 3.68%)       fastest


/get/var/2/3/4/5/6/7/8/9/0/1/2/3/4/5/6/7/8/9/0/1/2/3/4/5/6
router: many variables  201.1k (  4.97µs) (± 8.05%)  1.20× slower
 radix: many variables 240.32k (  4.16µs) (± 4.68%)       fastest


/get/foobarbizfoobarbizfoobarbizfoobarbizfoobarbizbat/3
router: long_segments   1.33M (751.86ns) (± 3.56%)       fastest
 radix: long_segments 880.11k (  1.14µs) (± 2.36%)  1.51× slower
```

drujensen:
@robacarp This is excellent news!  

The `deep`, `wrong`, and `long_segments` are more realistic types of routes we will see and I'm glad to see router can beat the performance of radix.  

The `many segments` and `many variables` are unrealistic and I think it's fine that router performs slower.  Even 8x slower is acceptable imo.

What I am a little concerned of is the `root` being almost 2x slower.  This will be the most common type of route.

marksiemers:
@drujensen - Maybe for the root, just have a case statement in front of the tree lookup? 

Perhaps here: https://github.com/robacarp/amber_router/blob/object_tree/src/route_set.cr#L165
Something to this effect could be added:
```crystal
def find(path) : RoutedResult(T)
  return @@root_route if @@root_route.matches(path)
  ...
end
```

Of course, this means storing to root route as a class variable (or otherwise) and treating it differently. I think since by definition there is only one root, this should be acceptable.

robacarp:
@marksiemers @drujensen the root route can be special cases but there’s another broader optimization that can happen that is a little broader. 

Right now the router doesn’t stop when it finds a route, it continues parsing through the rest of the tree. For a few reasons I can’t demonstrate now while I’m on mobile, I think this _might_ be the right choice. However, one way to solve the root performance is to short circuit and return when the first match is found. 

Unrelated, in Gitter someone asked to see the performance comparison with thousands of routes. For some trees of “thousands of routes” the performance of the router will suffer greatly but the performance scale isn’t unsolvable. If it’s needed, I think the router could pretty readily be extended to include different matching algorithms which optimize for different routing tiers. Does anyone have thoughts on what these sorts of demands would be?

drujensen:
I have been developing web apps for many years (20+) and have never built a web app that has thousands of routes. I have built dynamic routes using variables.  

I don’t see thousands of routes as a concern but maybe I’m missing something.

robacarp:
Back on a keyboard, here is the situation that this tree can get caught in that I think could cause some surprises. **Routes are not always selected in a top down fashion.**

This code builds a route tree, similar to the way radix runs, but it does it off of url segments. This is good, because trees are fast! For example this code: 

```crystal
route_set = Amber::Router::RouteSet(Symbol).new
route_set.add "/get/", :root
route_set.add "/get/users/:id", :users
route_set.add "/get/users/:id/books", :users_books
route_set.add "/get/books/:id", :books
route_set.add "/get/books/:id/chapters", :book_chapters
route_set.add "/get/books/:id/authors", :book_authors
route_set.add "/get/books/:id/pictures", :book_pictures
route_set.add "/get/users/:id/pictures", :users_pictures
```

Builds a route tree that looks like this:
```
  |--get
      |--(/get/)           # nodes wrapped in parenthesis are the routeable urls
      |--users
          |--:id
              |--(/get/users/:id)
              |--books
                  |--(/get/users/:id/books)
              |--pictures
                  |--(/get/users/:id/pictures)
      |--books
          |--:id
              |--(/get/books/:id)
              |--chapters
                  |--(/get/books/:id/chapters)
              |--authors
                  |--(/get/books/:id/authors)
              |--pictures
                  |--(/get/books/:id/pictures)
```

Insertion order is suggestive of matching order, but not indicative. This turns out to be important in this scenario:

```crystal
route_set = Amber::Router::RouteSet(Symbol).new
route_set.add "/get/*/slug", :slug
route_set.add "/get/products/:id/reviews", :amazon_style
route_set.add "/get/*", :catch_all
```

which builds this tree:
```
  |--get
      |--*
          |--slug
              |--(/get/*/slug)
          |--(/get/*)              # notice this route
      |--products
          |--:id
              |--reviews
                  |--(/get/products/:id/reviews)
```

The subtlety here is that someone from a rails background would expect the `catch_all` route to never match the url `/products/23/reviews` but it _will_. Because the route tree adds the `*` node before products, and the catch-all route is added as a subtree to that same `*` node, the router will actually match the `*` route before it matches the correct `products/:id/reviews` route.

The way this router currently handles that use case is that the first action is returned, but I think it can be a little more clever about it than that. I can think of a couple ways to prioritize routes in this way:

- I'd expect the router to return the most restrictive matching url. This could be a little more "fuzzy" but counting route params for each matched route, sorting, and returning the route with the least params is one idea.
- It is also possible to keep track of insertion order at a small cost by keeping a running count of insertions. Again, when multiple routes are returned, sort by this value and return the route with the smallest number.
- I could see both being relevant sorting patterns, and sorting first by "restrictive matching" and then by insertion order as a fallback to when two routes match but have the same level of matching.

Sorry for the book. It's kind of a subtle decision to make and I think it requires a bit of background. Hopefully it makes sense. 

@drujensen @marksiemers

eliasjpr:
@robacarp Thank you for the deep dive of the implementation of the router. I really do like it and I think it would be prudent to start an experiment branch using this router. The benefits and added features compensate the performance IMO. 

Lets start working on this router in a separate feature/experiment branch, if you don't mind I can create an Amber::Router repository and add you to it as the maintainer. I see this router being fully implemented by 0.5 version if not before, I guess depending on the progress, and how comfortable everyone feels, but not for version 0.4.0 since we already have defined the milestone. I do think we should try to optimize for performance as much as possible before we default to this router.   

With that said I think the team agrees that this is very much needed and worth while.

Thank you in advance for all your dedication

drujensen:
@robacarp Thanks for the book. ;-)

So would the process would be to look for all matches, sort and then choose the most restrictive?

In your example, `/get/products/3/reviews` would have 2 matches, so it would return both of them, then it would sort based on some algorithm?

What do you think of some sort of scoring algorithm, so * would be 5 points, variables would be 3 points and exact matches would be 1 point.  The match with the lowest point total would win? Just thinking out load here.  This may be too unpredictable.

robacarp:
@drujensen I'd considered some point based system as well but I think something a little more predictable and documentation friendly will be easier for people to grok. I think tracking the insertion order is easier to implement and easier to communicate.

src/amber/router/route_set.cr
marksiemers: A tuple literal:
```crystal
return { true, "" } if glob?
```
See https://carc.in/#/r/368l

robacarp: @marksiemers thanks! I did end up figuring that side out but I didn’t find a terse way to specify a tuple return type. Is there something more readable than explicitly typing [this function](https://github.com/amberframework/router/blob/master/src/route_set.cr#L160) with a tuple class literal?

marksiemers: I don't know if this creates the code style that you're looking for, but you can always be explicit about the return type (the compiler will complain if it finds a code path that returns something else - not sure about exception raising):
```crystal
# This might work?
def reverse_select_routes(path : Array(String), startpos, endpos = nil) : Tuple(Array(T), Int32)
  ...
end
```

robacarp:
@amberframework/contributors Thanks for exploring this with me. I'll keep iterating on the  [amber/router](/amberframework/router) shard, and getting it into better shape. I'll open another pull here when it's ready to start experimenting with.

PR #423, opened 2017-12-01
[ENV] Creates an Environment Module

eliasjpr:
### Description of the Change

This change facilitates the reuse of **Amber::Settings** encapsulating common functionality into an **environment** module.

- Adds environment module
- Moves **Environment**, **Environment Loader** and **Settings** to **Environment** module
- Moves **MainCommand** to its own file
- Converts **FileEncryptor** to a class
- **FileEncryptor** gets injected into the **EnvironmentLoader**
- Includes the **Environment** Module in **Amber** and **Amber::CLI**
- Fixes Specs

After this change the CLI and Amber will be able to share the same Amber.settings scope only loading the necessary module and files

The following methods are available to Amber and Amber::CLI

```crystal
settings
logger
env=
env
file_encryptor 
```

### Benefits

- High reusability

### Possible Drawbacks

- The private **encryption_key** is only used for encrypting and decrypting file and should probably not live where it is now.


src/amber/environment.cr
```diff
@@ -0,0 +1,40 @@
...
+
+    def self.env
+      @@env ||= Environment::Environment.new(CURRENT_ENVIRONMENT)
+    end
+
+    def self.file_encryptor
+      @@file_encryptor ||= Support::FileEncryptor.new(encryption_key)
+    end
+
+    private def self.encryption_key
```
elorest: Encryption key shouldn't be here since it's only used inside of FileEncryptor. Leaving encryption_key inside of FileEncryptor is more inline with SOLID principles.

src/amber/cli/commands.cr
```diff
@@ -16,41 +16,5 @@ class Cli::Command
 end
 
 module Amber::CLI
-  class MainCommand < ::Cli::Supercommand
```
elorest: Can we leave this here for now as it's not related to this PR? As the number if files increase for non abstraction reasons code bases become harder to navigate.

src/amber/environment.cr
```diff
@@ -0,0 +1,40 @@
...
+  macro included
+    SECRET_KEY  = "AMBER_ENCRYPTION_KEY"
+    SECRET_FILE = "./.encryption_key"
+    CURRENT_ENVIRONMENT = ENV["AMBER_ENV"] ||= "development"
+    class_property environment_path : String = "./config/environments/"
+
+    @@_settings : Settings?
+
+    def self.settings
+      @@_settings ||= EnvironmentLoader.new(CURRENT_ENVIRONMENT, @@environment_path, file_encryptor).settings
```
elorest: WE need to be aware that loading settings here will cause the entire CLI to exception if ENV is encrypted and they don't have the key. This could lead to confusion. IF `AMBER_ENV=production` and someone calls `amber g scaffold` or `amber routes` or any call not related to the environment it will still error.

Before this PR only a call to `amber database` or `amber encrypt` would have caused a CLI error.

src/amber/server/server.cr
```diff
@@ -99,7 +97,11 @@ module Amber
     end
 
     private def logger
-      settings.logger
+      Amber.logger
+    end
+
+    def settings
+      Amber.settings
```
elorest: I like this.

src/amber/support/file_encryptor.cr
```diff
@@ -1,29 +1,37 @@
...
-  module FileEncryptor
-    def self.encryption_key
-      ENV["AMBER_ENCRYPTION_KEY"]? || begin
-        if File.exists?(".encryption_key")
-          File.open(".encryption_key").gets_to_end.to_s
-        else
-          raise Exceptions::EncryptionKeyMissing.new
-        end
-      end
+  class FileEncryptor
```
elorest: I really wish this hadn't changed. We've added complexity here without improving functionality.

src/amber/environment/environment.cr
```diff
@@ -1,4 +1,4 @@
-module Amber
```
elorest: `environment/environment.cr` Doesn't look good. Could we call it `environment/env.cr` or `environment/name.cr`

src/amber/cli/commands.cr
eliasjpr: Will do

src/amber/environment.cr
```diff
@@ -0,0 +1,40 @@
...
+      ENV["AMBER_ENV"] = env.to_s
+      @@_settings = EnvironmentLoader.new(env, @@environment_path, file_encryptor).settings
+    end
+
+    def self.env
+      @@env ||= Environment::Environment.new(CURRENT_ENVIRONMENT)
+    end
+
+    def self.file_encryptor
+      @@file_encryptor ||= Support::FileEncryptor.new(encryption_key)
```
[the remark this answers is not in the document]
eliasjpr: :+1:

src/amber/environment.cr
eliasjpr: I agree and there was other part of the code using secret key. I think only 1 additional place

src/amber/environment/environment.cr
eliasjpr: I thought of this I will name it env

src/amber/environment/environment_loader.cr
```diff
@@ -1,15 +1,11 @@
-require "./settings"
-
-module Amber
+module Amber::Environment
```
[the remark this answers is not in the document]
eliasjpr: Same thought will name it loader

src/amber/environment/environment_loader.cr
```diff
@@ -20,16 +16,16 @@ module Amber
...
     end
 
     private def yml_settings_file
-      @yml_settings ||= "#{@path}/#{@environment}.yml"
+      @yml_settings ||= File.expand_path("#{@path}/#{@environment}.yml")
     end
 
     private def enc_settings_file
-      @enc_settings ||= "#{@path}/.#{@environment}.enc"
+      @enc_settings ||= File.expand_path("#{@path}/.#{@environment}.enc")
```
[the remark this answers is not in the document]
eliasjpr: People can store their environment config setting in a higher order directory outside of the project without expanding the path can run into a different path

src/amber/support/file_encryptor.cr
eliasjpr: I make it a module again.

src/amber/environment.cr
eliasjpr: The thought behind this is that we can have access to the same functionality from the cli to work with the environment.

src/amber/environment/environment_loader.cr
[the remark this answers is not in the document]
elorest: Could you show an example of how that would work with File.expand_path and not with just @path?

src/amber/environment.cr
elorest: If something else uses that it really shouldn't be. FileEncryptor should work in a vacuum and everything related to encrypted files should go through it.

src/amber/environment.cr
elorest: Yeah I totally get that. I'm just afraid that it will lead to issues. We could solve them by catching the exception in this case and loading defaults.

src/amber/cli/commands.cr
elorest: Thanks :)

eliasjpr: Leaving this since this is relative a small change.

src/amber/support/file_encryptor.cr
```diff
@@ -1,29 +1,26 @@
...
       File.write(filename, encryptor.encrypt_and_sign(body))
     end
 
-    def self.read_as_string(filename, encryption_key = self.encryption_key)
+    def self.read_as_string(filename, encryption_key = encryption_key)
       String.new(read(filename, encryption_key))
     end
+
+    private def self.encryption_key
+      ENV[SECRET_KEY]? || File.open(SECRET_FILE).gets_to_end.to_s
```
elorest: Why does this no longer raise an exception? If it tries reading a file that doesn't exist it will anyway. Since we're loading settings with cli now instead of for specific commands this will throw an exception unexpectedly.

Lets pair on this.

src/amber/support/file_encryptor.cr
```diff
@@ -1,29 +1,26 @@
...
+  SECRET_FILE = "./.encryption_key"
 
-    def self.read(filename : String, encryption_key = self.encryption_key)
+  module FileEncryptor
+    def self.read(filename : String, encryption_key = encryption_key)
       encryptor = Amber::Support::MessageEncryptor.new(encryption_key)
       encryptor.verify_and_decrypt(File.open(filename).gets_to_end.to_slice)
     end
 
-    def self.write(filename : String, body : (String | Bytes), encryption_key = self.encryption_key)
```
elorest: `encryption_key = self.encryption_key` looks a lot better because it doesn't look like it's assigning itself to itself.

src/amber/environment/loader.cr
```diff
@@ -25,11 +21,11 @@ module Amber
     end
 
     private def yml_settings_file
-      @yml_settings ||= "#{@path}/#{@environment}.yml"
+      @yml_settings ||= File.expand_path("#{@path}/#{@environment}.yml")
```
elorest: I still don't get why we're expanding the path. It results in the same thing anyway.

src/amber/environment/loader.cr
```diff
@@ -1,15 +1,11 @@
-require "./settings"
-
-module Amber
-  class EnvironmentLoader
-    # TODO: Should probably use Amber::Env type here but can't until we refactor the Amber module.
-    # Should default path to Amber.environment_path
-    def initialize(@environment : String | Symbol = :development, @path : String = "./config/environments")
+module Amber::Environment
+  class Loader
+    def initialize(@environment : Amber::Environment::EnvType, @path : String)
```
elorest: Could we default these to `:development` and `@@environment_path` or something?

src/amber/environment/loader.cr
```diff
@@ -1,15 +1,11 @@
...
       raise Exceptions::Environment.new(@path, @environment) unless settings_file_exist?
     end
 
     def settings
-      Amber::Settings.from_yaml(settings_content.to_s)
+      Settings.from_yaml(settings_content.to_s)
```
elorest: This is cleaner!

src/amber/cli/commands/main_command.cr
```diff
@@ -0,0 +1,38 @@
+module Amber::CLI
```
elorest: From an organization perspective it's cleaner to leave the main command outside of commands since it's a different namespace, than the things inside of commands.

src/amber/cli/commands.cr
elorest: This made more sense here since main command has a different namespace than the other commands.

src/amber/environment/loader.cr
eliasjpr: `Amber.environment_path` allows the user to set a different path for the yamls files. For instance `~/Some/Dir`, without expand path this will error.

src/amber/support/file_encryptor.cr
eliasjpr: One of the reasons that I went with a class was because of this and the encryption_key default value. With the class it guarantee to have a key before using the FileEncryptor and there is no assumptions made about a default value for the key

src/amber/environment/loader.cr
eliasjpr: This class should not make assumptions about the environment and the path. To ensure The default values are set  to the _Amber.environment_path_ property and **Amber.env.to_s**

src/amber/environment/loader.cr
elorest: Not sure that should be encouraged... The only way to do that in rails is with a symlink.

src/amber/support/file_encryptor.cr
elorest: True but that's just taking the problem and moving it further out from the center. It still may or may not exist and needs to be dealt with regardless of where we put it.

Moving it further from it's use case just adds complexity to the problem.

src/amber/environment/loader.cr
elorest: Defaults are fine as long as they can be overwritten, IMO.

src/amber/environment/loader.cr
elorest: NVM. I think you're right about this.

PR #426, opened 2017-12-01
[CLI] Add Amber::CLI::Config to cleanup database, language and model

eliasjpr:
The code looks into a `./.amber.yml` file to load predefined settings
for Amber::CLI, there is a lot of duplicate code around loading this
settings.

This Removes the code duplication by creating a Config module to load
the cli specific settings.

- Adds Amber::CLI::Config module
- Removes duplicate code


src/amber/cli/config.cr
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def self.database
+      amber_yml["database"].to_s
+    rescue
+      "pg"
+    end
+
+    def self.model
+      p amber_yml["model"]
```
faustinoaq: Debugging code?

faustinoaq:
Interesting amber is becoming more easy to read and contribute, because latest changes are removing a lot of duplicate code 👍

src/amber/cli/config.cr
```diff
@@ -0,0 +1,27 @@
...
+      amber_yml["model"].to_s
+    rescue
+      "granite"
+    end
+
+    private def amber_yml
+      raise "Missing Amber YAML file" unless File.exists?(AMBER_YML)
```
marksiemers: Again, I don't think raising an exception is the way to go here. At least not if it is going to be rescued by the methods above.

```crystal
private def amber_yml
  return nil unless File.exists?(AMBER_YML)
  @@amber_yml ||= YAML.parse(File.read AMBER_YML)
end
```

src/amber/cli/templates/auth.cr
```diff
@@ -14,8 +14,8 @@ module Amber::CLI
     @primary_key : String
 
     def initialize(@name, fields)
-      @database = database
-      @language = language
+      @database = Amber::CLI::Config.database
+      @language = Amber::CLI::Config.language
```
marksiemers: Do we want to fully namespace this, or have an `include Config` at the top of this class?

Like this: https://github.com/amberframework/amber/blob/master/src/amber/cli/templates/auth.cr#L7

src/amber/cli/templates/controller.cr
```diff
@@ -7,10 +7,9 @@ module Amber::CLI
 
     @name : String
     @actions = Hash(String, String).new
-    @language : String
+    getter language : String = Amber::CLI::Config.language
```
marksiemers: Similar comment to above. This is already in the `Amber::CLI` namespace, so it should work with just `Config.language`

src/amber/cli/templates/error.cr
```diff
@@ -8,8 +8,8 @@ module Amber::CLI
     @language : String
 
     def initialize(@name, actions)
-      @language = fetch_language
-      @actions = actions.map{|action| [action, "get"] }.to_h
+      @language = Amber::CLI::Config.language
```
marksiemers: `Config.language` should work.

src/amber/cli/templates/mailer.cr
```diff
@@ -10,24 +10,14 @@ module Amber::CLI
     @fields : Array(Field)
 
     def initialize(@name, fields)
-      @language = language
+      @language = Amber::CLI::Config.language
```
marksiemers: Config.language should work.

src/amber/cli/templates/migration.cr
```diff
@@ -13,7 +13,7 @@ module Amber::CLI
 
     def initialize(@name, fields)
       @fields = fields.map { |field| Field.new(field) }
-      @database = database
+      @database = Amber::CLI::Config.database
```
marksiemers: Config.language should work.

src/amber/cli/templates/model.cr
```diff
@@ -10,7 +10,7 @@ module Amber::CLI
     @database : String
 
     def initialize(@name, fields)
-      @database = database
+      @database = Amber::CLI::Config.database
```
marksiemers: Config.language should work.

src/amber/cli/templates/scaffold/controller.cr
```diff
@@ -12,8 +12,8 @@ module Amber::CLI::Scaffold
     @fields_hash = {} of String => String
 
     def initialize(@name, fields)
-      @language = language
-      @database = database
+      @language = Amber::CLI::Config.language
+      @database = Amber::CLI::Config.database
```
marksiemers: `Config.language` and `Config.database` should work.

src/amber/cli/config.cr
```diff
@@ -0,0 +1,27 @@
+module Amber::CLI
+  module Config
+    def self.language
+      amber_yml["language"].to_s
+    rescue
```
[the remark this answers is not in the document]
eliasjpr: I agree. An why it is still WIP. I do not like the amber_yml check here though I think if the method has reached this far is because amber_yml passed

src/amber/cli/config.cr
eliasjpr: This makes me thing that the config should be a class instead of a module.

marksiemers: High level, I think parsing a file is the job of a module and holding config information is the job of a class.

We have other yaml files that are loaded and written, correct? There could be a module that takes care of all of that, and a config class could leverage that module.

PR #428, opened 2017-12-01
Add support for PORT environment variable

faustinoaq:
### Description of the Change

Add support for PORT environment variable, 3000 is used by default.

```yaml
port: (ENV["PORT"]? || 3000).to_i
```

### Alternate Designs

No

### Benefits

- Allow to use [heroku-buildpack-amber](https://github.com/amberframework/heroku-buildpack-amber/blob/master/bin/release#L3)
- Allow to change PORT easier on services with random port like Heroku

### Possible Drawbacks

No, if PORT isn't defined port 3000 is used

### DIFF
--- a/src/amber/cli/templates/app/config/environments/development.yml.ecr
+++ b/src/amber/cli/templates/app/config/environments/development.yml.ecr
@@ -1,5 +1,5 @@
 secret_key_base: <%= SecureRandom.urlsafe_base64(32) %>
-port: 3000
+port: (ENV["PORT"]? || 3000).to_i
 name: <%= @name %>
 colorize_logging: true
 host: localhost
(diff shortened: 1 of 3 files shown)


faustinoaq:
I just realize this PR isn't needed, because now we are using `config/application.cr` to overwrite environment files, so we just need to fix PORT documentation [here](https://github.com/amberframework/amber/blob/support-port-env/src/amber/cli/templates/app/config/application.cr.ecr#L64) (Currently this has a comment in wrong place, see: `settings.process_count` vs `settings.port` )

elorest:
This was removed recently by @eliasjpr as env.yml's no longer support dynamic code. I'll submit a PR that fixes this now.

faustinoaq:
@elorest Yeah, I was using amber `v0.3.6` so I didn't notice this change

I think avoiding dynamic code in yaml files is ok, let me close this 👍

elorest:
https://github.com/amberframework/amber/pull/429

Please review this!

PR #429, opened 2017-12-02
Improved generated config/application.cr commented code to work once uncommented.

elorest:
### Description of the Change

1. Uncommented PORT env logic by default and converted to int32.
1. Added valid secret_key_base to commented out secret_key_base settings code.
1. Added checks so that ENV vars are only applied if they exist. That way the standard practice of using default settings in development and test but ENV vars in production will still work without constant code changes. 
1. Changed example settings values to match correct type for setting (bool, int32) etc. ENV vars are all strings and fail by default.

### Benefits

Within Rails and Phoenix every ENV override has default values that are used instead if they're not set.

For example: If a PORT isn't set it goes to 3000, if an RAILS_ENV isn't set it goes to development, if DATABASE_URL isn't set it uses settings from database.yml.
I don't think that it's clear to users that uncommenting the code in application.cr will work differently.
For instance if settings port is 3000, ENV["PORT"] should only override that if ENV["PORT"] has a value. Otherwise we just end up with a project that doesn't work.

`settings.port = ENV["PORT"]` will fail in too many situations that aren't expected by people coming from rails/laravel, pheonix, hanami.

Instead it should be `settings.port = ENV["PORT"].to_i if ENV["PORT"]?`.

Otherwise we just end up with a weird situation were we load values in and immediately overwrite them with values that may not even exist.

***We now end up with example code that will work out of the box, which will result in fewer complaints.***

### DIFF
--- a/src/amber/cli/templates/app/config/application.cr.ecr
+++ b/src/amber/cli/templates/app/config/application.cr.ecr
@@ -58,14 +58,14 @@ Amber::Server.configure do |settings|
   #
   # > Read more about Linux PORT REUSE https://lwn.net/Articles/542629/
   #
-  # settings.port_reuse = ENV["PORT_REUSE"] if ENV["PORT_REUSE"]?
+  # settings.port_reuse = true
   #
   #
   # Process Count: This will enable Amber to be used in cluster mode,
   # spinning an instance for each number of process specified here.
   # Rule of thumb, always leave at least 1 core available for system processes/resources.
   #
-  # settings.process_count = ENV["PROCESS_COUNT"] if ENV["PROCESS_COUNT"]?
+  # settings.process_count = ENV["PROCESS_COUNT"].to_i if ENV["PROCESS_COUNT"]?
   #
   #
   # PORT: This is the port that you're application will run on. Examples would be (80, 443, 3000, 8080)
--- a/src/amber/cli/templates/app/config/application.cr.ecr
+++ b/src/amber/cli/templates/app/config/application.cr.ecr
@@ -41,14 +41,14 @@ Amber::Server.configure do |settings|
   # initialized to a random key present in `ENV["AMBER_SECRET_KEY"]` or
   # `.amber_secret_key` in this order.
   #
-  # settings.secret_key_base= ENV["SECRET_KEY_BASE"]
+  # settings.secret_key_base= <%= SecureRandom.urlsafe_base64(32) %>
   #
   #
   # Host: is the application server host address or ip address. Useful for when
   # deploying Amber to a PAAS and likely the assigned server IP is either
   # known or unknown. Defaults to an environment variable HOST
   #
-  # settings.host = ENV["HOST"]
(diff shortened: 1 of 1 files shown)


src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -58,28 +58,33 @@ Amber::Server.configure do |settings|
...
-  # settings.port_reuse = ENV["PORT_REUSE"]
+  # settings.port_reuse = true
   #
   #
   # Process Count: This will enable Amber to be used in cluster mode,
   # spinning an instance for each number of process specified here.
   # Rule of thumb, always leave at least 1 core available for system processes/resources.
   #
-  # settings.port = (ENV["PORT"] ||= 80).to_i
+  # settings.process_count = ENV["PROCESS_COUNT"].to_i if ENV["PROCESS_COUNT"]?
```
marksiemers: Should there be a hardcoded fallback here?
```crystal
settings.process_count = ENV["PROCESS_COUNT"]? ? ENV["PROCESS_COUNT"].to_i : 1 
```

src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -58,28 +58,33 @@ Amber::Server.configure do |settings|
...
+  #
+  #
+  # PORT: This is the port that you're application will run on. Examples would be (80, 443, 3000, 8080)
+  #
+  settings.port = ENV["PORT"].to_i if ENV["PORT"]?
```
marksiemers: Same question here, should there be a default, or do we want it to fail loudly if that ENV variable isn't there?

src/amber/cli/templates/app/config/application.cr.ecr
faustinoaq: We already have `process_count = 1` [here](https://github.com/amberframework/amber/blob/c7ffd11d6c65c5347aca41fa0580c7c8d6265e60/src/amber/server/settings.cr#L27)

Notice `]?` , so returns `nil` if not found and setting attribute is not assigned.

src/amber/cli/templates/app/config/application.cr.ecr
faustinoaq: We're already using `port = 3000` [here](https://github.com/amberframework/amber/blob/c7ffd11d6c65c5347aca41fa0580c7c8d6265e60/src/amber/server/settings.cr#L25)

`:ditto:`

elorest:
It's pointless to have a default if we're going to overwrite it with a non existent environment var.

I strongly feel that we should follow the same pattern that all the other Frameworks we have used for the last 10 years follow.

Use the default unless the environment variable exists.

src/amber/cli/templates/app/config/application.cr.ecr
elorest: Yes. There is already a fall back for all of them. They just weren't being used if the environment var was missing.

PR #431, opened 2017-12-03
Update dependency and move sentry_helper to helper dir

faustinoaq:
### Description of the Change

- Update amber dependency, depends on https://github.com/amberframework/quartz-mailer/pull/8 **(Done!)**
- Moves `sentry_command_helper.cr` to `helpers/sentry.cr`

### Alternate Designs

No

### Benefits

Use latest version, fixes this 👉 https://github.com/crystal-lang/shards/issues/183

### Possible Drawbacks

No, this time no more version mismatch 🎉 people using old projects can install dependencies without issues because `~> 0.2.0` is equal to `0.2.0 <= version < 0.3.0`

So `v0.3.0` will apply only for new amber projects 👍

### DIFF
--- a/src/amber/cli/commands/routes.cr
+++ b/src/amber/cli/commands/routes.cr
@@ -1,6 +1,6 @@
 require "cli"
 require "shell-table"
-require "../helper/sentry"
+require "../helpers/sentry"
 
 module Amber::CLI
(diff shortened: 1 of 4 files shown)


faustinoaq:
Depends on https://github.com/amberframework/quartz-mailer/pull/8 to be able to pass Travis

PR #432, opened 2017-12-03
switch to amber_router

robacarp:
### Description of the Change

As minimally as possible, swap out radix for the new amber/router shard.

### Benefits

As described in #416, and on the https://github.com/amberframework/router readme. The highlights are:

- no problem with two different parameter names at the same url location: `/users/:id` and `/user/:user_id/books` routes can coexist.
- provides the ability to write glob routes with suffixes: `/products/*slug/comments`
- tree re-traversal built right in, so there's no risk of getting stuck in one wing of the tree and failing to match anything.

### Possible Drawbacks

The code is pretty fresh, and it'll need to be put through some paces. I'm happy to screenshare to explain through the code and take feedback if that's easier.

### Still needs:

- [x] In my personal application, I needed to change the way the static route handler was defined but I know there's been some changes there. Might need a generator change or two. Addressed by #583 

- [x] More testing. **If you have an application**, please try pointing your shards file at this fork of amber and see what happens when routes are generated, parsed, and matched.


elorest:
@jwaldrip Just wrote an improved radix tree as well. I would suggest checking it out.

marksiemers:
@robacarp - Should this remain open?

There may be possible improvements over Radix, but we aren't sure if it is causing a significant bottleneck for amber.

The features that you mentioned - are those dealbreakers for some people (using radix)?

robacarp:
@marksiemers,

Good question. I haven’t tested the latest radix to see if it fails to renavigate a route tree when it mismatches or not. My last look, my project wouldn’t render any files from `public/dist` because I had a route elsewhere that started with a `d`, which I consider a total showstopper for Amber.

marksiemers:
@robacarp - Do you know if that is a limitation of the radix tree traversal algorithm itself, or the implementation in crystal?

Is it possible to extend radix tree to handle a "re-traversal"?

robacarp:
@marksiemers I'm not intimate with the formal definition of the radix tree traversal algorithm, but the tree traversal proposed in this pull request is by definition similar because of the problem space overlap.

The biggest difference is that instead of rejecting all sibling nodes when a single node matches, every node must explicitly fail to match. Candidate matches which match all route segments are then sorted by declaration order and precedence is given to routes by order of definition.

Second to that, amberframework/router builds a tree out of whole url segments instead of individual characters. I don't believe there is any performance gain to doing this, but to my very subjective eye the code is much more readable, especially when unicode codepoints get involved.

There are a few other upgrades this router could provide Amber if the core team wants to include it. [They are documented here](https://github.com/amberframework/amber/pull/416) in the section labeled _benefits_.

src/amber/router/router.cr
```diff
@@ -58,24 +55,7 @@ module Amber
         @routes_hash["#{controller}##{action}"]
       end
 
-      def all
```
eliasjpr: This is fine I added those methods to for the `amber routes` command so we can get the routes that have been register with the app. It turns out that we are currently parsing the routes.cr file which does not tell the full truth since routes can be register via code.

src/amber/router/context.cr
```diff
@@ -53,7 +53,7 @@ class HTTP::Server::Context
   end
 
   def request_handler
-    route.payload
+    route.payload.not_nil!
```
eliasjpr: A route should always match no?

robacarp: Yes because routes which don't match are aborted [over here](https://github.com/robacarp/amber/blob/ae786b5f79ee648dcec4b5fd6198b27bed9f8d9a/src/amber/pipes/pipeline.cr#L16).

marksiemers:
The failure is a real failure, either the tests need to be updated to reflect new functionality or syntax, or there is something the new router isn't handling as we expect:
```
  2) PostControllerTest renders post show template
       Missing hash key: :static
```

shard.yml
```diff
@@ -14,9 +14,9 @@ targets:
     main: src/amber/cli.cr
 
 dependencies:
-  radix:
-    github: luislavena/radix
-    version: ~> 0.3.8
+  amber_router:
+    github: amberframework/router
+    commit: 5883a23
```
faustinoaq: @robacarp I think we should release a new amber_router version to avoid using commit

robacarp: @faustinoaq Yeah, when this gets settled, I'll make a real release tag and update here before merging.

robacarp:
Test failure in generated app was caused by an interesting collision of a couple bugs:

- Garnet specs generated by `amber generate scaffold` do not include a Static valve, but the application routes file is read which references a static valve.
- The bug in #583 is more important here because this RouteSet matches top-down more explicitly than radix, so the static wildcard route was matching and failing to find the valve for `:static`.

Moving the generated static route to the bottom of the file solves the test failure.

eliasjpr:
Im fine merging this in

PR #433, opened 2017-12-03
[Params] Appends Files to params hash

eliasjpr:
Currently there is `context.files` and `context.params` having two
scopes to what is considered params can be confusing.

The changes in this PR consolidates files with params

- Adds Amber::Router::Files::File to params hash
- Updates Specs and files

With this change files are now part of the `context.params`

### DIFF
--- a/src/amber/router/params.cr
+++ b/src/amber/router/params.cr
@@ -1,13 +1,7 @@
 module Amber::Router
-  # The Parameters module will parse parameters from a URL, a form post or a JSON
-  # post and provide them in the self params hash.  This unifies access to
-  # parameters into one place to simplify access to them.
-  # Note: other params from the router will be handled in the router handler
-  # instead of here.  This removes a dependency on the router in case it is
-  # replaced or not needed.
-
-  class Params < Hash(String, String | Amber::Router::Files::File)
+  class Params < Hash(String, String)
     alias KeyType = String | Symbol
+    property files = {} of String => Amber::Router::Files::File
 
     def json(key : KeyType)
       JSON.parse(self[key]?.to_s)
--- a/src/amber/router/params.cr
+++ b/src/amber/router/params.cr
@@ -6,7 +6,7 @@ module Amber::Router
   # instead of here.  This removes a dependency on the router in case it is
   # replaced or not needed.
 
-  class Params < Hash(String, String)
+  class Params < Hash(String, String | Amber::Router::Files::File)
     alias KeyType = String | Symbol
 
     def json(key : KeyType)
--- a/src/amber/router/params_parser.cr
+++ b/src/amber/router/params_parser.cr
@@ -115,10 +115,18 @@ module Amber::Router
     def parse_multipart
       HTTP::FormData.parse(request) do |upload|
         next unless upload
+        # Note:
+        # Filename is followed by a string containing the original name of the file transmitted.
+        # The filename is always optional and must not be used blindly by the application:
+        # path information should be stripped, and conversion to the server file system rules
+        # should be done. This parameter provides mostly indicative information.
+        #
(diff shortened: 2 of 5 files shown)


src/amber/validations/params.cr
```diff
@@ -55,7 +55,7 @@ module Amber::Validators
     getter raw_params = Amber::Router::Params.new
     getter errors = [] of Error
     getter rules = [] of BaseRule
-    getter params = {} of String => String?
+    getter params = {} of String => String? | Amber::Router::Files::File
```
robacarp: This union type is expressed several places, is it worth having an alias for it?

elorest:
In theory I like this and have proposed it many times. Ultimately we've left it like this, though because by adding a union we run into the same issue on param values that we did with JSON::Type. 

Because the value could be a String or a File we now have to check if it's a string everywhere before using the param.

Possibly this could be solved by inheriting File < String, where self is file name and other properties like body are added.

Since I wrote the file parsing param code I'd like to handle this change if possible.

src/amber/validations/params.cr
marksiemers: Agreed, perhaps:
```crystal
alias Amber::ParamValue = String | Amber::Router::Files::File | Nil
# then
String => Amber::ParamValue
```

eliasjpr:
@elorest this is a quick fix since someone today was asking why Amber has `context.files` and `context.params`.

 I made some `low hanging fruit` changes that should not interfere with Granite::ORM and allows for everything to work, the only difference is that Amber no longer has a `context.files` and instead has `context.params.files`. 

Let me know what you think.

src/amber/validations/params.cr
eliasjpr: This has change since the union here interferes with `Granite::ORM` when it shouldn't

eliasjpr:
Also I left a comment on the `filename` on the file upload parsing:

> The Parameters module will parse parameters from a URL, a form post or a JSON
   post and provide them in the self params hash.  This unifies access to
   parameters into one place to simplify access to them.
   Note: other params from the router will be handled in the router handler
   instead of here.  This removes a dependency on the router in case it is
   replaced or not needed.

See for more detail https://tools.ietf.org/html/rfc7578#section-4.2

src/amber/router/params.cr
```diff
@@ -1,13 +1,7 @@
...
-  # The Parameters module will parse parameters from a URL, a form post or a JSON
-  # post and provide them in the self params hash.  This unifies access to
-  # parameters into one place to simplify access to them.
-  # Note: other params from the router will be handled in the router handler
-  # instead of here.  This removes a dependency on the router in case it is
-  # replaced or not needed.
-
   class Params < Hash(String, String)
     alias KeyType = String | Symbol
+    property files = {} of String => Amber::Router::Files::File
```
elorest: Since this is inside of `Amber::Router` it should just be String => `Amber::Router::Files::File`

elorest: Since this is inside of `Amber::Router` it should just be String => `Amber::Router::Files::File`

src/amber/router/params_parser.cr
```diff
@@ -115,10 +115,18 @@ module Amber::Router
     def parse_multipart
       HTTP::FormData.parse(request) do |upload|
         next unless upload
+        # Note:
```
elorest: I don't think we need this note here. In the case that a file is submitted via a multipart form filename is there. If not it isn't. This check is important to make sure that empty files don't end up in the Files hash. It also saves us from time on parsing.

eliasjpr: The reason I added the note is because of this line: `The filename is always optional and must not be used blindly by the application:`

PR #434, opened 2017-12-04
Provide attributes for git config

faustinoaq:
### Description of the Change

- Add README enhancements
- Add attributes for `git config` (based on [crystal tools](https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/init.cr))

### Alternate Designs

No

### Benefits

Useful data on default README

### Possible Drawbacks

No

### DIFF
--- a/src/amber/cli/templates/app.cr
+++ b/src/amber/cli/templates/app.cr
@@ -1,7 +1,5 @@
 module Amber::CLI
   class App < Teeplate::FileTree
-    WHICH_GIT_COMMAND = "which git >/dev/null"
-
     directory "#{__DIR__}/app"
     getter database_name_base
 
@@ -33,19 +31,27 @@ module Amber::CLI
       @name.gsub('-', '_')
     end
 
+    def which_git_command
+      `which git >/dev/null`
+    end
+
     def fetch_author
-      return "[your-name-here]" unless system(WHICH_GIT_COMMAND)
-      `git config --get user.name`.strip
+      default = "[your-name-here]"
+      return default unless which_git_command
+      user_name = `git config --get user.name`.strip
+      user_name.empty? ? default : user_name
     end
 
     def fetch_email
-      return "[your-email-here]" unless system(WHICH_GIT_COMMAND)
-      `git config --get user.email`.strip
+      default = "[your-email-here]"
+      return default unless which_git_command
+      user_email = `git config --get user.email`.strip
+      user_email.empty? ? default : user_email
     end
 
     def fetch_github_name
       default = "[your-github-name]"
-      return default unless system(WHICH_GIT_COMMAND)
+      return default unless which_git_command
       github_user = `git config --get github.user`.strip
       github_user.empty? ? default : github_user
     end
--- a/src/amber/cli/templates/app.cr
+++ b/src/amber/cli/templates/app.cr
@@ -36,21 +36,21 @@ module Amber::CLI
     end
 
     def fetch_author
-      default = "[your-name-here]"
+      default = "your-name-here"
       return default unless which_git_command
       user_name = `git config --get user.name`.strip
       user_name.empty? ? default : user_name
     end
 
     def fetch_email
-      default = "[your-email-here]"
+      default = "your-email-here"
       return default unless which_git_command
       user_email = `git config --get user.email`.strip
       user_email.empty? ? default : user_email
     end
 
     def fetch_github_name
-      default = "[your-github-name]"
+      default = "your-github-name"
       return default unless which_git_command
       github_user = `git config --get github.user`.strip
       github_user.empty? ? default : github_user
--- a/src/amber/cli/templates/app.cr
+++ b/src/amber/cli/templates/app.cr
@@ -13,6 +13,7 @@ module Amber::CLI
     @author : String
     @email : String
     @github_name : String
+    @git_command : Bool
 
     def initialize(@name, @database = "pg", @language = "slang", @model = "granite")
       @db_url = ""
--- a/src/amber/cli/templates/app.cr
+++ b/src/amber/cli/templates/app.cr
@@ -36,24 +36,27 @@ module Amber::CLI
     end
 
     def fetch_author
-      default = "your-name-here"
-      return default unless which_git_command
-      user_name = `git config --get user.name`.strip
-      user_name.empty? ? default : user_name
+      if which_git_command
+        user_name = `git config --get user.name`.strip
+        user_name = nil if user_name.empty?
+      end
+      user_name || "your-name-here"
     end
 
     def fetch_email
-      default = "your-email-here"
-      return default unless which_git_command
-      user_email = `git config --get user.email`.strip
-      user_email.empty? ? default : user_email
+      if which_git_command
+        user_email = `git config --get user.email`.strip
+        user_email = nil if user_email.empty?
+      end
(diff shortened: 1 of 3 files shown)


src/amber/cli/templates/app/README.md.ecr
```diff
@@ -1,52 +1,84 @@
...
 ## Installation
 
+Install back-end and front-end dependencies.
+  
+```
+shards install
+npm install
+```
+
 Configure the `config/environments/development.yml` and set the `database_url` with your credentials to your database.
```
faustinoaq: Is this line still valid?, Should we add `amber encrypt` reference to this README too? What steps are needed to setup `production.yml`? What about `.encryption_key`?

src/amber/cli/templates/app/README.md.ecr
```diff
@@ -1,52 +1,84 @@
...
+npm run watch
+```
+
+### Production
+
+To build a production release:
+
+```
+AMBER_ENV=production
+shards build --production --release <%= @name %>
```
faustinoaq: Maybe we can use production automatically if `--release` flags is used? WDYT?

```
# config/application.cr
{% if flag :release %}
settings.env = "production"
{% end %}
```

robacarp: I'd rather not, that defeats the purpose of the environment variable.

src/amber/cli/templates/app.cr
```diff
@@ -24,5 +32,22 @@ module Amber::CLI
...
+    end
+
+    def fetch_email
+      return "[your-email-here]" unless system(WHICH_GIT_COMMAND)
+      `git config --get user.email`.strip
+    end
+
+    def fetch_github_name
+      default = "[your-github-name]"
+      return default unless system(WHICH_GIT_COMMAND)
```
veelenga: `system(WHICH_GIT_COMMAND)` will be called 3 times (`getch_author`, `fetch_email`, `fetch_github_name`). Better to check whether git exists or not earlier.

src/amber/cli/templates/app.cr
```diff
@@ -24,5 +32,22 @@ module Amber::CLI
...
     end
+
+    def fetch_author
+      return "[your-name-here]" unless system(WHICH_GIT_COMMAND)
+      `git config --get user.name`.strip
+    end
+
+    def fetch_email
+      return "[your-email-here]" unless system(WHICH_GIT_COMMAND)
+      `git config --get user.email`.strip
```
veelenga: User may not also have `user.email` and `user.name` configured, so `fetch_email` and `fetch_author` will return an empty string (instead of `"[your-email-here]"`).

src/amber/cli/templates/app/README.md.ecr
veelenga: Sorry, what does `--release` do in `shards build` command? I haven't seen it as a valid flag:

```sh
❯ shards --version                                                                                                                                                                     
Shards 0.7.1 (2017-10-12)
❯ shards build -h                                                                                                                                                                        
shards [options] <command>

Commands:
    build [targets] [options]
    check
    init
    install
    list
    prune
    update

Options:
    --no-color
    --version
    --production
    -v, --verbose
    -q, --quiet
    -h, --help

```

faustinoaq: > I'd rather not, that defeats the purpose of the environment variable.

@robacarp Yeah I know, that's Why I asked,

Sometimes I just think compiling amber projects for production is so repetitive

```
# Enable production env to read production.enc
AMBER_ENV=production

 # --production does not install DevDependencies
shards insttall --production

# --release enable crystal build optimizations for production
shards build --release <%= @name %>
```

Maybe we should add something like `amber build -e production`

We already have `amber run -e production` but it runs the binary by default, so I think this is more useful during development.

Of course I can do something like `AMBER_ENV=production shards build myproject` at works (unoptimized). I just thought maybe we can take advantage from `shards --production` and `crystal --release` flags.

> Builds an executable with --release flag?

@veelenga Yeah, `--release` builds executable with release flag, seems to be not documented yet

src/amber/cli/templates/app.cr
faustinoaq: Well, we already used this without major problems for all crystal projects until now, check: https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/init.cr

src/amber/cli/templates/app.cr
faustinoaq: Maybe we should change it on crystal itself too?
https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/init.cr

src/amber/cli/templates/app/README.md.ecr
robacarp: Hm, I believe `shards build --release` does check dependencies for you, no need to do `shards install` beforehand.

I'd like to avoid making the way amber uses the underlying crystal commands and libraries too opaque, because it handicaps people when they need to wander outside the bounds of an ordinary Amber application. Perhaps a command like `amber build -e production` could be more transparent by echoing out what it's doing as it happens?

src/amber/cli/templates/app/README.md.ecr
robacarp: That might live nicely in a Deployment section

src/amber/cli/templates/app/README.md.ecr
faustinoaq: > Hm, I believe shards build --release does check dependencies for you, no need to do shards install beforehand.

Yeah, you can use `shards build --production --release`

> Perhaps a command like amber build -e production could be more transparent by echoing out what it's doing as it happens?

Yeah I like that 👍

faustinoaq: > ...echoing out what it's doing as it happens?

We should do the same with `amber run` and `amber watch` using some kind of flag like `--verbose`

src/amber/cli/templates/app.cr
veelenga: Sorry, why to change it on crystal report if you propose the change here?

src/amber/cli/templates/app.cr
veelenga: I didn't say it has major problems, it just doesn't work in all cases proreply and easy to fix.

marksiemers:
Can this write the name and email to the generated `shards.yml` as well?

src/amber/cli/templates/app/README.md.ecr
faustinoaq: @robacarp I added a link to `amber encrypt` documentation 👍

src/amber/cli/templates/app/README.md.ecr
faustinoaq: Well, seems that `amber build` is useless, we already have `shards build`, currently I'm thinking on remove `amber run` too

src/amber/cli/templates/app.cr
faustinoaq: @veelenga Done! Thank you for your suggestion! 👍

src/amber/cli/templates/app.cr
faustinoaq: @veelenga Done! Thank you for your suggestion! 👍

PR #437, opened 2017-12-07
support modules of code with controller and views in the same folder

damianham:
### Description of the Change

Adds a render_module macro that renders a view from the same folder as the calling controller in addition to the standard way of separating controllers and views.

E.g. A module of code for a feature contains the following artifacts in src/modules/chart

  _chart_lib.cr  # business logic for charts
  _form.slang
 chart_controller.cr
 chart.cr  # the model
 edit.slang
 index.slang
 new.slang
 show.slang

And config/application.cr has been modified by adding
require "../src/modules/**" 

Methods in the controller call render_module("filename") instead of render("filename").
Partial rendering in views uses render_module("filename") instead of render("filename").

### Alternate Designs

No other design considered, this was the quick and easy solution.

### Benefits

Code that is organised into modules is considered to be a better way to organise code.  Everything related to a particular feature of the app is located in the same folder.

### Possible Drawbacks

Extracting the routes into a route file in the modules folder didn't work, routes still have to be created in config/routes.cr

### DIFF
--- a/src/amber/cli/commands/watch.cr
+++ b/src/amber/cli/commands/watch.cr
@@ -19,6 +19,8 @@ module Amber::CLI
       def run
         options.watch << "./config/**/*.cr"
         options.watch << "./src/views/**/*.slang"
+        options.watch << "./src/modules/**/*.cr"
+        options.watch << "./src/modules/**/*.slang"
         super
       end
     end
--- a/src/amber/controller/helpers/render.cr
+++ b/src/amber/controller/helpers/render.cr
@@ -2,6 +2,33 @@ module Amber::Controller::Helpers
   module Render
     LAYOUT = "application.slang"
 
+    macro render_module(template = nil, layout = true, partial = nil, path = "src/views", folder = __DIR__)
+      {% if !(template || partial) %}
+        raise "Template or partial required!"
+      {% end %}
+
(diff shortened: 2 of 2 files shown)


robacarp:
I think it'd be easier to reason about and allow the framework code to remain terse and simple to mount an entire amber app to a route position. This would solve your drawback about specifying routes outside of the module as well.  

I can't find it now but there has been discussion around allowing controllers to be attached to a route directly instead of one method at a time to allow for Hanami style single action per file, and I think this would fit well with that.

elorest:
This goes against the amber MVC convention, and doesn't really make sense in this project since the CLI is never going to generate models, views and controllers in the same folder.

@damianham I would suggest that you create a new shard called `amber_render_from_same_folder` or something. Which will add this macro if included in the shard file.

damianham:
I think mounting an entire app on a route position is a good idea for a collection of modules that implement a large feature e.g. an admin app, or blog, cms etc. but it is a bit of overkill for a simple feature where you will only have the model, controller, views, business logic and styling.  Creating a shard to add the macro to the controller base class is a good idea.  What would also be nice is adding a module generator to the CLI that can be used instead of scaffold to place all the artifacts in the same folder.

PR #440, opened 2017-12-09
CLI exceptions show up as readable return values instead of backtrace.

elorest:
### Description of the Change

Makes error messages look better by using CLI `exit!` command.

### DIFF
--- a/src/amber/cli/commands/database.cr
+++ b/src/amber/cli/commands/database.cr
@@ -97,6 +97,8 @@ module Amber::CLI
         ENV["DATABASE_URL"]? || begin
           CLI.settings.database_url
         end
+      rescue e
+        puts e.message 
       end
     end
   end
--- a/src/amber/cli/commands/database.cr
+++ b/src/amber/cli/commands/database.cr
@@ -23,36 +23,34 @@ module Amber::CLI
         args.commands.each do |command|
           Micrate::Cli.setup_logger
           Micrate::DB.connection_url = database_url
-          begin
-            case command
-            when "drop"
-              drop_database
-            when "create"
-              create_database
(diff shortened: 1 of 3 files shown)


src/amber/cli/commands/database.cr
```diff
@@ -89,7 +87,7 @@ module Amber::CLI
           Micrate::DB.connection_url = url.gsub(path, "/#{uri.scheme}")
           return path.gsub("/", "")
         else
-          raise "could not determine database name"
+          raise "Could not determine database name"
```
eliasjpr: Use exit here as well

src/amber/cli/commands/encrypt.cr
```diff
@@ -19,6 +19,10 @@ module Amber::CLI
         encrypted_file = "config/environments/.#{env}.enc"
         unencrypted_file = "config/environments/#{env}.yml"
 
+        unless File.exists?(unencrypted_file) || File.exists?(encrypted_file)
+          raise Exceptions::Environment.new("./config/environments/", env)
```
eliasjpr: Exit?

src/amber/cli/commands/database.cr
elorest: This is within the scope of the main `rescue` in run. Raising here is more object oriented as exiting is a way that we handle exceptions, not a replacement for exceptions.

src/amber/cli/commands/encrypt.cr
elorest: Also covered by existing rescue block. This is necessary for consistent messages. 

Again exiting isn't supposed to replace exceptions just handle them in a prettier way for the CLI.

PR #442, opened 2017-12-13
Adds environment changes to amber deploy

elorest:
### Description of the Change

1. Adds .encryption_key to deployed project so `.production.enc` can be decrypted and used.
1. Adds a flag to not use a local database in docker.
1. Adds the ability to use `environment` files to use any database or redis settings you want.
1. Adds `--environment` option so you could specify staging or something instead of production.

### Benefits

Makes amber deploy more useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,12 +170,7 @@ This will generate scaffolding for a Post:
 ## Running App Locally
 To test the generated App locally:
 
-1. Make sure your app has the right credentials via the following methods: 
-  * Add it in `config/database.yml`
-  or
-  * Run `export DATABASE_URL=postgres://[username]:[password]@localhost:5432/[your_app]_development`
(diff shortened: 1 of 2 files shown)


elorest:
Any idea why these tests are failing? They all pass locally. No changes here should have broken any tests.

faustinoaq:
Looks like CSRF is failing, maybe we are losing context on Travis, so CSRF token doesn't work.

faustinoaq:
@elorest Are you able to trigger Travis without doing a commit? I mean something like "restart build":

screenshot_20171213_023511

elorest:
Yeah I restarted the build but it still fails... Stupid error too. Doesn't happen on my Macs or linux boxes.

elorest:
Cool it passes now.

PR #443, opened 2017-12-13
Modifications to stop file descriptors from getting out of hand.

elorest:
### Description of the Change

1. Adds option for keep alive or old http behavior of connection close.
1. Sets timeout on keep-alive when selected.
1. Restarts server of too many open file descriptors make it crash.

### DIFF
--- a/src/amber/environment/settings.cr
+++ b/src/amber/environment/settings.cr
@@ -8,7 +8,6 @@ module Amber::Environment
     property database_url : String
     property host : String
     property name : String
-    property http_keep_alive : Bool
     property port : Int32
     property port_reuse : Bool
     property process_count : Int32
@@ -24,7 +23,6 @@ module Amber::Environment
       database_url: {type: String?, default: nil},
       host: {type: String, default: "localhost"},
       name: {type: String, default: "Amber_App"},
-      http_keep_alive: {type: Bool, default: false},
       port: {type: Int32, default: 3000},
       port_reuse: {type: Bool, default: true},
       process_count: {type: Int32, default: 1},
--- a/src/amber/environment/settings.cr
+++ b/src/amber/environment/settings.cr
@@ -8,6 +8,7 @@ module Amber::Environment
     property database_url : String
     property host : String
     property name : String
+    property http_keep_alive : Bool
     property port : Int32
     property port_reuse : Bool
     property process_count : Int32
@@ -23,6 +24,7 @@ module Amber::Environment
       database_url: {type: String?, default: nil},
       host: {type: String, default: "localhost"},
       name: {type: String, default: "Amber_App"},
+      http_keep_alive: {type: Bool, default: false},
       port: {type: Int32, default: 3000},
       port_reuse: {type: Bool, default: true},
       process_count: {type: Int32, default: 1},
--- a/src/amber/router/context.cr
+++ b/src/amber/router/context.cr
@@ -122,12 +122,8 @@ class HTTP::Server::Context
   end
 
   protected def finalize_response
-    if Amber.settings.http_keep_alive
-      response.headers["Connection"] = "Keep-Alive"
-      response.headers["Keep-Alive"] = "timeout=5, max=1000"
-    else
-      response.headers["Connection"] = "Close"
-    end
+    response.headers["Connection"] = "Keep-Alive"
+    response.headers.add("Keep-Alive", "timeout=5, max=300000")
     response.print(@content)
   end
 end
--- a/src/amber/router/context.cr
+++ b/src/amber/router/context.cr
@@ -123,7 +123,7 @@ class HTTP::Server::Context
 
   protected def finalize_response
     response.headers["Connection"] = "Keep-Alive"
-    response.headers.add("Keep-Alive", "timeout=5, max=300000")
+    response.headers.add("Keep-Alive", "timeout=5, max=10000")
     response.print(@content)
   end
 end
--- a/src/amber/router/context.cr
+++ b/src/amber/router/context.cr
@@ -122,6 +122,12 @@ class HTTP::Server::Context
   end
 
   protected def finalize_response
+    if Amber.settings.http_keep_alive
+      response.headers["Connection"] = "Keep-Alive"
+      response.headers["Keep-Alive"] = "timeout=5, max=1000"
+    else
+      response.headers["Connection"] = "Close"
+    end
     response.print(@content)
   end
 end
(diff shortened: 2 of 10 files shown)


src/amber/environment/settings.cr
```diff
@@ -8,6 +8,7 @@ module Amber::Environment
     property database_url : String
     property host : String
     property name : String
+    property http_keep_alive : Bool
```
eliasjpr: Do we want this as a setting? This looks strange as a setting.

src/amber/server/server.cr
```diff
@@ -67,9 +67,16 @@ module Amber
         exit
       end
 
-      logger.info "Server started in #{colorize(Amber.env, :yellow)}."
-      logger.info colorize("Startup Time #{Time.now - time}\n\n", :white)
-      server.listen(settings.port_reuse)
+      loop do
```
eliasjpr: The loop here feels odd to me since the server.listen creates something similar to listen for requests. I would suggest to just rescue the error for now  without the auto boot.

src/amber/server/server.cr
```diff
@@ -67,9 +67,16 @@ module Amber
...
+        begin
+          logger.info "Server started in #{colorize(Amber.env, :yellow)}."
+          logger.info colorize("Startup Time #{Time.now - time}\n\n", :white)
+          server.listen(settings.port_reuse)
+        rescue e
```
eliasjpr: Can we rescue a specific error here, this is rescuing all errors. This means that if there is an error for some other reason the server will keep rebooting and failing. I will suggest to remove the loop.

src/amber/environment/settings.cr
elorest: If we recognize that `keep-alive` is an header some people would want then it makes as much as really any other setting.

src/amber/server/server.cr
elorest: Rescuing the error does not good what so ever if we don't restart the server. This is the same as retry in ruby.

```rb
begin
  #do things
rescue
  retry
end
```

elorest: This is how @asterite recommended doing retry's in crystal.

elorest: @drujensen agreed that we shouldn't let DDOS's crash the server and require manual restart. This is the obviously way to do that unless we want to force everyone to configure a shell script to do it. CTRL-C still kills the server as well as killing the PID.

src/amber/server/server.cr
elorest: Rescuing a specific error makes sense. This is still a WIP as I haven't benchmarked changes yet.

elorest: Best option would probably be to leave the loop but exit/raise the error inside or rescue unless it's the errors we want to retry.

src/amber/environment/settings.cr
eliasjpr: What about instead making a PIPE for it then. I don't see this as a setting I would make a pipe and have people add or remove the pipe as they wish, this way people can use set the `keep-alive` in the pipelines they want instead globally.

src/amber/router/context.cr
```diff
@@ -122,6 +122,12 @@ class HTTP::Server::Context
   end
 
   protected def finalize_response
+    if Amber.settings.http_keep_alive
```
eliasjpr: Let's make this a `KeepAlive` pipe instead.

src/amber/server/server.cr
eliasjpr: I agree about the DDOS and Retry as long as we rescue a specific error. There is no `retry` keyword in Crystal? That's sad if it doesn't

src/amber/environment/settings.cr
elorest: Static, web etc are in different pipelines though so it would have to be enabled in both or it would cause problems. Might be the best way to do it though...

Since crystal for some reason turns on keep alive by default, a handler to disable it makes more sense, except it sort of seems like it should be off by default...

src/amber/router/context.cr
elorest: The only real issue I see with this is that keep-alive needs to be enabled globally to do any good not just in one pipeline.

elorest:
@faustinoaq 
Yes kemal does have this issue, but my kemal site isn't as popular as amber so it hasn't crashed.

Looking through the code lucky will also have this issue.

It's not really a bug, just a consequence of using keep-alive.

src/amber/router/context.cr
```diff
@@ -122,6 +122,8 @@ class HTTP::Server::Context
   end
 
   protected def finalize_response
+    response.headers["Connection"] = "Keep-Alive"
```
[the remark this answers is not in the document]
elorest: @eliasjpr Currently crystal sends back `Connection: Keep-Alive` by default regardless of request header which usually doesn't actually contain a connection header as it is a Server thing that tells the browser what it's ok with. I'm not sure why but changing it to `Connection: Close` is a huge performance hit. Slower than rails....

Changing the timeout reduces the number of File descriptors dramatically while still being fast. I believe that this is the best course of action since going from 115,000 rps to 1500 is not ok in my book.

src/amber/router/context.cr
```diff
@@ -122,6 +122,8 @@ class HTTP::Server::Context
...
+    response.headers.add("Keep-Alive", "timeout=5, max=10000")
```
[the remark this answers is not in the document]
elorest: Sure but since it's always "Keep-Alive" (see above) we don't have to waste time with a check here.

src/amber/server/server.cr
```diff
@@ -67,9 +67,17 @@ module Amber
...
-      server.listen(settings.port_reuse)
+      loop do
+        begin
+          logger.info "Server started in #{colorize(Amber.env, :yellow)}."
+          logger.info colorize("Startup Time #{Time.now - time}\n\n", :white)
+          server.listen(settings.port_reuse)
+          exit
+        rescue e : Errno
+          logger.error e.message
+          logger.info "Restarting server..."
```
robacarp: I think it would be helpful to admins if this type of event could be easily hooked with an initializer. Perhaps something like this:

```crystal
# ...
rescue e : Errno
  server_crash(e)
  logger.error e.message
  logger.info "Restarting server..."
end

def server_crash(exception : Errno)
  # noop
end
```

elorest: @robacarp Feel free to submit a PR on that. I'm super swamped right now and want to get this merged in since it's causing issues.

PR #444, opened 2017-12-13
Emergency Bug Fix: Fixed order dependent CSRF spec.

elorest:
### Description of the Change

CSRF tests randomly fail depending on the system they're ran on. 
This is due to different file systems loading files in different orders.
Fixed by loading settings from env file again at the beginning of crsf_spec.

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -12,7 +12,7 @@ CURRENT_DIR       = Dir.current
 
 Amber.environment_path = "./spec/support/config"
 Amber.env=(ENV[Amber::AMBER_ENV])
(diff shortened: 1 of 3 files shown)


spec/spec_helper.cr
```diff
@@ -12,7 +12,7 @@ CURRENT_DIR       = Dir.current
 
 Amber.environment_path = "./spec/support/config"
 Amber.env=(ENV[Amber::AMBER_ENV])
-Amber.settings.redis_url = ENV["REDIS_URL"]? || Amber.settings.redis_url
+Amber.settings.redis_url = ENV["REDIS_URL"] if ENV["REDIS_URL"]?
```
eliasjpr: 👍

PR #445, opened 2017-12-13
[WIP][Logger] Improves logger

eliasjpr:
#### Addresses Issue

Issue: https://github.com/amberframework/amber/issues/346

#### Description of the Change

Currently Amber outputs very little information of the request in the terminal
console. Would be useful to show query data on logs too, just like other frameworks
Also adding time and aligned/tabulated logs would be useful and more readable.

####  Changes:

- Makes logs more readable
- Adds Request Session Header Cookies information to the logger
- Filters sensitive data from logs by providing a filtered list of keys
- Skips a list of keys to be printed in the console
- Uses Settings logger in CLI and Amber
- Uses Amber.settings.colorize_logging setting for logs, no need to toggle anymore
- Updates files to use the logger for printing to the output
- Cleans/Refactor Sentry Process Run
- When in `Severity::DEBUG` does not print date time information

#### Usage:

```crystal
Amber.settings.logger.puts "Message you want to log", "Section", :blue
```
**Configuring the logger**
 
```crystal
Amber.configure do |setting|
  pipeline :web do
    plug Amber::Pipe::Logger.new(
       # Filters sensitive information from being log 
       filtered: ["password", "ssn", "email"],
       # Skips keys from being log 
       reject: ["somekey"],
       # Only log for development environment
     ) if Amber.env.development?
  end
end
```

### Benefits

- More readable and useful logs
- Makes logger more configurable
- Adds useful information to the logger
- Makes it easy to read logs

### Possible Drawbacks

- Requests speed decreases because of the logging.

#### Example Screenshots

screen shot 2017-12-13 at 4 33 42 pm

screen shot 2017-12-12 at 4 53 53 pm

### DIFF
--- a/src/amber/cli/helpers/process_runner.cr
+++ b/src/amber/cli/helpers/process_runner.cr
@@ -1,0 +1,114 @@
+require "sentry"
+
+module Sentry
+  class ProcessRunner
+    property processes = [] of Process
+    property process_name : String
+    property should_build = true
+    property files = [] of String
+    @logger : Amber::Environment::Logger
+
+    def initialize(
+                   @process_name : String,
+                   @build_command : String,
+                   @run_command : String,
+                   @build_args : Array(String) = [] of String,
+                   @run_args : Array(String) = [] of String,
+                   files = [] of String,
+                   should_build = true,
+                   @logger = Amber::CLI.logger)
+      @files = files
+      @should_build = should_build
+      @should_kill = false
+      @app_built = false
+    end
+
+    def run
+      loop do
+        if @should_kill
+          break
+        end
+        scan_files
+        sleep 1
+      end
+    end
+
+    # Compiles and starts the application
+    #
+    def start_app
+      stop_all_processes
+      start_all_processes
+    end
+
+    private def scan_files
+      file_changed = false
+
+      Dir.glob(files) do |file|
+        timestamp = get_timestamp(file)
+        if FILE_TIMESTAMPS[file]? && FILE_TIMESTAMPS[file] != timestamp
+          FILE_TIMESTAMPS[file] = timestamp
+          file_changed = true
+          log "#{file.capitalize.colorize(:light_gray)}"
+        elsif FILE_TIMESTAMPS[file]?.nil?
+          FILE_TIMESTAMPS[file] = timestamp
+          file_changed = true
+          log "Watching file: #{file.capitalize.colorize(:light_gray)}"
+        end
+      end
+
+      start_app if (file_changed)
+    end
+
+    private def start_all_processes
+      log "Compiling #{project_name}..."
+      create_all_processes
+    end
+
+    private def stop_all_processes
+      log "Terminating app #{project_name}..."
+      @processes.each do |process|
+        process.kill unless process.terminated?
+      end
+      processes.clear
+    end
+
+    private def create_all_processes
+      build_app_process
+      @processes << create_watch_process
+      sleep 3
+      create_npm_process
+    end
+
+    private def build_app_process
+      log "Building project #{project_name}..."
+      Process.run(@build_command, shell: true, output: true, error: true)
+    end
+
+    private def create_watch_process
+      log "Starting #{project_name}..."
+      Process.new(@run_command, output: true, error: true)
+    end
+
+    private def create_npm_process
+      node_log "Installing dependencies..."
+      Process.new("npm install && npm run watch", output: false, error: true, shell: true)
+      node_log "Watching public directory"
+    end
+
+    private def get_timestamp(file : String)
+      File.stat(file).mtime.to_s("%Y%m%d%H%M%S")
+    end
+
+    private def project_name
+      process_name.capitalize.colorize(:white)
+    end
+
+    private def log(msg)
+      @logger.puts msg, "Watcher", :light_gray
+    end
+
+    private def node_log(msg)
+      @logger.puts msg, "Node", :dark_gray
+    end
+  end
+end
--- a/src/amber/cli/helpers/process_runner.cr
+++ b/src/amber/cli/helpers/process_runner.cr
@@ -4,7 +4,6 @@ module Sentry
   class ProcessRunner
     property processes = [] of Process
     property process_name : String
-    property should_build = true
     property files = [] of String
     @logger : Amber::Environment::Logger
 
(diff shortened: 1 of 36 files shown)


eliasjpr:
Fixing specs...

src/amber/cli/helpers/process_runner.cr
```diff
@@ -0,0 +1,114 @@
...
+      @should_build = should_build
+      @should_kill = false
+      @app_built = false
+    end
+
+    def run
+      loop do
+        if @should_kill
+          break
+        end
```
marksiemers: This looks like a good place for a post condition:
```
break if @should_kill
```

src/amber/router/pipe/logger.cr
```diff
@@ -3,26 +3,48 @@ require "colorize"
...
+        log "Time Elapsed: #{elapsed.colorize(color)}", "Request", color
       end
 
-      def path(context)
-        "\"" + colorize(context.request.path.to_s, :yellow) + "\" "
+      def log_other(other, name, color = :light_cyan)
+        other.to_h.each do |key, val|
+          next if @reject.includes? key
+          if @filtered.includes? key.to_s
+            log "#{key}: #{"FILTERED".colorize(:white).mode(:underline)}", name, color
```
marksiemers:
```crystal
private def filtered_param_log_text(key)
  "#{key}: #{"FILTERED".colorize(:white).mode(:underline)}"
end

private def param_log_text(key,val)
  "#{key}: #{val.colorize(color)}"
end

log filtered_param_log_text(key), name, color
...
log param_log_text(key,val), name, color
```

eliasjpr:
@marksiemers 

I guess we can pass a hash of options something like:

```crystal
options = {
  development: ["request", "headers", "cookies"]
}
```

How would suggest handling it?

src/amber/environment/settings.cr
```diff
@@ -1,9 +1,8 @@
-require "logger"
 require "yaml"
 
 module Amber::Environment
   class Settings
-    property logger : ::Logger = Logger.new(STDOUT)
+    property logger : Logger = Environment::Logger.new(STDOUT)
```
elorest: Weren't we going to remove this from settings since we don't have a way to configure it through env.ymls anymore?

src/amber/server/server.cr
```diff
@@ -56,28 +56,33 @@ module Amber
...
       Signal::INT.trap do
-        logger.info "Shutting down Amber"
+        log "Shutting down Amber"
         server.close
         exit
       end
 
-      logger.info "Server started in #{colorize(Amber.env, :yellow)}."
-      logger.info colorize("Startup Time #{Time.now - time}\n\n", :white)
+      log "Started in #{Amber.env.colorize(:yellow)}"
```
elorest: So what happened to logger.info and logger.error etc?

src/amber/environment/settings.cr
eliasjpr: I didn't remove since people can still change it via `config/application.cr`, we can reconsider in a separate PR.

src/amber/server/server.cr
eliasjpr: You can still use it, the `info`, and other severity can still be used. If you look at this https://github.com/amberframework/amber/pull/445/files#diff-7db448dcdac4549398cfd26fa10c9d55R20 it will show the severity when is not DEBUG

marksiemers:
> How would suggest handling it?

If it is possible to do in an optional initializer, or in `application.cr` and the applicable environment config files.

It looks like the concept of log levels is built in, but I think this may be more detailed than that. We may need to decide on common terminology, where there are "levels" (INFO, WARN, DEBUG, etc.), and "items" (request, headers, cookies, session, etc.)

One possible example for an initializer:
```crystal
Amber.settings.logger.configure do |config|
  config.development.items = ["request", "headers", "cookies"]
  config.test.items << "headers"
  config.production.logger.level = Amber::Logger::INFO
end
```
If it is done in `config/application.cr` then we could have options in the yaml file(s):
```yaml
# config/environments/development.yml
port: 3000
name: microsecond-blog
log: "::Logger.new(STDOUT)"
log_level: "::Logger::INFO"
log_items: ["request", "headers", "cookies"]
color: true
...
```

@amberframework/core-team - WDYT?
- Should we provide this kind of configuration?
- Should it be done in an initializer or in `config/application.cr` and the environment yml files?

src/amber/environment/settings.cr
elorest: Ok.

elorest: We do probably need to think of some way to at least configure log level via env.yml's. Since people probably want different things logged in different environments.

elorest: It might be good to do it in this PR since it relates to logging and configuring logging level. This next release will contain all of our changes to settings to it would make sense to include that unless it's a huge rework which it doesn't seem like.

src/amber/server/server.cr
elorest: So is `log` by default just info? It looks as if all referrences to logger.info and error have been replaced with just log?

src/amber/environment/logger.cr
```diff
@@ -0,0 +1,10 @@
+require "logger"
+require "colorize"
+
+module Amber::Environment
+  class Logger < ::Logger
+    def puts(message, program = "Amber", color = :magenta)
+      log(self.level, message, "#{program} |".colorize(color))
```
faustinoaq: @eliasjpr Can we be able to disable colored output? maybe using `--no-color` flag already implemented on #272

eliasjpr: I will take a look at this, but I a mostly certain this can be done

eliasjpr:
@elorest I agree that the logger should have a way to point where to store the logs files vs console. 

The purpose of this was to improve the logger to the needs specified in the issue. It would be good to follow up with another PR to continue improving the logger. 

Can we add that as an issue "Allow logger output to be store in a log file"?

sam0x17:
I really like the idea of this being highly configurable. For my purposes I would just want the params, the host, and the user agent. It's crazy how much data rails logs can take up on a large app, especially when there are data retention requirements (e.g. financial application).

eliasjpr:
Closing this will reopen

Issue #379, opened 2017-11-10
Filter security / credential parameters from log files


fridgerator:
### Description

Filter certain parameters from log files: `password`, `password_confirmation`, credit card information, etc

### Additional Information

Exposing passwords to log files is a security risk.  Other frameworks filter certain parameters in the logs.

Rails uses `Rails.application.config.filter_parameters` array

faustinoaq:
Yeah, An issue to take in account on #346

fridgerator:
IMO this should be prioritized, this is as bad as storing passwords in a database as plain text.

eliasjpr:
Addressed in https://github.com/amberframework/amber/pull/450

PR #450, opened 2017-12-18
[Log] Better logs

eliasjpr:
#### Addresses Issue

Issue: 
https://github.com/amberframework/amber/issues/346
https://github.com/amberframework/amber/issues/379

> Note: A separate PR for the CLI logging will branch out of this PR to keep this PR small 

#### Description of the Change

Currently Amber outputs very little information of the request in the terminal
console. Would be useful to show query data on logs too, just like other frameworks
Also adding time and aligned/tabulated logs would be useful and more readable.

####  Changes:

- Makes logs more readable
- Adds Request Session Header Cookies information to the logger
- Filters sensitive data from logs by providing a filtered list of keys
- Skips a list of keys to be printed in the console
- Uses Settings logger in CLI and Amber
- Uses Amber.settings.colorize_logging setting for logs, no need to toggle anymore
- Updates files to use the logger for printing to the output
- Cleans/Refactor Sentry Process Run
- When in `Severity::DEBUG` does not print date time information

#### Usage:

```yaml
# environment yamls new logging settings 
logging:
  severity: debug
  color: true
  time: false
  level: false
  filter:
    - password
    - confirm_password
  skip:
    -
  context:
    - request
    - session
    - headers
    - cookies
    - params
```

```crystal
Amber.settings.logger.puts "Message you want to log", "Section", :blue
```
**Configuring the logger**

The **Amber::Pipe::Logger** will default to the environment settings values and it can be overwritten as follows:

```crystal
Amber.configure do |setting|
  pipeline :web do
    plug Amber::Pipe::Logger.new(
       # Filters sensitive information from being log 
       filtered: ["password", "ssn", "email"],
       # Skips keys from being log 
       skip: ["somekey"],
       # context allows to specify which sections to output
       context: ["request", "headers"]
     )
  end
end
```

### Benefits

- More readable and useful logs
- Makes logger more configurable
- Adds useful information to the logger
- Makes it easy to read logs

### Possible Drawbacks

- Requests speed decreases because of the logging.

#### Example Screenshots

screen shot 2017-12-13 at 4 33 42 pm

screen shot 2017-12-12 at 4 53 53 pm

### DIFF
--- a/spec/support/config/development.yml
+++ b/spec/support/config/development.yml
@@ -1,4 +1,8 @@
-colorize_logging: true
+logging:
+  color: true
+  time: false
+  level: false
+
 database_url: "mysql://root@localhost:3306/development_settings_test"
 host: "localhost"
 name: development_settings
--- a/spec/support/config/development.yml
+++ b/spec/support/config/development.yml
@@ -2,7 +2,7 @@ logging:
   color: true
   time: false
   level: false
-
+  severity: "info"
 database_url: "mysql://root@localhost:3306/development_settings_test"
 host: "localhost"
 name: development_settings
--- a/spec/support/config/development.yml
+++ b/spec/support/config/development.yml
@@ -1,8 +1,18 @@
 logging:
-  color: true
-  time: false
-  level: false
-  severity: "info"
+  severity: debug
+  colorize: true
+  filter:
+    - password
+    - confirm_password
+  skip:
+    -
+  context:
+    - request
+    - session
+    - headers
+    - cookies
+    - params
+
 database_url: "mysql://root@localhost:3306/development_settings_test"
 host: "localhost"
 name: development_settings
--- a/src/amber/cli/templates/app/config/application.cr.ecr
+++ b/src/amber/cli/templates/app/config/application.cr.ecr
@@ -30,9 +30,12 @@ Amber::Server.configure do |settings|
   #
   #
   # Colorize Logging: specifies whether or not to use ANSI color codes
-  # when logging information. Defaults to true.
+  # when logging information, display the time and/or to display the severity level.
+  # Defaults to true.
   #
-  # settings.colorize_logging = true
+  # settings.logging["color"] = true
+  # settings.logging["time"] = false
+  # settings.logging["level"] = false
   #
   #
   # Secret Key Base: is used for specifying a key which allows sessions
@@ -121,5 +124,5 @@ Amber::Server.configure do |settings|
   # Log Level defines the verbosity of the Amber logger. This option defaults to
   # :debug for all environments. The available log levels are: :debug, :info, :warn,
   # :error, :fatal, and :unknown.
-  settings.logger.level        = ::Logger::DEBUG
+  settings.logger.level = ::Logger::DEBUG
 end
--- a/src/amber/cli/templates/app/config/application.cr.ecr
+++ b/src/amber/cli/templates/app/config/application.cr.ecr
@@ -33,9 +33,14 @@ Amber::Server.configure do |settings|
   # when logging information, display the time and/or to display the severity level.
   # Defaults to true.
   #
-  # settings.logging["color"] = true
-  # settings.logging["time"] = false
-  # settings.logging["level"] = false
+  # Log Level defines the verbosity of the Amber logger. This option defaults to
+  # debug for all environments. The available log levels are: debug, info, warn,
+  # error, fatal, and unknown.
+  #
+  # settings.logging.color = true
+  # settings.logging.time = false
+  # settings.logging.level = false
+  # settings.logging.severity = debug
   #
   #
   # Secret Key Base: is used for specifying a key which allows sessions
@@ -121,8 +126,4 @@ Amber::Server.configure do |settings|
   # settings.logger = ::Logger.new(STDOUT)
   #
   #
-  # Log Level defines the verbosity of the Amber logger. This option defaults to
-  # :debug for all environments. The available log levels are: :debug, :info, :warn,
-  # :error, :fatal, and :unknown.
-  settings.logger.level = ::Logger::DEBUG
 end
--- a/src/amber/cli/templates/app/config/application.cr.ecr
+++ b/src/amber/cli/templates/app/config/application.cr.ecr
@@ -37,10 +37,11 @@ Amber::Server.configure do |settings|
   # debug for all environments. The available log levels are: debug, info, warn,
   # error, fatal, and unknown.
   #
-  # settings.logging.color = true
-  # settings.logging.time = false
-  # settings.logging.level = false
+  # settings.logging.colorize = true
   # settings.logging.severity = debug
+  # settings.logging.filter = %w(password confirm_password)
+  # settings.logging.skip = %w()
+  # settings.logging.context = %w(request headers cookies session params)
   #
   #
   # Secret Key Base: is used for specifying a key which allows sessions
--- a/src/amber/cli/templates/scaffold/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
+++ b/src/amber/cli/templates/scaffold/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
@@ -28,7 +28,6 @@ class <%= class_name %>ControllerTest < GarnetSpec::Controller::Test
     @handler = Amber::Pipe::Pipeline.new
     @handler.build :web do
       plug Amber::Pipe::Error.new
-      plug Amber::Pipe::Logger.new
       plug Amber::Pipe::Session.new
       plug Amber::Pipe::Flash.new
     end
--- a/src/amber/environment/loader.cr
+++ b/src/amber/environment/loader.cr
@@ -2,14 +2,11 @@ module Amber::Environment
   class Loader
     def initialize(@environment : Amber::Environment::EnvType = Amber.env.to_s,
                    @path : String = Amber.environemnt_path)
+      raise Exceptions::Environment.new(@path, @environment) unless settings_file_exist?
     end
 
     def settings
-      if content = settings_content
-        Settings.from_yaml(content.to_s)
-      else
-        Settings.from_yaml("default: settings")
-      end
+      Settings.from_yaml(settings_content.to_s)
     end
 
     private def settings_content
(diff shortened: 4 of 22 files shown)


robacarp:
This is a great improvement. I have two questions regarding similar features:

- Given that Granite or some other ORM is being used, how can those queries be logged as well? Logs like you have are great for debugging the http side of an application, but not the database interactions. 
- These logs are quite verbose, but the options flagging is nice. In production, it can be very helpful to keep each request to a single log line. It makes it easy to pass off those logs to an ELK stack, CloudWatch, etc. Is there any thoughts for a formatter that can render logs in that way?

eliasjpr:
@robacarp you can still change the logger formatter by simply calling `Amber.logger.formatter` as is it explained here https://github.com/crystal-lang/crystal/blob/e2a1389e8165fb097c785524337d7bb7b550a086/src/logger.cr#L59

src/amber/environment/loader.cr
```diff
@@ -2,11 +2,14 @@ module Amber::Environment
   class Loader
     def initialize(@environment : Amber::Environment::EnvType = Amber.env.to_s,
                    @path : String = Amber.environemnt_path)
-      raise Exceptions::Environment.new(@path, @environment) unless settings_file_exist?
     end
 
     def settings
-      Settings.from_yaml(settings_content.to_s)
+      if content = settings_content
```
elorest: Do we really want to load default values if they don't have env files?

elorest: Do we really want to load default values if they don't have env files?

src/amber/environment/logger.cr
```diff
@@ -0,0 +1,16 @@
+require "logger"
+require "colorize"
+
```
elorest: I like this.

src/amber/environment/settings.cr
```diff
@@ -1,10 +1,18 @@
-require "logger"
+require "./logger"
 require "yaml"
 
 module Amber::Environment
   class Settings
-    property logger : ::Logger = Logger.new(STDOUT)
-    property colorize_logging : Bool
+    alias LoggingType = NamedTuple(severity: String,
```
elorest: Could this be cleaned up a bit?

src/amber/server/server.cr
```diff
@@ -67,17 +67,17 @@ module Amber
         exit
       end
 
-      logger.info "Server started in #{colorize(Amber.env, :yellow)}."
-      logger.info colorize("Startup Time #{Time.now - time}\n\n", :white)
+      logger.info "Server started in #{Amber.env.colorize(:yellow)}."
```
elorest: Like that we're using `logger.info` again.

src/amber/environment/loader.cr
eliasjpr: We have to load default because the CLI will not be able to run `amber new {project_name}`

src/amber/environment/settings.cr
eliasjpr: We can break the subclasses into its own file. Or do you mean clean the **LoggingType**?

elorest: I meant with logging type.

It would be more readable and accomplish the same thing to just set logger as a default named tuple.
`property logging = {severity: "info, color: true, time: false, level: false, etc}`

Then all the values would still have to match that.

src/amber/environment/settings.cr
```diff
@@ -13,13 +21,21 @@ module Amber::Environment
...
     property ssl_cert_file : String
+    property logger : Logger?
 
     YAML.mapping(
-      colorize_logging: {type: Bool, default: true},
+      logging: {type: LoggingType, default: {
+        severity: "info",
+        color:    true,
+        time:     false,
+        level:    false,
```
elorest: Why is level a boolean?

eliasjpr: this is to enable printing the output string `(Debug)`, `(INFO)`. Severity if to assign the type.

src/amber/environment/settings.cr
eliasjpr: I did it with the logging type with the thought that we do not want to default at the property level and only default within the YAML

src/amber/environment/settings.cr
marksiemers: Semantically, `level` does seem strange for this. If I understand correctly, this is used to determine whether or not the `severity` is output by the logger, correct?

It may be a little long, but `output_severity` might work better.

What is common in other frameworks using `severity` vs `level` for logging?

src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -30,9 +30,20 @@ Amber::Server.configure do |settings|
...
   #
   # Colorize Logging: specifies whether or not to use ANSI color codes
-  # when logging information. Defaults to true.
+  # when logging information, display the time and/or to display the severity level.
+  # Defaults to true.
   #
-  # settings.colorize_logging = true
+  # Log Level defines the verbosity of the Amber logger. This option defaults to
+  # debug for all environments. The available log levels are: debug, info, warn,
+  # error, fatal, and unknown.
```
marksiemers: In this comment, it seems that it is talking about `severity` and not `level` (as coded).

Do we want `severity` to be called `level` and `level` to be called `output_log_level`?

src/amber/cli/templates/scaffold/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
```diff
@@ -28,7 +28,6 @@ class <%= class_name %>ControllerTest < GarnetSpec::Controller::Test
     @handler = Amber::Pipe::Pipeline.new
     @handler.build :web do
       plug Amber::Pipe::Error.new
-      plug Amber::Pipe::Logger.new
```
marksiemers: The idea here is that the logger is always there now, you can't remove it, correct?

src/amber/environment/settings.cr
```diff
@@ -52,5 +76,73 @@ module Amber::Environment
...
+    def initialize(io, logging)
+      Colorize.enabled = logging.color
+      @logger = Environment::Logger.new(io)
+      @logger.level = logging.severity
+      @logger.progname = "Server"
+      @logger.formatter = Logger::Formatter.new do |severity, datetime, progname, message, io|
+        if logging.time == true
+          io << datetime.to_s("%Y-%m-%d %I:%M:%S")
+          io << " "
+        end
```
marksiemers: Is there an advantage or a reason to have it this way instead of:
```crystal 
io << datetime.to_s("%Y-%m-%d %I:%M:%S") << " " if logging.time == true
# If a newline is needed, could this work:
io << datetime.to_s("%Y-%m-%d %I:%M:%S") << "\n" if logging.time == true
io << severity << "\n" if logging.level == true
...
```

src/amber/cli/templates/app/config/application.cr.ecr
eliasjpr: I thought of prefixing the boolean fields with `show_` so it would read

show_color
show_time
show_level
level
skip
context
filter

src/amber/cli/templates/scaffold/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
eliasjpr: We can remove it. I don't see the need for the logger to be on the pipeline for a test.

src/amber/environment/settings.cr
eliasjpr: We can prefix the booloan with `show_` which dictates some intention

src/amber/environment/settings.cr
eliasjpr: I find the above easier to read and is practically the same

src/amber/environment/loader.cr
elorest: That is a limitation added by the CLI logging PR. It doesn't really seem like the CLI needs a configurable logger though. Is this the best way to go about it?

src/amber/environment/settings.cr
elorest: I agree. It's just that defining a NamedTuple type looks even uglier. I'm fine with it though.

src/amber/cli/templates/app/config/application.cr.ecr
robacarp: Since level is an abstraction that indicates how severities relate to each other for the purpose of stream filtering, I think a log _message_ has a severity and a log stream has a severity level filter, aka `log.level`.

Choosing what to print for a given log message would be `show_severity` and choosing which logs to filter out would be `level`.

Rails config doesn't nest logger settings in a logging namespace, but the parallel is here:

> config.log_level defines the verbosity of the Rails logger.

As far as I can tell, Rails doesn't provide the configuration option to prevent display of  timestamps and severity without implementing your own log formatter class, though they do provide an option called `config.colorize_logging`. 

Since there is a format string option, is providing the boolean flags `show_time` and `show_level` or `show_severity` redundant?

src/amber/cli/templates/scaffold/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
elorest: Question: Can logger be turned off for benchmarks, and really fast response times?

spec/support/config/development.yml
```diff
@@ -1,4 +1,18 @@
-colorize_logging: true
+logging:
+  severity: debug
```
elorest: Should we perhaps put a comment here about other severity levels?

src/amber/environment/settings.cr
```diff
@@ -1,10 +1,18 @@
...
+require "./logger"
 require "yaml"
 
 module Amber::Environment
   class Settings
-    property logger : ::Logger = Logger.new(STDOUT)
-    property colorize_logging : Bool
+    alias LoggingType = NamedTuple(
+      severity: String,
+      color: Bool,
```
elorest: Color should probably be "show_color" since `color` seems like the value should be `blue` or something instead of true/false.

faustinoaq: Maybe not `color` nor `show_color` but `colorize` ? 😅

spec/support/config/development.yml
eliasjpr: There is information about the severity levels in **config/application.cr**

elorest:
So if logger isn't a pipe anymore how would we completely disable it to avoid the overhead?

eliasjpr:
The Pipe::Logger is to log requests. Anywhere else where `Amber.logger` is called will add some overhead. 

A way to disable the logger will be to set `Amber.settings.logger=nil`

Issue #347, opened 2017-10-31
Allow hot code reloading using Amber tasks


faustinoaq:
### Description

Currently amber don't support reloading front-end changes out-the-box

We need to execute `amber watch` and then in other terminal `npm run watch`. However, only reload back-end changes, we still needs hit reload button in the browser.

Two alternatives:

1. Use `webpack-dev-server` by default, using port forwarding and handling both, back-end and front-end using `Amber::Tasks` for watching changes.

**Advantages:** Use webpack, used by front-end developers, no changes in workflow.
**Disadvantages:** Bind Amber to webpack ignoring other technologies (rollup, browserify,brunch, etc), also need to download even more npm depencencies 100MB+

2. Using Amber websockets to watch changes in backend and reload the browser, see [my experiment](https://gist.github.com/faustinoaq/2fd199a0cd92c824ee2f7d62635e73ab)

**Advantages:** Don't requires new npm dependencies, and uses amber framework capabilities.
**Disadvantages:** Hard to implement ? without modifying websocket pipeline 

### Steps to Reproduce

1. Create a new amber app run `shards install` and `npm install`
2. Run `amber watch`
3. Open new terminal and run `npm run  watch`

**Expected behavior:** Browser is reloaded automatically on changes

**Actual behavior:** Only backend is recompiled, browser still needs to hit reload button

**Reproduces how often:** 100%

### Versions

100%

### Additional Information

![Amber client reloading](http://i.imgur.com/JCR10m4.png)

@amberframework/contributors WDYT?

marksiemers:
@faustinoaq - It looks like you've already done most of the work.

For the socket pipeline, I think it'll just take a macro and checking the environment that you are in "development"

At first glance, I vote for option 2.

robacarp:
anything that simplifies the development cycle is a win in my book

eliasjpr:
@faustinoaq I think you should go ahead and implement this with Amber Tasks. We can later cleanup the console log output.

faustinoaq:
@eliasjpr Yeah, I will a PR on next weeks 👍

faustinoaq:
Client reloading without npm dependencies: https://github.com/tapio/live-server/blob/master/injected.html

Proposed flag:

```
amber watch --hot-reload
```

robacarp:
If the command is clever enough to not blow up when assets don’t exist, just make it the default?

faustinoaq:
See #457

eliasjpr:
👍

PR #451, opened 2017-12-19
[CLI] Adds consistent logging to CLI

eliasjpr:
Adds consistent logging for CLI

issue:

https://github.com/amberframework/amber/issues/347

### DIFF
--- a/src/amber/cli/commands/database.cr
+++ b/src/amber/cli/commands/database.cr
@@ -4,14 +4,19 @@ require "mysql"
 require "sqlite3"
 
 module Amber::CLI
+  CLI.logger.progname = "Database"
+  Micrate.logger = settings.logger
+  Micrate.logger.progname = "Database"
+
   class MainCommand < ::Cli::Supercommand
     command "db", aliased: "database"
 
-    class Database < ::Cli::Command
+    class Database < Command
       command_name "database"
 
       class Options
         arg_array "commands", desc: "drop create migrate rollback redo status version seed"
+        bool "--no-color", desc: "# Disable colored output", default: false
         help
       end
 
@@ -20,17 +25,17 @@ module Amber::CLI
       end
 
       def run
+        CLI.toggle_colors(options.no_color?)
         args.commands.each do |command|
-          Micrate::Cli.setup_logger
           Micrate::DB.connection_url = database_url
           case command
           when "drop"
-            drop_database
+            Micrate.logger.info drop_database
           when "create"
-            create_database
+            Micrate.logger.info create_database
           when "seed"
             `crystal db/seeds.cr`
-            puts "Seeded database"
+            Micrate.logger.info "Seeded database"
           when "migrate"
             Micrate::Cli.run_up
           when "rollback"
@@ -48,7 +53,7 @@ module Amber::CLI
       rescue e : Micrate::UnorderedMigrationsException
         exit! Micrate::Cli.report_unordered_migrations(e.versions), error: true
       rescue e : DB::ConnectionRefused
-        exit! "Connection refused: #{Micrate::DB.connection_url}", error: true
+        exit! "Connection refused: #{Micrate::DB.connection_url.colorize(:light_blue)}", error: true
       rescue e : Exception
         exit! e.message, error: true
       end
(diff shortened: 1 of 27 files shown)


src/amber/cli/commands/database.cr
```diff
@@ -87,14 +92,16 @@ module Amber::CLI
...
       end
 
       private def database_url
-        ENV["DATABASE_URL"]? || begin
-          CLI.settings.database_url
-        end
+        ENV["DATABASE_URL"]? || CLI.settings.database_url
+      end
+
+      private def puts(msg)
```
elorest: This method seems redundant.

src/amber/cli/commands/deploy.cr
```diff
@@ -147,15 +148,23 @@ module Amber::CLI
...
-      def update_project
+      private def update_project
         remote_cmd(%Q("cd amberproject && git pull"))
       end
+
+      private def server_name
+        @server_name.colorize(:green)
+      end
+
+      private def puts(msg)
```
elorest: I don't think this should be defined for every command. Could you make it a protected method `::CLI::Command`

src/amber/cli/helpers/process_runner.cr
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    private def create_watch_process
+      log "Starting #{project_name}..."
+      Process.new(@run_command, output: true, error: true)
+    end
+
+    private def create_npm_process
+      node_log "Installing dependencies..."
+      Process.new("npm install && npm run watch", output: false, error: true, shell: true)
```
drujensen: What do you think of allowing this to be configurable?  A Procfile that specifies what should be watched and ran when files change?
```
app:
  watch: ./src/**
  run: crystal run ./src/app.cr
web:
  watch: ./src/assets/**
  run: npm run build
sidekiq:
  watch ./src/jobs/**
  run: crystal ./src/sidekiq.cr
```

drujensen: Also, What if someone doesn't want to use npm?  I would prefer this not being a dependency in the amber watch command.

marksiemers: @drujensen - @faustinoaq was working on a crystal/amber solution that used WebSockets, but I'm not sure the status of it. When that is done, it will be easier to pull the npm dependency.

drujensen: @marksiemers not sure what kind of solution would use WebSockets unless you're talking about hot reloading?

In either case, I don't think we should hardcode an `npm` command in the CLI.  We should be agnostic to the users choice of front-end solution.  If someone is just creating a backend api or if they want to use a different solution, we shouldn't force them to install node.

elorest: https://github.com/oltmannsdaniel/frontender

We could still do something like that and remove npm dependency entirely.

faustinoaq: > was working on a crystal/amber solution that used WebSockets, but I'm not sure the status of it.

@marksiemers The WebSocket stuff is a bit different, the idea is doing client reloading using a browser script and check file changes trough an `Amber::Pipe`. Compilation stuff using command line would need NPM though. I was a bit busy with my parents during latest weeks, so I haven't done much advance on it 😅

eliasjpr: @drujensen I like the idea of having a profile, maybe this can just live in the `.amber.yml`, but there seems to be a lot of configurations, app, logging, and now the watcher. I would like for for us

What about for now just add a `flag`? For example:

```
amber w --front-end
```

Until we can come up with a better solution.

drujensen: @eliasjpr I prefer we take this out until we come up with a proper solution.  Adding a flag now and then possibly removing it would be jarring to users.

Also, I was surprised to see this in a PR about logging.  It's not really related to logging so I think it should be in a separate PR anyway.

src/amber/cli/helpers/process_runner.cr
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    private def build_app_process
+      log "Building project #{project_name}..."
+      Process.run(@build_command, shell: true, output: true, error: true)
+    end
+
+    private def create_watch_process
+      log "Starting #{project_name}..."
+      Process.new(@run_command, output: true, error: true)
```
drujensen: Just FYI, looks like the parameters changed in 0.24.1 based on Brian Cardiff's PR.  Not sure which will be merged first:
```
output: Process::Redirect::Inherit, error: Process::Redirect::Inherit
```

src/amber/cli/helpers/process_runner.cr
elorest: I think that would be a good idea. This isn't a Tax bill...

elorest: _We do need to be careful about keeping PR's on point, and making the minimum amount of changes required for the stated feature_**.

For now though I wonder if we should just release npm compile as part of watch, but not document it. We can always change it later? Or remove it and create a new PR where we can discuss that.

elorest: I just want to get this released before Christmas.
There are a lot of differences between 0.3.6 and master. It's hard to document right now because it's different depending on master or 0.3.6.

Issue #435, opened 2017-12-06
Remove amber run


faustinoaq:
### Description

We already have other commands for running crystal apps:

1. `crystal run src/app.cr` to run main file (development)
2. `shards build app` to install and build the project (production/development)

Keeping `amber watch` for detecting changes, recompiling and restarting the server during development

### Steps to Reproduce

1. `amber run`
2. Compare to `crystal run`
3. Very similar

**Expected behavior:** `amber run` is useless, let's use `crystal run`

**Actual behavior:** `amber run` exists.

**Reproduces how often:** 100%

### Versions

0.3.6

### Additional Information

I don't recommend to use `amber run` on production, best use `shards build` and then execute the binary.

faustinoaq:
@amberframework/contributors

eliasjpr:
Addressed with #452

PR #452, opened 2017-12-21
[CLI] Removes Run and Perform Commands

eliasjpr:
This removes the Run and Perform command in favor of existing commands
that perform the same tasks.

- Run is replaced by using `crystal build` or `shards build` these
commands builds the project and later the project can be ran with
`./bin/{project_name}`
- Perform is replaced in favor of the exec command that can execute
existing files.
- Adds header to commands for better help display.

Issues:

https://github.com/amberframework/amber/issues/435


src/amber/cli/commands/database.cr
```diff
@@ -71,7 +71,9 @@ module Amber::CLI
       private def create_database
         url = Micrate::DB.connection_url.to_s
         if url.starts_with? "sqlite3:"
-          puts "For sqlite3, the database will be created during the first migration."
+          puts <<-MSG
+          No migration files were found. For sqlite3, the database will be created during the first migration.
```
faustinoaq: 👍

PR #453, opened 2017-12-22
Add support for crystal v0.24.1

faustinoaq:
### Description of the Change

Add support for crystal for v0.24.1

### Alternate Designs

Remove compatibility with v0.23.1

### Benefits

I think v0.23.1 is still relevant

### Possible Drawbacks

No

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM drujensen/crystal:0.23.1
+FROM drujensen/crystal:0.24.1
 
 WORKDIR /app/user
 
--- a/shard.yml
+++ b/shard.yml
@@ -8,7 +8,7 @@ authors:
   - Nick Franken <<shnick@gmail.com>>
   - Benoit de Chezelles <<benoit.dechezelles@gmail.com>>
 
-crystal: 0.23.1
+crystal: 0.24.1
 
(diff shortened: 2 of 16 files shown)


wontruefree:
0.24.0 was auto upgraded on a few of my systems.
I would like to follow the upgrade.  I think there are a few PRs on supporting projects.

robacarp:
It might be simpler to update the code to the 24.1 syntax and switch on monkeypatching the 23.1 classes. Random/secure should be easy enough to backport, and that covers several of the changes made here.

elorest:
I think when 24.1 comes out we should just add support for it and mark 0.24.1 as the minimum version that amber 0.5.1 supports.

I don't think we should try supporting 2 versions of crystal. When 0.24.1 officially comes out most people will only have it.

drujensen:
@faustinoaq They will need to publish the assets for me to update the docker image to 0.24.1.  

for example, `https://github.com/crystal-lang/crystal/releases/download/0.24.0/crystal-0.24.0-2-linux-x86_64.tar.gz`

wontruefree:
Is there documentation on the syntax changes of the 24.1 release?

faustinoaq:
Moved to #454

Issue #67, opened 2017-05-19
Route Scopes


elorest:
Scopes are a way to group routes under a common path prefix and scoped set of plug middleware. We might want to do this for admin functionality, APIs, and especially for versioned APIs. Let's say we have user generated reviews on a site, and that those reviews first need to be approved by an admin. The semantics of these resources are quite different, and they might not share the same controller. Scopes enable us to segregate these routes.

eliasjpr:
Resolved with 

https://github.com/Amber-Crystal/amber/pull/43
https://github.com/Amber-Crystal/amber/pull/40
https://github.com/Amber-Crystal/amber/pull/35

PR #457, opened 2017-12-24
[Pipe] Add support client reloading

### LINKED ISSUE #347: Allow hot code reloading using Amber tasks

faustinoaq:
### Description of the Change

Hi team 😄 Happy Christmas! 🎉 

- Add Hot reloading for CSS
- Add reloading for files on public folder
- Allow reloading browser after server reload

ezgif-5-1c180670af

_`Backend reloading, client reloading and hot reloading (preserving state)`_

### Alternate Designs

Using `Amber::Tasks` ¿? @eliasjpr 

### Benefits

Fix  #347 definitely

### Possible Drawbacks

- ~~`Amber::Pipe::Reload` keeps enabled on production environments~~
- `Amber::Pipe::Reload` modifies `"content-type"`, so json/text/other responses will have injected code too, a posible solution would be adding `Pipe::Reload` after Controllers processing, then injected code only exists if `"content-type"` is equal to `"text/html"` instead of enforcing `"text/html"`

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -27,7 +27,7 @@ dependencies:
 
   slang:
     github: jeromegn/slang
-    branch: master 
+    branch: master
 
   redis:
     github: stefanwille/crystal-redis
@@ -40,10 +40,6 @@ dependencies:
   teeplate:
     github: amberframework/teeplate
     version: ~> 0.5.0
-
-  sentry:
-    github: samueleaton/sentry
-    version: ~> 0.1.0
 
   micrate:
     github: juanedi/micrate
@@ -67,8 +63,18 @@ dependencies:
 
   sqlite3:
     github: crystal-lang/crystal-sqlite3
+<<<<<<< HEAD
     version: ~> 0.9.0
 
 development_dependencies:
   ameba:
     github: veelenga/ameba
+||||||| parent of 7c3a6b4... Cleanup
+    version: ~> 0.8.3
+
+  watcher:
+    github: faustinoaq/watcher
+    version: ~> 0.2.0
+=======
+    version: ~> 0.8.3
+>>>>>>> 7c3a6b4... Cleanup
--- a/shard.yml
+++ b/shard.yml
@@ -63,18 +63,9 @@ dependencies:
 
   sqlite3:
     github: crystal-lang/crystal-sqlite3
-<<<<<<< HEAD
     version: ~> 0.9.0
 
 development_dependencies:
   ameba:
     github: veelenga/ameba
-||||||| parent of 7c3a6b4... Cleanup
-    version: ~> 0.8.3
-
-  watcher:
-    github: faustinoaq/watcher
-    version: ~> 0.2.0
-=======
-    version: ~> 0.8.3
->>>>>>> 7c3a6b4... Cleanup
+    version: ~> 0.4.0
--- a/src/amber/cli/helpers/process_runner.cr
+++ b/src/amber/cli/helpers/process_runner.cr
@@ -1,11 +1,10 @@
-require "sentry"
-
 module Sentry
   class ProcessRunner
     property processes = [] of Process
     property process_name : String
     property files = [] of String
     @logger : Amber::Environment::Logger
+    FILE_TIMESTAMPS = {} of String => String
 
     def initialize(
                    @process_name : String,
@@ -16,6 +15,8 @@ module Sentry
                    files = [] of String,
                    @logger = Amber::CLI.logger)
       @files = files
+      @npm_process = false
+      @app_running = false
     end
 
     def run
@@ -26,34 +27,35 @@ module Sentry
     end
 
     # Compiles and starts the application
-    #
     def start_app
-      stop_all_processes
-      start_all_processes
+      build_result = build_app_process
+      if build_result && build_result.success?
+        stop_all_processes
+        create_all_processes
+        @app_running = true
+      elsif !@app_running
+        log "Compile time errors detected. Shutting down..."
+        exit 1
+      else
+        log "Unknown error occurred, Shutting down..."
+        exit 1
+      end
     end
 
     private def scan_files
       file_changed = false
-
       Dir.glob(files) do |file|
         timestamp = get_timestamp(file)
-        if FILE_TIMESTAMPS[file]? && FILE_TIMESTAMPS[file] != timestamp
+        msg_prefix = FILE_TIMESTAMPS[file]? ? "" : "Watching file: "
+
+        if FILE_TIMESTAMPS[file]? != timestamp
           FILE_TIMESTAMPS[file] = timestamp
+          log "#{msg_prefix}./#{file.colorize(:light_gray)}"
           file_changed = true
-          log "#{file.capitalize.colorize(:light_gray)}"
-        elsif FILE_TIMESTAMPS[file]?.nil?
-          FILE_TIMESTAMPS[file] = timestamp
-          file_changed = true
-          log "Watching file: #{file.capitalize.colorize(:light_gray)}"
         end
       end
 
-      start_app if (file_changed)
-    end
-
-    private def start_all_processes
-      log "Compiling #{project_name}..."
-      create_all_processes
+      start_app if file_changed
     end
 
     private def stop_all_processes
@@ -65,10 +67,11 @@ module Sentry
     end
 
     private def create_all_processes
-      build_app_process
       @processes << create_watch_process
-      sleep 3
-      create_npm_process
+      unless @npm_process
+        create_npm_process
+        @npm_process = true
+      end
     end
 
     private def build_app_process
@@ -83,7 +86,7 @@ module Sentry
 
     private def create_npm_process
       node_log "Installing dependencies..."
-      Process.new("npm install && npm run watch", shell: true, output: Process::Redirect::Inherit, error: Process::Redirect::Inherit)
+      Process.new("npm install --loglevel=error && npm run watch", shell: true, output: Process::Redirect::Inherit, error: Process::Redirect::Inherit)
       node_log "Watching public directory"
     end
 
--- a/src/amber/cli/templates/app/config/routes.cr
+++ b/src/amber/cli/templates/app/config/routes.cr
@@ -7,6 +7,11 @@ Amber::Server.configure do |app|
     plug Amber::Pipe::Session.new
     plug Amber::Pipe::Flash.new
     plug Amber::Pipe::CSRF.new
+
+    # Reload clients browsers (development only)
+    if Amber.env.development?
+      plug Amber::Pipe::Reload.new
+    end
   end
 
   # All static content will run these transformations
--- a/src/amber/cli/templates/app/config/routes.cr
+++ b/src/amber/cli/templates/app/config/routes.cr
@@ -9,9 +9,7 @@ Amber::Server.configure do |app|
     plug Amber::Pipe::CSRF.new
 
     # Reload clients browsers (development only)
-    if Amber.env.development?
-      plug Amber::Pipe::Reload.new
-    end
+    plug Amber::Pipe::Reload.new
   end
 
   # All static content will run these transformations
--- a/src/amber/router/pipe/reload.cr
+++ b/src/amber/router/pipe/reload.cr
@@ -1,0 +1,27 @@
+require "../../support/client_reload"
+
+module Amber
+  module Pipe
+    # Reload clients browsers using `ClientReload`.
+    #
+    # NOTE: Amber::Pipe::Reload is intended for use in a development environment.
+    # ```
+    # pipeline :web do
+    #   plug Amber::Pipe::Reload.new
+    # end
+    # ```
+    class Reload < Base
+      def initialize
+        Support::ClientReload.new
+        super
+      end
+
+      def call(context : HTTP::Server::Context)
+        if context.format == "html"
+          context.response.print Support::ClientReload::INJECTED_CODE
+        end
+        call_next(context)
+      end
+    end
+  end
+end
--- a/src/amber/router/pipe/reload.cr
+++ b/src/amber/router/pipe/reload.cr
@@ -11,14 +11,13 @@ module Amber
     # end
     # ```
     class Reload < Base
-      def initialize
+      def initialize(@env : Amber::Environment::Env = Amber.env)
         Support::ClientReload.new
-        super
       end
 
       def call(context : HTTP::Server::Context)
-        if context.format == "html"
-          context.response.print Support::ClientReload::INJECTED_CODE
+        if @env.development? && context.format == "html"
+          context.response << Support::ClientReload::INJECTED_CODE
         end
         call_next(context)
       end
--- a/src/amber/support/client_reload.cr
+++ b/src/amber/support/client_reload.cr
@@ -1,0 +1,106 @@
+module Amber::Support
+  # Used by `Amber::Pipe::Reload`
+  #
+  # Allow clients browser reloading using WebSockets and file watchers.
+  struct ClientReload
+    FILE_TIMESTAMPS = {} of String => String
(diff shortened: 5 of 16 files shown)


robacarp:
This is great!

src/amber/support/client_reload.cr
```diff
@@ -0,0 +1,94 @@
...
+        session.@ws.send msg
+      end
+    end
+
+    private def watch_files
+      puts "🔃  Your ClientBot is vigilant. beep-boop..."
+      watch(["public/**/*"]) do |event|
+        event.on_change do |files|
+          files.each do |file, timestamp|
+            puts "🔃  watching file: ./#{file}"
```
eliasjpr: This should use the new logger `Amber.logger.puts "watching file: ./#{file}", "Watcher`

src/amber/support/client_reload.cr
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    private def reload_clients(msg)
+      SESSIONS.each do |session|
+        session.@ws.send msg
+      end
+    end
+
+    private def watch_files
+      puts "🔃  Your ClientBot is vigilant. beep-boop..."
```
eliasjpr: I would remove this log is not useful info

shard.yml
```diff
@@ -68,3 +68,7 @@ dependencies:
   sqlite3:
     github: crystal-lang/crystal-sqlite3
     version: ~> 0.8.3
+
+  watcher:
+    github: faustinoaq/watcher
+    version: ~> 0.2.0
```
eliasjpr: New line missing

src/amber/router/pipe/reload.cr
```diff
@@ -0,0 +1,26 @@
...
+    # end
+    # ```
+    class Reload < Base
+      def initialize
+        Support::ClientReload.new
+        super
+      end
+
+      def call(context : HTTP::Server::Context)
+        context.response.print Support::ClientReload::INJECTED_CODE
```
eliasjpr: Does print close the response?

faustinoaq: @eliasjpr I don't think so

What other method should I use?

src/amber/support/client_reload.cr
faustinoaq: Yeah, I thought you weren't using logger on watcher, I just realize latest amber `v0.5.0` was republished to fix logs. 😅 

Let me fix this 👍

shard.yml
```diff
@@ -68,3 +68,7 @@ dependencies:
   sqlite3:
     github: crystal-lang/crystal-sqlite3
     version: ~> 0.8.3
+
+  watcher:
+    github: faustinoaq/watcher
```
drujensen: Is this a replacement for sentry? If so, can we remove sentry?

src/amber/support/client_reload.cr
```diff
@@ -0,0 +1,94 @@
...
+      // <![CDATA[  <-- For SVG support
+      if ('WebSocket' in window) {
+        (function() {
+          function refreshCSS() {
+            console.log('Reloading CSS...');
+            var sheets = [].slice.call(document.getElementsByTagName('link'));
+            var head = document.getElementsByTagName('head')[0];
+            for (var i = 0; i < sheets.length; ++i) {
+              var elem = sheets[i];
+              head.removeChild(elem);
```
drujensen: Should we only remove and add the link if its a `stylesheet` or should we do this for all links?

src/amber/cli/templates/app/shard.yml.ecr
```diff
@@ -57,3 +57,7 @@ development_dependencies:
   garnet_spec:
     github: amberframework/garnet-spec
     version: ~> 0.1.1
+
```
drujensen: If Amber requires this as a dependency, do we need it here as well?

faustinoaq: You're right, let me fix this 👍

src/amber/support/client_reload.cr
faustinoaq: I just copied a working code from nodejs package `live-server`, What you suggest we could use? 😅

src/amber/router/pipe/reload.cr
eliasjpr: If it does not close the response thats good ccause `Pipe::Last` has to run.

src/amber/support/client_reload.cr
drujensen: The logic here gets all the `links`, removes it from the head, determines if its a `stylesheet` on line 62.  If it is, it adds a `_cacheOverride={Date()}` to the link to break the cache.  Then it adds the link back into the head.

What I'm wondering if we need to remove it then append it back on #67 if its not a stylesheet.  Why remove it and add it back if your not going to add a `cacheOverride`?

What if we only remove and add it back if its a stylesheet?  Something like this:
```
if (elem.href && typeof rel != 'string' || rel.length == 0 || rel.toLowerCase() == 'stylesheet') {
  head.removeChild(elem);
  var url = elem.href.replace(/(&|\\?)_cacheOverride=\\d+/, '');
  elem.href = url + (url.indexOf('?') >= 0 ? '&' : '?') + '_cacheOverride=' + (new Date().valueOf());
  head.appendChild(elem);
}
```

faustinoaq:
Hi team, I have some comments about amber and this PR:

> Does this replaces the npm dependency?

@eliasjpr A backend asset proccess like NPM is still needed to compile and bundle things like TypeScript and SCSS files. This code reloading feature just check file changes on public files. You can use Vanilla JS and CSS without NPM.

> Do we need to remove the npm run script from in process runner? See

I think NPM inside `amber watch` is nice, but what about other tools like yarn, or what happens if a developers want to use webpack-dev-server with proxy or rollupjs instead.

We can just disable `Amber::Pipe::Reload` by commenting this on pipeline, but How can we disable NPM on `amber watch`?

Maybe we can use flags like `--npm` and `--yarn` to support them optionally.

@eliasjpr Before you include NPM inside `amber watch` I have been using an extra terminal for assets stuff. Maybe backend developers like all in one place working by default, but I think front-end developers have so many JS tools, so enforcing them to use NPM only isn't good 😅

Other comments:

1) Sentry is not needed anymore, It has been embedded completely inside amber, see `Sentry::ProcessRunner` implemented on #458 
2) Amber `v0.5.0` is running NPM install on every reload, this is a unwanted behavior, so I added a flag to run it at start only.
3) I removed NPM warnings using `--loglevel=error` option

shard.yml
faustinoaq: Done!

src/amber/support/client_reload.cr
faustinoaq: Done!

src/amber/support/client_reload.cr
faustinoaq: Done!

shard.yml
faustinoaq: I removed both 👍 less dependencies, less to worry about

src/amber/support/client_reload.cr
faustinoaq: Done!

src/amber/support/client_reload.cr
```diff
@@ -29,19 +34,38 @@ module Amber::Support
...
+        timestamp = get_timestamp(file)
+
+        if FILE_TIMESTAMPS[file]? && FILE_TIMESTAMPS[file] != timestamp
+          FILE_TIMESTAMPS[file] = timestamp
+          file_changed = true
+          Amber.logger.puts "./#{file.colorize(:light_gray)}", "Watcher", :light_gray
+        elsif FILE_TIMESTAMPS[file]?.nil?
+          FILE_TIMESTAMPS[file] = timestamp
+          file_changed = true
+          Amber.logger.puts "Watching file: ./#{file.colorize(:light_gray)}", "Watcher", :light_gray
```
marksiemers: Do the messages need to be different - one saying `Watching file: ...` the other, looks the same other than that prefix.

Could this be simplified?
```crystal
private def scan_files
  Dir.glob(["public/**/*"]) do |file|
    timestamp = get_timestamp(file)
    msg_prefix = FILE_TIMESTAMPS[file]? ? "" : "Watching file: "

    if FILE_TIMESTAMPS[file]? != timestamp
      FILE_TIMESTAMPS[file] = timestamp
      Amber.logger.puts "#{msg_prefix}./#{file.colorize(:light_gray)}", "Watcher", :light_gray
      check_file(file)
    end
  end
end
```

src/amber/router/pipe/reload.cr
```diff
@@ -0,0 +1,27 @@
...
+    # end
+    # ```
+    class Reload < Base
+      def initialize
+        Support::ClientReload.new
+        super
+      end
+
+      def call(context : HTTP::Server::Context)
+        if context.format == "html"
```
faustinoaq: BTW, @eliasjpr I found another issue here 😅 context.format always return "html", so the following code, gets injected too:

```crystal
class HomeController < ApplicationController
  def index
    context.response.headers["context-header"] = "application/json"
    {foo: "bar"}.to_json
  end
end
```

How can I fix this?

src/amber/cli/helpers/process_runner.cr
```diff
@@ -26,10 +27,16 @@ module Sentry
...
-      start_all_processes
+      build_result = build_app_process
+      if build_result && build_result.success?
+        stop_all_processes
+        create_all_processes
+        @app_running = true
+      elsif !@app_running
+        log "Compile time errors detected. Shutting down..."
+        exit 1
+      end
```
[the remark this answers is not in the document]
faustinoaq: Done!  👍

src/amber/support/client_reload.cr
faustinoaq: Thank you! done 👍

drujensen:
> Maybe we can use flags like --npm and --yarn to support them optionally.

I think we should consider changing `amber watch` to be configurable similar to `foreman`using a `Procfile`. 

We specify a list of processes providing what files to watch and what command to run if any of the files change.  This will allow `amber watch` to be used for any toolset the user wants and it prevents the dependency on `npm`.  For example, they may want to watch the `/db` directory and restart `crystal db migrate`.  Or they will want to watch `/jobs` directory and restart `crystal run src/sidekiq.cr`.  These are just a couple examples.

faustinoaq:
I'm having some issues on TravisCI:

Passing 👉 https://travis-ci.org/amberframework/amber/jobs/321150453#L1606 ✔️ 
Failing 👉 https://travis-ci.org/amberframework/amber/jobs/321967054#L1620 ❎ (same code)

```crystal
require "../../../../spec_helper"

module Amber
  module Pipe
    describe Reload do
      it "client should contain injected code" do
        reload = Reload.new
        pipeline = Pipeline.new
        request = HTTP::Request.new("GET", "/reload")

        Amber::Server.router.draw :web do
          get "/reload", HelloController, :index
        end

        pipeline.build :web do
          plug Amber::Pipe::Reload.new
        end

        reload.next = ->(context : HTTP::Server::Context) { "Hello World!" }
        response = create_request_and_return_io(reload, request)

        response.body.should contain "Code injected by Amber Framework"
      end
    end
  end
end
```

faustinoaq:
@drujensen Maybe we can use `.amber.yml` adding some extra information, something like:

```yml
type: app
database: sqlite
language: slang
model: granite
watch:
  server:
    files: ["src/**/*.cr", "src/**/*.ecr"]
    command: shards build
  client:
    files: "public/**/*"
    command: npm run watcher
  db:
    files: "db/**/*"
    command: amber db migrate
  jobs:
    files: "src/jobs/*"
    command: crystal run src/sidekiq.cr
```

WDYT?

robacarp:
A dedicated Procfile has the added benefit of being an accepted standard on several existing servers and hosts

drujensen:
@faustinoaq yes, that is inline with what I was thinking.

@robacarp I wish the format for `Procfile` was more flexible.  Unfortunately, without the ability to specify which files to watch, we wouldn't be able to use it.

faustinoaq:
I think we can implement an `amber watch` file configuration in other PR 😅 

Something like:

```yaml
watch:
  server:
    files: ["src/**/*.cr", "src/**/*.ecr"]
    command: shards build
  client:
    files: "public/**/*"
    command: ["yarn install", "yarn run watch"]
  db:
    files: "db/**/*"
    command: amber db migrate
  jobs:
    files: "src/jobs/*"
    command: crystal run src/sidekiq.cr
```

WDYT?

BTW, I'm trying to pass Travis, but I don't know why how to fix/implement [this](https://github.com/amberframework/amber/pull/457/files#diff-946aa78ba53835a4d30a60e6737989a6R6) spec

eliasjpr:
@faustinoaq I think this commit https://github.com/amberframework/amber/pull/457/commits/a3826434900d6fc3e07bfb31d143ad0547a31aa6 fixes the specs. Let me know what you think.

eliasjpr:
Also I think you might want to rebase interactive to drop/squash some commits before merging

src/amber/cli/templates/app/config/routes.cr
```diff
@@ -7,6 +7,11 @@ Amber::Server.configure do |app|
     plug Amber::Pipe::Session.new
     plug Amber::Pipe::Flash.new
     plug Amber::Pipe::CSRF.new
+
+    # Reload clients browsers (development only)
+    if Amber.env.development?
```
elorest: Does this logic work? Seems to be a conditional around a macro that generates code.

faustinoaq:
> @faustinoaq I think this commit a382643 fixes the specs. Let me know what you think.

Yeah, Thank you! 🎉 

> Also I think you might want to rebase interactive to drop/squash some commits before merging

I agree 👍

src/amber/cli/templates/app/config/routes.cr
faustinoaq: I tried it and work as expected

drujensen:
I added #476 to make `amber watch` configurable

faustinoaq:
> Looks good. Don't think it's quite ready yet though is it?

@elorest Yeah, Still wondering How to fix [this](https://github.com/amberframework/amber/pull/457#discussion_r158740400) 😅 

Currently this implementation always injects the client code on every response because `context.format` always return `html` inside `Amber::Pipe::Reload`.

Finally, Another PR is needed to achieve #476

src/amber/router/pipe/reload.cr
eliasjpr: The  `Accept` request header must be set in order for it to return the correct format otherwise it will return `html` by default.

eliasjpr: Currently working on a fix for this

PR #460, opened 2017-12-24
official amber docker image

drujensen:
### Description of the Change

Amber needs an official docker image.  I have setup an `amberframework/amber` on docker hub.  

Let me know what your docker hub username is and I will add you to the organization.

This is a free service that will automatically build docker images.  Tags will be built using that version number and latest is pointing to master.

I have updated the Dockerfile to install the amber cli command to this image.

This will remove the `drujensen/crystal` dependency and use the official crystallang/crystal image as the root image.  Then it will install the amber dependencies, build amber cli and make it available to other images.

### Benefits

- This will increase the speed of launching docker for new projects since amber will already be built on the image.  

- We no longer need the `amber` target in the shards.yml for new projects since amber will exist on the docker image already.

- The image will match the release version.  This allows for older versions of amber to continue to work without breaking changes.

- The official amber docker image can be used on all platforms that support docker.

### Possible Drawbacks

- We are adding a new service `docker hub` to the mix.


Dockerfile
```diff
@@ -1,9 +1,18 @@
...
 
-WORKDIR /app/user
+# Install Dependencies
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update -qq && apt-get install -y --no-install-recommends libpq-dev libsqlite3-dev libmysqlclient-dev libreadline-dev git curl vim netcat
 
-ADD . /app/user
+# Install Node
+RUN curl -sL https://deb.nodesource.com/setup_6.x | bash -
+RUN apt-get install -y nodejs
```
faustinoaq: Why we still depend on node? Am I missing something? 😅

drujensen: We are generating webpack and use npm packages for front-end assets which is dependent on node.

drujensen:
@elorest I merged the two `Dockerfile`'s since they are now the same.  I've never tried `amber deploy` but I think I updated it correctly.  Let me know if I broke anything.

PR #461, opened 2017-12-25
Crystal 0.24.1 upgrade [DEPS]

elorest:
### Description of the Change

Multiple dependancies are broken so this can't be merged in yet.

So far I know that the following dependancies are broken.

* mosop/cli - IO has been change from module to class. Should be an easy fix.
* teeplate - Not sure what's wrong with this one.
* will probably fine more after fixing teeplate.


elorest:
When this is merged in don't squash.

shard.yml
```diff
@@ -38,8 +38,8 @@ dependencies:
     version: ~> 0.6.10
 
   teeplate:
-    github: mosop/teeplate
-    version: ~> 0.4.5
+    github: amberframework/teeplate
+    branch: master
```
faustinoaq: Why master?, I think we should create a tag for this 😅

elorest: Yeah that makes sense.

PR #462, opened 2017-12-25
handle the case where there is no accept header

damianham:
### Description of the Change

A HTTP client may not send an Accept header so the content type cannot be determined from the request headers.  This change returns the default content type when there is no Accept request header.

### Alternate Designs

None considered - Amber::Support::MimeTypes already has a default mime type and that is the value returned.

### Benefits

requests without an Accept header will not fail.

### Possible Drawbacks
None forseen

### DIFF
--- a/src/amber/router/context.cr
+++ b/src/amber/router/context.cr
@@ -70,14 +70,20 @@ class HTTP::Server::Context
   {% end %}
 
   def format
-    content_type = request.headers[FORMAT_HEADER].split(",").first
-    type = if content_type.includes?(";")
-             content_type.split(";").first
-           else
-             content_type
-           end
+    content_type = request.headers[FORMAT_HEADER]?
 
-    Amber::Support::MimeTypes.format(type)
+    if content_type
+      content_type = content_type.split(",").first
+      type = if content_type.includes?(";")
+        content_type.split(";").first
+      else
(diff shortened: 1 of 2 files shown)


src/amber/router/context.cr
```diff
@@ -70,14 +70,20 @@ class HTTP::Server::Context
...
+      type = if content_type.includes?(";")
+        content_type.split(";").first
+      else
+        content_type
+      end
+
+      Amber::Support::MimeTypes.format(type)
+    else
+      Amber::Support::MimeTypes.default
+    end
```
marksiemers: Sorry, I'm late to the game here. I'd like to see a refactoring of this code, perhaps something like this:

```crystal
def format
  return @format if @format
  if content_type
    @format = Amber::Support::MimeTypes.format(content_type)
  else
    @format = Amber::Support::MimeTypes.default
  end
end

private def content_type
  return @content_type if @content_type_parsed
  @content_type_parsed = true
  if (content_type = request.headers[FORMAT_HEADER]?)
    @content_type = content_type.split(",").first.split(";").first
  else
    @content_type = nil
  end
end
```
See some ad-hoc tests here: https://carc.in/#/r/3ajh

@amberframework/core-team - WDYT?

PR #464, opened 2017-12-25
Official Amber Docker Image

drujensen:
### Description of the Change

Amber needs an official docker image.  I have setup an `amberframework/amber` on docker hub.  

Let me know what your docker hub username is and I will add you to the organization.

This is a free service that will automatically build docker images.  Tags will be built using that version number and latest is pointing to master.

I have updated the Dockerfile to install the amber cli command to this image.

This will remove the `drujensen/crystal` dependency and use the official crystallang/crystal image as the root image.  Then it will install the amber dependencies, build amber cli and make it available to other images.

### Benefits

- This will increase the speed of launching docker for new projects since amber will already be built on the image.  

- We no longer need the `amber` target in the shards.yml for new projects since amber will exist on the docker image already.

- The image will match the release version.  This allows for older versions of amber to continue to work without breaking changes.

- The official amber docker image can be used on all platforms that support docker.

### Possible Drawbacks

- We are adding a new service `docker hub` to the mix.


Dockerfile
```diff
@@ -1,9 +1,18 @@
...
+RUN curl -sL https://deb.nodesource.com/setup_6.x | bash -
+RUN apt-get install -y nodejs
 
-RUN crystal deps
+WORKDIR /opt/amber
+
+# Build Amber
+ENV PATH /opt/amber/bin:$PATH
+ADD . /opt/amber
+RUN shards build amber
```
drujensen: This will add the `amber` cli command to the docker image.  I put it under `/opt/amber/bin` and added it to the PATH.  Crystal is under `/opt/crystal-head/bin` in their base image.

Dockerfile
```diff
@@ -1,9 +1,18 @@
...
 
 CMD ["crystal", "spec", "-D", "run_build_tests"]
```
drujensen: Now we have one docker image that contains crystal and amber.  It is the same image we use in travis that the tests are run against.

src/amber/cli/templates/app/docker-compose.yml.ecr
```diff
@@ -15,36 +15,29 @@ services:
...
-      - '3000:3000'
+      - 3000:3000
     links:
       - db
-      - mail
     volumes:
-      - '.:/app/local'
+    - .:/app
+    - nodes:/app/node_modules
+    - shards:/app/lib
```
drujensen: This is pretty cool feature I just found out.  We can add a volume to the node_modules and the shards directories.  This removes the need to run `npm install` and `shards install` locally.  It will use the `/lib` and `/node_modules` on the docker image.

now you can get up to speed even faster:
```bash
amber new blog
cd blog
docker-compose up -d
```

src/amber/cli/templates/app/docker-compose.yml.ecr
```diff
@@ -15,36 +15,29 @@ services:
...
-    working_dir: /app/local
     environment:
       DATABASE_URL: <%= @db_url %>
     volumes:
-      - '.:/app/local'
+      - .:/app
     links:
       - db
 
-  mail:
```
drujensen: I removed `mailcatcher` .  Although this is really cool, it slows down the startup of docker-compose.  I think we can document this capability instead of force it on all users since many apps will not require it.

src/amber/cli/templates/app/shard.yml.ecr
```diff
@@ -12,9 +12,6 @@ targets:
   <%= @name %>:
     main: src/<%= @name %>.cr
 
-  amber:
-    main: lib/amber/src/amber/cli.cr
-
```
drujensen: This target is no longer needed!  This was very confusing to see in the shard.yml.  Since amber is now on the docker image, we don't need to build it every time someone wants to use docker.

src/amber/cli/commands/deploy.cr
```diff
@@ -120,17 +120,17 @@ module Amber::CLI
         puts "Deploying project..."
         parallel(
           remote_cmd("docker network create --driver bridge ambernet"),
-          remote_cmd("docker build -t amberimage -f amberproject/config/deploy/Dockerfile amberproject")
+          remote_cmd("docker build -t amberimage -f amberproject/Dockerfile amberproject")
```
drujensen: Now the `deploy` command can use the base image as well.

Dockerfile
```diff
@@ -1,9 +1,18 @@
-FROM drujensen/crystal:0.23.1
+FROM crystallang/crystal:0.23.1
```
drujensen: I'm torn here.  Should we maintain a crystal image since the official one takes forever to get updated?

robacarp: The official crystal image takes forever to get updated and it's _massive_. I went down the path of trying to get crystal built under a much more minimal image when I built [my amber docker image](https://github.com/robacarp/amber_build/blob/master/Dockerfile) and ran too far into the weeds to continue, but I'd love to see an amber image that isn't so ginormous.

Dockerfile
robacarp: I don't see the benefit of putting binaries which are core to the function of the docker container out in `/opt/`. The [filesystem standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) says that `/opt` is for optional application software. Given that the point of this docker image is to run an amber server, the amber binary isn't optional. Should it live in `/usr/bin` or even `/bin`? Perhaps with the source in `/usr/src/amber`?

Dockerfile
robacarp: Actually, along those lines, a 2 stage docker file which builds amber and only copies the runtime dependencies into the final image, omitting the source code, would prevent the image from growing as the repo history grows.

Dockerfile
drujensen: @robacarp thanks for reviewing this.  

I think this depends on the purpose of the image.  This image is not meant for a production deployment, it's only focused on development.

I consider the crystal compiler and amber cli as development tools to generate an application.  The crystal image is adding their compiler to `/opt` as well and I think that makes sense, but i'm open to debate here.

Dockerfile
drujensen: @robacarp yeah, I noticed the crystal image is pretty big.  I think its including everything to build crystal including llvm. Also crystallang/crystal:0.24.1 hasn't been released yet.  This held us up last time as well.  I will change this back to my image.

Regarding the 2 stage docker file, For development, It's something @elorest and I discussed and sounds like we should consider it.

For production, I think `crystal` and `amber` don't need to exist on the image.  An alpine flavor linux base image and only the application binary produced with the dependent files would be needed.  Maybe we should consider doing this for the user with the `deploy` command?

robacarp: @drujensen In building my production image, I found that I needed the amber binary to run migrations. Is that actually an issue? Or is there a better way to run migrations through the compiled app binary?

Dockerfile
robacarp: Either way it’s a semantic issue not standing in the way of the image functionality. No sweat. The only thing I could imagine tripping people up is writing a path variable from scratch, but if the primary purpose of the image is development I don’t think that’s likely.

Dockerfile
robacarp: Tangentially, a crystal image based off of alpine is what I tried to build and ran into packages needed to build the compiler not being quickly available. LLVM, I think.

drujensen: > In building my production image, I found that I needed the amber binary to run migration

Interesting point.   For production at work, I have been pushing to have the main web application service to be minimalistic and immutable.   

https://diogomonica.com/2016/11/19/increasing-attacker-cost-using-immutable-infrastructure/

For migrations, I create a temporary service that performs the migration and then exits.  I guess that could use the `amber` image to run the migration? Another option is we provide the DDL from the migrations that can be ran manually. Not sure.  I haven't spent much time thinking about a production solution for Amber but I think it's good to start that conversation.

Regarding Alpine, yes, very minimalistic.  Perfect for production images if you can figure it out ;-).  I don't doubt you would struggle to get crystal on it. After we compile the application, we no longer need crystal and amber, or anyway, I'm hopeful that is true.

robacarp: Hm, I'm pushing to a dokku host (self hosted heroku-ish) which runs the same container for migrations. That might not be the only way to do it with dokku, and I'm not sure what heroku  needs to run migrations. @faustinoaq do you know?

The crecto migration files are essentially SQL and, while that does cause some problems with unspoken expectations about what an orm is expecting data-type wise, even a simple bash script is capable of chopping the file and running the SQL. Knowing which part of which migration to run is a little more subtle.

At the very least, dockerhub makes it really easy to cascade-rebuild an image when one of its dependencies changes so making multiple variant child images which have or don't have the  build stack, migration tools, etc wouldn't be hard.

faustinoaq: > I'm not sure what heroku needs to run migrations. @faustinoaq do you know?

@eliasjpr implemented migration on [heroku buildpack for amber](https://github.com/amberframework/heroku-buildpack-amber/pull/3)

eliasjpr: @robacarp to run migrations on heroku it just needs to be specified in the Procfile in the `release` stage see https://amberframework.org/guides/recipes/examples/heroku-deployments.md#heroku-deployments

Issue #469, opened 2017-12-31
Unexpected restarting server loop


faustinoaq:
### Description

Recently we implemented a restart loop to restart amber server when it fails (see #443). However this behavior is unexpected in other conditions like `bind: Cannot assign requested address` and other deploy errors. So instead of failing once, the server never starts but keep in a restart loop taking CPU at 100%.

### Steps to Reproduce

1. Configure development.yml with a wrong IP
2. Try to start the compiled server `./bin/myapp`
3. See restart loop and 100% of CPU usage.

**Expected behavior:** `./bin/myapp` fails once, showing error

**Actual behavior:** `./bin/myapp` fails forever, showing a restart loop and taking 100% CPU

**Reproduces how often:** 100%

### Versions

Amber `v0.5.0`

### Additional Information

![screenshot_20171231_153419](https://user-images.githubusercontent.com/3067335/34464098-128108a6-ee40-11e7-9169-6f3a2f53f87c.png)

I had to do `pkill myapp` because I was unable to use `Ctrl+C`. The restarting loop was too heavy 😞

eliasjpr:
This is something that was added recently here https://github.com/amberframework/amber/blob/master/src/amber/server/server.cr#L70 I am not a fan of the approach but this was to address a specific issue. See https://github.com/amberframework/amber/pull/443

faustinoaq:
@eliasjpr Yeah, I understand, but I think we should avoid restarting for first start. I mean:

```crystal
break if !@app_running
```

Like I implemented here 👉 https://github.com/amberframework/amber/pull/457/files#diff-9b4b623c272084b77e1c74c4e3111a48R36

BTW, on latest amber `v0.6.0`, `npm install && npm run watch` is executed on every change, is a bit annoying 😅 , also fixed here 👉 https://github.com/amberframework/amber/pull/457/files#diff-9b4b623c272084b77e1c74c4e3111a48R71 (run npm once at beginning)

veelenga:
Why not to stop restarting it after a couple of failed attempts:

```crystal
attempts = 0

loop do
  begin
    attempts += 1
    puts "trying to start server"
    raise Exception.new
  rescue e : Exception
    raise e if attempts >= 3
    puts "restarting"
  end
end

# trying to start server
# restarting
# trying to start server
# restarting
# trying to start server
# (Exception)

```

elorest:
I have a solution to this issue. Turned out that getting 0.24.1 working
[...]

PR #472, opened 2018-01-01
Only restarts server if it crashed due to File handler error.

elorest:
### Description of the Change

Check message and restart if its a file descriptor issue. Exist otherwise. There may be other errors we want to restart on as well but this is a good start.

Fixes https://github.com/amberframework/amber/issues/469

### DIFF
--- a/src/amber/server/server.cr
+++ b/src/amber/server/server.cr
@@ -74,8 +74,14 @@ module Amber
           server.listen(settings.port_reuse)
           exit
         rescue e : Errno
-          logger.error e.message
-          logger.info "Restarting server..."
+          if e.message == "accept: Too many open files"
+            logger.error e.message
+            logger.info "Restarting server..."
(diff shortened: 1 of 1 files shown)


src/amber/server/server.cr
```diff
@@ -74,8 +74,14 @@ module Amber
           server.listen(settings.port_reuse)
           exit
         rescue e : Errno
-          logger.error e.message
-          logger.info "Restarting server..."
+          if e.message == "accept: Too many open files"
```
eliasjpr: What is the error type in this case? Is the error type too generic to just rescue the this type of errors?

elorest: The error type is the one I'm rescuing from. Errno

elorest: Errno is the error returned for all errors related to creating a tcp socket.

PR #473, opened 2018-01-02
Fix checkbox overlapping

faustinoaq:
### Description of the Change

Fixes and apply https://github.com/amberframework/jasper-helpers/issues/21

Depends on https://github.com/amberframework/jasper-helpers/pull/22, (A new version of jasper-helper needs to be released)

### Alternate Designs

No

### Benefits

Improve accessibility 

### Possible Drawbacks

No


eliasjpr:
This is still failing even after merging amberframework/jasper-helpers#22 PR we have to release Jasper Helpers in order to have this pass. 

Error: `PostCommentController#label' is not expected to be invoked with a block, but a block was given`

Issue #467, opened 2017-12-28
CSRF token usability problems


infernalmaster:
Amber generates new CSRF token for every request. So when user have site opened in few tabs and did some POST action on one of them, it will create new token and invalidate token for all other tabs.
And this also creates some problems with POST Ajax action, because I need to update token inside every form.

 I looked at Rails sources and it generates one token on session start https://github.com/rails/rails/blob/5.1.3/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L373 . Then it generates masked token for every request that is used inside html https://github.com/rails/rails/blob/5.1.3/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L291 . So I don't understand this part: Why it generates masked token if it can be easily unmasked? 

Good article about rails csrf: https://medium.com/rubyinside/a-deep-dive-into-csrf-protection-in-rails-19fa0a42c0ef

So I for my needs I just simply patched pipe to used persistent token without masking (as I said I don't understand the purpose of mask). https://github.com/infernalmaster/brendari/blob/master/src/handlers/csrf.cr Maybe this is less secure but much more easy to use

eliasjpr:
@infernalmaster can you please open a PR with your proposed fix if we find it viable we would most likely approve the PR and merge to the codebase.

infernalmaster:
With this PR token regeneration on every request can be turned of like that

```
plug Amber::Pipe::CSRF.new(false)
```

This should be mentioned in docs or maybe this value can be `false` by default.

infernalmaster:
I read more about rails implementation and it's done like that to prevent http://breachattack.com. And I should improve code to prevent this type of attacks.

infernalmaster:
So I did it like Rails do. IMHO performance of module `TokenOperations` can be improved. Old functionality can be turned on like that: 

```
Amber::Pipe::CSRF.token_strategy = Amber::Pipe::CSRF::RefreshableToken
```

I leaved it because it much faster than new one.

elorest:
@infernalmaster what problems does refreshing every time have? Ajax forms should still be loaded dynamically. Even if they're stored just once per session you have to make sure that the ajax forms have the correct csrf. 

Refreshing every time is more secure because for long logins CSRF attacks could get smart and start sending the the correct csrf along with the post.

infernalmaster:
I have form with file upload widget. Widgets sends files with Ajax to server then returns filename on server. Then one of form fields with that filename. With refreshable token:
[...]

PR #474, opened 2018-01-03
Add optional ability to have persistent csrf token

infernalmaster:
Related to #467

### DIFF
--- a/src/amber/router/pipe/csrf.cr
+++ b/src/amber/router/pipe/csrf.cr
@@ -9,13 +9,12 @@ module Amber
       HEADER_KEY    = "X-CSRF-TOKEN"
       PARAM_KEY     = "_csrf"
       CSRF_KEY      = "csrf.token"
+      TOKEN_LENGTH = 32
 
-      def initialize(@refresh_token : Bool = true)
-      end
+      class_property token_strategry = PersistantToken
 
       def call(context : HTTP::Server::Context)
-        if valid_http_method?(context) || valid_token?(context)
-          context.session.delete(CSRF_KEY) if @refresh_token
+        if valid_http_method?(context) || self.class.token_strategry.valid_token?(context)
           call_next(context)
         else
           raise Amber::Exceptions::Forbidden.new("CSRF check failed.")
@@ -26,15 +25,8 @@ module Amber
         !CHECK_METHODS.includes?(context.request.method)
       end
 
-      def valid_token?(context)
-        request_token = context.params[PARAM_KEY]? || context.request.headers[HEADER_KEY]?
-        session_token = self.class.token(context).to_s
-
-        request_token && Crypto::Subtle.constant_time_compare(request_token, session_token)
-      end
-
       def self.token(context)
-        context.session[CSRF_KEY] ||= Random::Secure.urlsafe_base64(32)
+        token_strategry.token(context)
       end
 
       def self.tag(context)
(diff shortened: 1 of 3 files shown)


infernalmaster:
[First time](https://github.com/amberframework/amber/pull/474/commits/de2552678709ab11c87565e421483f505646375e) I implemented that simple way. But than I looked to Rails sources and found that [comment](https://github.com/rails/rails/blob/5.1.3/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L294)

elorest:
So the general idea is that the token can't be the same every time or someone could guess the ssl encryption key after multiple requests. Similar to padding oracle.

Thanks.

src/amber/router/pipe/csrf.cr
```diff
@@ -22,20 +25,89 @@ module Amber
...
+        module TokenOperations
+          extend self
+
+          # Creates a masked version of the authenticity token that varies
+          # on each request. The masking is used to mitigate SSL attacks
+          # like BREACH.
+          def mask(unmasked_token)
+            one_time_pad = Bytes.new(TOKEN_LENGTH).tap { |buf| Random::Secure.random_bytes(buf) }
+            encrypted_csrf_token = xor_byte_strings(one_time_pad, unmasked_token.bytes)
+            "#{String.new(one_time_pad)}#{encrypted_csrf_token}"
```
elorest: String.new is problematic as the bytes contained in `one_time_pad` aren't all ascii chars. We'll need to base64 encode it first.

src/amber/router/pipe/csrf.cr
```diff
@@ -22,20 +25,89 @@ module Amber
...
+
+          def unmask(masked_token)
+            one_time_pad = masked_token.bytes[0...TOKEN_LENGTH]
+            encrypted_csrf_token = masked_token.bytes[TOKEN_LENGTH..-1]
+            xor_byte_strings(one_time_pad, encrypted_csrf_token)
+          end
+
+          def xor_byte_strings(s1, s2) : String
+            s1.each_with_index { |c1, i| s2[i] ^= c1 }
+            String.new(s2.to_unsafe, TOKEN_LENGTH)
```
elorest: Same here with String.new.

src/amber/router/pipe/csrf.cr
```diff
@@ -22,20 +25,89 @@ module Amber
...
+            "#{String.new(one_time_pad)}#{encrypted_csrf_token}"
+          end
+
+          def unmask(masked_token)
+            one_time_pad = masked_token.bytes[0...TOKEN_LENGTH]
+            encrypted_csrf_token = masked_token.bytes[TOKEN_LENGTH..-1]
+            xor_byte_strings(one_time_pad, encrypted_csrf_token)
+          end
+
+          def xor_byte_strings(s1, s2) : String
```
elorest: These are bytes not strings. I would suggest returning bytes as well, then you can base64 the entire thing once.

infernalmaster:
Sorry for late reply. I did some changes, I'm not sure if it's what you wanted. This is the best I can do according to my knowledges of Crystal.

src/amber/router/pipe/csrf.cr
```diff
@@ -22,20 +25,94 @@ module Amber
...
+      module PersistentToken
+        extend self
+        extend BaseToken
+
+        def valid_token?(context)
+          request_token = request_token(context)
+          return false if request_token.nil?
+          decoded_request_token = Base64.decode_string(request_token)
+          return false if decoded_request_token.bytesize != 2 * TOKEN_LENGTH
+          unmaseked_request_token = TokenOperations.unmask(decoded_request_token)
```
elorest: Typo.

src/amber/router/pipe/csrf.cr
```diff
@@ -8,10 +9,12 @@ module Amber
       HEADER_KEY    = "X-CSRF-TOKEN"
       PARAM_KEY     = "_csrf"
       CSRF_KEY      = "csrf.token"
+      TOKEN_LENGTH  = 32
+
+      class_property token_strategy : PersistentToken | RefreshableToken = PersistentToken
```
elorest: Has anyone benchmarked the difference between these two?

eliasjpr:
@elorest can you address your comments in a commit so we can merge it in? WDYT?

eliasjpr:
Issue https://github.com/amberframework/amber/issues/467

infernalmaster:
For benchmarks I just wrote this code inside test cases 1:

```
          csrf = CSRF.new
          request = HTTP::Request.new("PUT", "/")

          CSRF.token_strategy = Amber::Pipe::CSRF::RefreshableToken
          Benchmark.ips do |x|
            x.report("RefreshableToken") do
              token = CSRF.token(context)
              context.params[Amber::Pipe::CSRF::PARAM_KEY] = token.to_s
              result = csrf.call(context)
              if result != nil
                raise Exception.new("csrf failed")
              end
            end
          end

          CSRF.token_strategy = Amber::Pipe::CSRF::PersistentToken
          Benchmark.ips do |x|
            x.report("PersistentToken") do
              token = CSRF.token(context)
              context.params[Amber::Pipe::CSRF::PARAM_KEY] = token.to_s
              result = csrf.call(context)
              if result != nil
                raise Exception.new("csrf failed")
              end
            end
          end
```

Than I used flag --release and results are:
```
RefreshableToken 380.88k (  2.63µs) (± 4.40%) fastest
PersistentToken 308.93k (  3.24µs) (± 6.16%) fastest
```

I also tried almost the same (test case 2):

```

          n = _____
          Benchmark.bm do |x|
            x.report("RefreshableToken:") do
              CSRF.token_strategy = Amber::Pipe::CSRF::RefreshableToken
              csrf = CSRF.new
              request = HTTP::Request.new("PUT", "/")

              n.times do
                token = CSRF.token(context)
                context.params[Amber::Pipe::CSRF::PARAM_KEY] = token.to_s
                result = csrf.call(context)
                if result != nil
                  raise Exception.new("csrf failed")
                end
              end
            end
            x.report("PersistentToken:") do
              CSRF.token_strategy = Amber::Pipe::CSRF::PersistentToken
              csrf = CSRF.new
              request = HTTP::Request.new("PUT", "/")

              n.times do
                token = CSRF.token(context)
                context.params[Amber::Pipe::CSRF::PARAM_KEY] = token.to_s
                result = csrf.call(context)
                if result != nil
                  raise Exception.new("csrf failed")
                end
              end
            end
          end
```

With n=100_000:

```
RefreshableToken:   0.050000   0.230000   0.280000 (  0.265488)
PersistentToken:    0.100000   0.240000   0.340000 (  0.313201)
```

N=1000

```
RefreshableToken:   0.000000   0.000000   0.000000 (  0.002422)
PersistentToken:    0.010000   0.010000   0.020000 (  0.004350)
```


Than I tried simple test (test case 3):

```
          n = 100000
          Benchmark.bm do |x|
            x.report("RefreshableToken:") do
              CSRF.token_strategy = Amber::Pipe::CSRF::RefreshableToken

              n.times do
                t = CSRF.token(context)
                if t == "x"
                  raise Exception.new("csrf failed")
                end
              end
            end
            x.report("PersistentToken:") do
              CSRF.token_strategy = Amber::Pipe::CSRF::PersistentToken

              n.times do
                t = CSRF.token(context)
                if t == "x"
                  raise Exception.new("csrf failed")
                end
              end
            end
          end
```
Results:
```
RefreshableToken:   0.010000   0.000000   0.010000 (  0.002404)
PersistentToken:    0.080000   0.270000   0.350000 (  0.298745)
```

So I tried to understand why previous results where close for this strategies and tried to run test case 2 (n=100_000) without this operations `context.session.delete(CSRF_KEY)` inside RefreshableToken:

```
RefreshableToken:   0.020000   0.000000   0.020000 (  0.022798)
PersistentToken:    0.100000   0.250000   0.350000 (  0.318500)
```

So looks like session modification is very slow: 0.265488 - 0.022798 = 0.24269 seconds.

PR #480, opened 2018-01-05
Fix Static's fallthrough and directory_listing options

docelic:
Amber::Pipe::Static's initializer is inherited from parent class
and setting fallthrough and directory_listing options to false
in Amber did not have the desired effect of disabling them.

The solution is to override the initializer to have them set to
off by default as intended.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,25 +1,25 @@
 OUT_DIR=bin
-PREFIX ?=/usr/local
+PREFIX=/usr/local
 
 all: build force_link
 
 install: build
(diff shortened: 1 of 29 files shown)


eliasjpr:
This looks good. Can you please include a unit test to verify that this is correct.

eliasjpr:
Would have kept the other commits for separate PRs since they seemed to be unrelated to the static handler.

Now we have to test that the make file still works with the brew formula ☹️

Over all everything looks good. Thank you for fixing the typos.

docelic:
Will submit separate PRs.

PR #484, opened 2018-01-06
Makefile improvements

docelic:
- Remove invocations of `pwd` where not necessary
- Add invocations of `pwd` where necessary
- Replace PREFIX?= with PREFIX=

(The removal of ? in PREFIX?= makes this setting not susceptible to
unintentionally taking the value from the environment variable.
The correct way to set PREFIX is by doing 'make install
PREFIX=/path'
(as opposed to 'PREFIX=/path make install').)

- Do not run link (which requires admin privileges) in default
  target (build)
- On install, remove destination file before copying the binary
  (solves problem when a symlink already exists and copying fails)
- Make build depend on lib/
- Run shards only once, if lib/ doesn't exist
- Better qualify $OUT_DIR
- Add missing echo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,11 @@ install: build
 run:
 	$(OUT_DIR)/amber
 
-link:
+link: build
 	@echo "Symlinking $(OUT_DIR)/amber to $(PREFIX)/bin/amber"
 	@ln -s $(OUT_DIR)/amber $(PREFIX)/bin/amber
 
-force_link:
+force_link: build
 	@echo "Symlinking $(OUT_DIR)/amber to $(PREFIX)/bin/amber"
 	@ln -sf $(OUT_DIR)/amber $(PREFIX)/bin/amber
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,26 +1,33 @@
-OUT_DIR=bin
-PREFIX ?=/usr/local
+OUT_DIR=$(shell pwd)/bin
```
drujensen: :+1:

Makefile
```diff
@@ -1,26 +1,33 @@
...
 	@echo "Building amber in $(shell pwd)"
-	@mkdir -p `pwd`/$(OUT_DIR)
-	@crystal build -o `pwd`/$(OUT_DIR)/amber src/amber/cli.cr -p --no-debug
+	@mkdir -p $(OUT_DIR)
+	@crystal build -o $(OUT_DIR)/amber src/amber/cli.cr -p --no-debug
+
+lib:
+	@crystal deps
+
+install: build
```
drujensen: yeah!

PR #485, opened 2018-01-06
Minor updates to README

docelic:
(This PR is dependent on #484 because it mentions running just 'make' as a way to build including downloading the shards, which #484 makes possible.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,16 +21,16 @@
 
 **Amber** is a web application framework written in [Crystal](http://www.crystal-lang.org) inspired by Kemal, Rails, Phoenix and other popular application frameworks.
 
-The purpose of Amber is not to create yet another framework, but to take advantage of the beautiful Crystal language capabilities and provide engineers an efficient, cohesive, and well maintained web framework for the crystal community that embraces the language philosophies, conventions, and guides.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,20 +58,22 @@ Ensure you have the necessary dependencies:
...
-Once you have these dependencies, You can build the `amber` tool from source:
+Once you have these dependencies, you can build `amber` from source:
 
 
 ```shellsession
 $ git clone git@github.com:amberframework/amber.git
-$ cd amber/
-$ shards install
+$ cd amber
+$ make
```
eliasjpr: you have to run the shards install before running make

README.md
```diff
@@ -213,7 +215,7 @@ It's not enough to be brilliant when you're alone in your programming lair. You
 
 We have adopted the Contributor Covenant to be our [CODE OF CONDUCT](.github/CODE_OF_CONDUCT.md) guidelines for Amber.
 
-## Have a Amber based project?
+## Have an Amber-based Project?
```
eliasjpr: 👍

README.md
eliasjpr: It seems that this is related to this change https://github.com/amberframework/amber/pull/484/files#diff-b67911656ef5d18c4ae36cb6741b7965R14

If so running `make install` is sufficient here?

docelic: Yes, 'make install' would download deps, compile, and install. (Assuming that the privileges under which process is running are enough to do an installation in /usr/local/bin)

Issue #295, opened 2017-10-07
amber routes missing socket routes


drujensen:
### Description

When i call amber routes in cli, the socket routes does not show in result table.

### Steps to Reproduce

define a socket route in the routes.cr file

run `amber routes`

**Expected behavior:** [What you expect to happen]

to see the socket route

**Actual behavior:** [What actually happens]

no socket route is displayed

**Reproduces how often:** [What percentage of the time does it reproduce?]

Always

### Versions

0.2.7

### Additional Information

This issue was submitted via Gitter by @iambudi

faustinoaq:
I can reproduce it on amber `v0.3.6` too

docelic:
https://github.com/amberframework/amber/pull/494

marksiemers:
#494

PR #492, opened 2018-01-09
Add websocket routes to 'amber routes' output (#295)

docelic:
- Show websocket routes in 'amber routes'
- Avoid string "[Time] Class |" that used to be printed in the first line
  of output
- Fix general issue where in case of Con::Troller, only "Con" would be printed
  in Controller column

### DIFF
--- a/src/amber/cli/commands/routes.cr
+++ b/src/amber/cli/commands/routes.cr
@@ -5,9 +5,10 @@ require "../helpers/sentry"
 module Amber::CLI
   class MainCommand < ::Cli::Supercommand
     class Routes < Command
-      RESOURCE_ROUTE_REGEX = /(\w+)\s+\"([^\"]+)\",\s*(\w+)(?:,\s*(\w+)\:\s*\[([^\]]+)\])?/
-      VERB_ROUTE_REGEX     = /(\w+)\s+\"([^\"]+)\",\s*(\w+),\s*:(\w+)/
-      PIPE_SCOPE_REGEX     = /routes\s+\:(\w+)(?:,\s+\"([^\"]+)\")?/
+      RESOURCE_ROUTE_REGEX = /(\w+)\s+\"([^\"]+)\",\s*([\w:]+)(?:,\s*(\w+)\:\s*\[([^\]]+)\])?/
+      VERB_ROUTE_REGEX     = /(\w+)\s+\"([^\"]+)\",\s*([\w:]+),\s*:(\w+)/
+      WEBSOCKET_ROUTE_REGEX= /(websocket)\s+\"([^\"]+)\",\s*([\w:]+)/
+      PIPE_SCOPE_REGEX     = /(routes)\s+\:(\w+)(?:,\s+\"([^\"]+)\")?/
 
       LABELS         = ["Verb", "Controller", "Action", "Pipeline", "Scope", "URI Pattern"]
       ACTION_MAPPING = {
@@ -34,6 +35,7 @@ module Amber::CLI
       def run
         parse_routes
         print_routes_table
+        exit 0
       rescue
         puts "Error: Not valid project root directory.".colorize(:red)
         puts "Run `amber routes` in project root directory.".colorize(:light_blue)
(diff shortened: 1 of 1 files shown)


src/amber/cli/commands/routes.cr
```diff
@@ -120,8 +128,7 @@ module Amber::CLI
             row.add_column route[l].to_s
           end
         end
-        puts table
-        exit
```
marksiemers: Any reason to remove `exit`?

src/amber/cli/commands/routes.cr
```diff
@@ -5,9 +5,10 @@ require "../helpers/sentry"
 module Amber::CLI
   class MainCommand < ::Cli::Supercommand
     class Routes < Command
-      RESOURCE_ROUTE_REGEX = /(\w+)\s+\"([^\"]+)\",\s*(\w+)(?:,\s*(\w+)\:\s*\[([^\]]+)\])?/
-      VERB_ROUTE_REGEX     = /(\w+)\s+\"([^\"]+)\",\s*(\w+),\s*:(\w+)/
-      PIPE_SCOPE_REGEX     = /routes\s+\:(\w+)(?:,\s+\"([^\"]+)\")?/
```
marksiemers: Did any of these actually change? Or is this a whitespace diff?

src/amber/cli/commands/routes.cr
```diff
@@ -104,8 +112,8 @@ module Amber::CLI
 
       private def set_pipe(pipe_string)
         if route_match = pipe_string.to_s.match(PIPE_SCOPE_REGEX)
-          @current_pipe = route_match[1]?
-          @current_scope = route_match[2]?
+          @current_pipe = route_match[2]?
+          @current_scope = route_match[3]?
```
marksiemers: Now I'm guessing those Regexs did change. Is there an extra capture group now?

src/amber/cli/commands/routes.cr
```diff
@@ -62,6 +64,12 @@ module Amber::CLI
             action: route_match[4]?, pipeline: current_pipe,
             scope: current_scope, uri_pattern: route_match[2]?
           )
+        elsif route_match = route_string.to_s.match(WEBSOCKET_ROUTE_REGEX)
+          build_route(
+            verb: route_match[1]?, controller: route_match[3]?,
+            action: "", pipeline: current_pipe,
+            scope: current_scope, uri_pattern: route_match[2]?
+          )
```
marksiemers: It looks like there is only one difference here:
```crystal
build_route(
  verb: route_match[1]?, controller: route_match[3]?,
  action: "", pipeline: current_pipe,
  scope: current_scope, uri_pattern: route_match[2]?
)
```
The `action` is an empty string when it is a websocket. Can this be DRYed up?

src/amber/cli/commands/routes.cr
docelic: They changed. Two changes are made:

1) A ":" is added inside the regex so that controllers containing ":" in the name would work properly
2) And () were added around 'routes' and 'websockets' words even though not strictly necessary technically. (I've added them for consistency with other regexes and future expansion to cover multiple words)

src/amber/cli/commands/routes.cr
docelic: Changed due to your #1 question above (my answer #2 there), so that's why the number increased.

src/amber/cli/commands/routes.cr
docelic: Yes, there is no reason for the function named "print" to also exit automatically. I've moved "exit" to the run() method, where it properly exits after printing the table.

src/amber/cli/commands/routes.cr
docelic: I was thinking this would be here for future more detailed separation/improvements to HTTP and websocket code paths which wouldn't be shared between both. But sure, can simplify for now, and the () around regexes that I've added at the top make this trivial. Will add a commit soon.

docelic: Actually, not simple to DRY. If you find a way, please improve upon my patch.

marksiemers:
@docelic - I added some tests in #494 
I didn't have write access to your repo (I don't think) so I included your commit in another branch.

docelic:
Great, I will close this PR and assume you'll handle #494. Thanks!

PR #494, opened 2018-01-09
Feature/add ws to routes cli

marksiemers:
### Description of the Change

Adds WebSocket routes to `amber routes` table. Also adds Specs for `amber routes` in general.

### Alternate Designs

Non considered

### Benefits

More complete routes table and test suite.

### Possible Drawbacks

Longer running test suite and fuller routes table.

### DIFF
--- a/spec/amber/cli/commands/routes/routes_spec.cr
+++ b/spec/amber/cli/commands/routes/routes_spec.cr
@@ -7,6 +7,7 @@ include CLIHelper
 include Cli::Spec::Helper
 
 module Amber::CLI
+  extend Helpers
   describe "amber routes" do
     begin
       context "in an `amber new` app with default options" do
@@ -14,12 +15,8 @@ module Amber::CLI
         scaffold_app(TESTING_APP)
         output = ""
         ENV["CRYSTAL_CLI_ENV"] = "test"
-        MainCommand.run %w(routes) do |cmd|
-          output = cmd.out.gets_to_end
-        end
-        output_lines = output.split("\n").reject do |line|
-          line =~ /(─┼─|═╦═|═╩═)/
-        end
+        MainCommand.run %w(routes) { |cmd| output = cmd.out.gets_to_end }
+        output_lines = route_table_rows(output)
 
         it "outputs the correct headers" do
           headers = %w(Verb Controller Action Pipeline Scope) << "URI Pattern"
@@ -45,6 +42,22 @@ module Amber::CLI
           expectations = %w(get HomeController index web /)
           expectations.each do |expectation|
             line.should contain expectation
+          end
+        end
+
+        describe "with a websocket route" do
+          add_routes :web, %(websocket "/electric", ElectricSocket)
+          MainCommand.run %w(routes) { |cmd| output = cmd.out.gets_to_end }
+          output_lines = route_table_rows(output)
+
+          it "outputs the web socket route" do
+            expected = "websocket"
+            output.should contain expected
+            line = output_lines.find(""){ |line| line.includes? expected }
+            expectations = %w(websocket ElectricSocket web /electric)
+            expectations.each do |expectation|
+              line.should contain expectation
+            end
           end
         end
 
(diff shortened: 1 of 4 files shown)


marksiemers:
This could still use some additional tests for the actual WebSocket routes.

src/amber/cli/commands/routes.cr
```diff
@@ -97,15 +96,23 @@ module Amber::CLI
         routes << route
       end
 
+      private def build_route(route_match)
+        build_route(
+          verb: route_match[1]?, controller: route_match[3]?,
+          action: route_match[4]? || "", pipeline: current_pipe,
+          scope: current_scope, uri_pattern: route_match[2]?
```
marksiemers: @docelic - This overload allowed for refactoring.

marksiemers:
The additional test is there. ready for review.

src/amber/cli/commands/routes.cr
docelic: Great, didn't think of it. Thanks for letting me know.

spec/amber/cli/commands/routes/routes_spec.cr
```diff
@@ -0,0 +1,69 @@
...
+module Amber::CLI
+  extend Helpers
+  describe "amber routes" do
+    begin
+      context "in an `amber new` app with default options" do
+        cleanup
+        scaffold_app(TESTING_APP)
+        output = ""
+        ENV["CRYSTAL_CLI_ENV"] = "test"
+        MainCommand.run %w(routes) { |cmd| output = cmd.out.gets_to_end }
```
robacarp: Since these two lines (18 and 19) aren't required to run for the test at line 53, should these and the following 3 tests be wrapped in a describe?

src/amber/cli/commands/routes.cr
```diff
@@ -5,9 +5,10 @@ require "../helpers/sentry"
 module Amber::CLI
   class MainCommand < ::Cli::Supercommand
     class Routes < Command
-      RESOURCE_ROUTE_REGEX = /(\w+)\s+\"([^\"]+)\",\s*(\w+)(?:,\s*(\w+)\:\s*\[([^\]]+)\])?/
-      VERB_ROUTE_REGEX     = /(\w+)\s+\"([^\"]+)\",\s*(\w+),\s*:(\w+)/
-      PIPE_SCOPE_REGEX     = /routes\s+\:(\w+)(?:,\s+\"([^\"]+)\")?/
+      RESOURCE_ROUTE_REGEX = /(\w+)\s+\"([^\"]+)\",\s*([\w:]+)(?:,\s*(\w+)\:\s*\[([^\]]+)\])?/
+      VERB_ROUTE_REGEX     = /(\w+)\s+\"([^\"]+)\",\s*([\w:]+),\s*:(\w+)/
+      WEBSOCKET_ROUTE_REGEX= /(websocket)\s+\"([^\"]+)\",\s*([\w:]+)/
+      PIPE_SCOPE_REGEX     = /(routes)\s+\:(\w+)(?:,\s+\"([^\"]+)\")?/
```
robacarp: Actually, one last comment. Regexes this big are hard to reason about. Since Crystal supports PCRE extended, what about adding comments to make this more friendly to future devs?

Something like this: https://stackoverflow.com/questions/34578333/correct-way-to-add-comments-to-a-regular-expression-in-php

spec/amber/cli/commands/routes/routes_spec.cr
marksiemers: done

src/amber/cli/commands/routes.cr
marksiemers: I didn't write any of the Regexs. Many were already there and @docelic modified them slightly and added the websocket one.

It would take me a while to figure them out (to your point), then comment them.
@elorest and/or @docelic - would you be willing to add comments to the regular expressions here?

docelic: Heya, my vote would be in favor of just committing as-is, because what we have is an improvement. Further incremental improvements can be done by other contributors if/when they want. But in case @elorest can spare some time for this, awesome!

Issue #491, opened 2018-01-08
`crystal spec` in a generated app runs against the development database


robacarp:
Amber new generates a spec_helper which doesn't set the `AMBER_ENV` var explicitly to test, which means that any time specs [are actually written](https://github.com/amberframework/amberframework.org/blob/master/spec/controllers/guide_controller_spec.cr) and [used](https://github.com/amberframework/amberframework.org/blob/master/.travis.yml), the `Model.clear` statements dropped all over the place nuke the development database.

### Additional Information

I'm able to get a halfway decent working `crystal spec` command with this spec_helper:

```crystal
require "spec"
ENV["AMBER_ENV"] = "test"

require "amber"
# require config before models, so they know how to connect to the database
require "../config/*"

# load up some database stuff and blindly run migrations, hoping for the best
require "micrate"
require "pg"
require "sqlite"
require "mysql"
Micrate::DB.connection_url = Amber.settings.database_url
Micrate::Cli.run_up

require "../src/ext/extensions"
require "../src/controllers/application_controller"
require "../src/controllers/**"
require "../src/mailers/**"
require "../src/models/**"
require "../src/policies/**"
require "../src/jobs/**"

Spec.before_each do
  Model.clear
  Model2.clear
  # etc
end
```

🥔

eliasjpr:
With the new version of amber your spec file should look more like this 

```
require "spec"
require "garnet_spec"
require "../config/*"

module Spec
  DRIVER = :chrome
  PATH   = "/usr/local/bin/chromedriver"

  # Not all server implementations will support every WebDriver feature.
  # Therefore, the client and server should use JSON objects with the properties
  # listed below when describing which features a session supports.
  capabilities = {
    browserName:              "chrome",
    version:                  "",
    platform:                 "ANY",
    javascriptEnabled:        true,
    takesScreenshot:          true,
    handlesAlerts:            true,
    databaseEnabled:          true,
    locationContextEnabled:   true,
    applicationCacheEnabled:  true,
    browserConnectionEnabled: true,
    cssSelectorsEnabled:      true,
    webStorageEnabled:        true,
    rotatable:                true,
    acceptSslCerts:           true,
    nativeEvents:             true,
    args:                     "--headless",
  }
end
```

marksiemers:
@robacarp - Just in case you aren't aware of the behavior of crystals `Spec.before_each` here:
```crystal
Spec.before_each do
  Model.clear
  Model2.clear
  # etc
end
```
[...]

PR #500, opened 2018-01-11
[Spec] Defaults specs environment to Tests

eliasjpr:
Issue: https://github.com/amberframework/amber/issues/491

### Description of the Change

Defaults spec to test environment for an Amber project.

### DIFF
--- a/src/amber/cli/templates/app/spec/spec_helper.cr
+++ b/src/amber/cli/templates/app/spec/spec_helper.cr
@@ -1,3 +1,5 @@
+ENV["AMBER_ENV"] ||= "test"
+
 require "spec"
 require "garnet_spec"
 require "../config/*"


src/amber/cli/templates/app/spec/spec_helper.cr
```diff
@@ -1,3 +1,5 @@
+ENV["AMBER_ENV"] ||= "test"
```
veelenga: Why is this set conditionally? User can accidentally export `AMBER_ENV`, forget about it and then run specs. ie:

```crystal
$ export AMBER_ENV=production
# do some stuff...
$ crystal spec
```

eliasjpr: The reason is that you can have the AMBER_ENV can be different for CI environment

robacarp: Would it be sufficient to exit when `AMBER_ENV == production`? Or perhaps issue a compile-time warning when `AMBER_ENV` already exists?

```crystal
{% if ENV["AMBER_ENV"]
puts "Warning: AMBER_ENV is set to #{ENV["AMBER_ENV"]}"
end %}
```

eliasjpr: @veelenga what I am trying to state is that having AMBER_ENV set to `test` will remove the ability from devs to use a different `environment.yml` config file for a continuous integration environment like travis.

See https://github.com/rspec/rspec-rails/blob/master/lib/generators/rspec/install/templates/spec/rails_helper.rb#L3

veelenga: I got. But look two lines below:

https://github.com/rspec/rspec-rails/blob/master/lib/generators/rspec/install/templates/spec/rails_helper.rb#L5-L6

You may do the same here.

eliasjpr: Nice catch!

PR #501, opened 2018-01-11
Change README header to Markdown only

faustinoaq:
### Description of the Change

Make README more compatible with other platforms like [Docker](https://hub.docker.com/r/amberframework/amber/)

### Alternate Designs

No

### Benefits

Fix this:

screenshot_20180111_124534

### Possible Drawbacks

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,14 @@
-<p align="center">
-  <img src="https://github.com/amberframework/site-assets/raw/master/images/amber.png" width="200">
-  <p align="center">Productivity. Performance. Happiness.<p>
-  <p align="center">
-    <sup>
-      <i>
-        Amber makes building web applications fast, simple, and enjoyable - with fewer bugs and blazing fast performance.
-      </i>
-    </sup>
-  </p>
-  <p align="center">
-    <a href="https://travis-ci.org/amberframework/amber"><img src="https://img.shields.io/travis/amberframework/amber.svg?maxAge=360"></a>
-    <a href="https://github.com/amberframework/amber/releases/latest"><img src="https://img.shields.io/github/tag/amberframework/amber.svg?maxAge=360"></a>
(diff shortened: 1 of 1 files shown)


marksiemers:
Good point on the centering.

Can just the `<p align="center">` remain as html and wrap the rest of the markdown or will that not parse?

elorest:
I feel like this was done because all markdown ways of doing it didn't look the same. If everything is the same I say we merge it though.

elorest:
Isn't the Docker readme just a copy of this? Or is it directly linked to the github readme. If the former then just changing it on the docker one makes sense.

marksiemers:
There was a specific commit that looks like the inverse of this: https://github.com/amberframework/amber/commit/07718026480f33fb883788a110c4374835f42561#diff-04c6e90faac2675aa89e2176d2eec7d8

faustinoaq:
> Can just the <p align="center"> remain as html and wrap the rest of the markdown or will that not parse?

Then the README would display a raw `<p align="center">` is some platforms like docker or `crystal docs` itself, WDYT?

> Isn't the Docker readme just a copy of this? Or is it directly linked to the github readme. If the former then just changing it on the docker one makes sense.

Perhaps, we can just modify docker readme and discard this PR.

> There was a specific commit that looks like the inverse of this PR.

@marksiemers Yeah, I did it, and worked fine until I saw https://hub.docker.com/r/amberframework/amber

eliasjpr:
We should just edit the docker Readme instead.

drujensen:
I updated the docker description with these changes.

faustinoaq:
Docker was fixed, this PR is not needed anymore.

Thank you @drujensen !

faustinoaq:
Oh, looks like docker is showing plain html again :sweat_smile: 

Issue #488, opened 2018-01-07
README Information on production mode


Xosmond:
I have noticed that there is not information on running Amber on production mode.

By the way, 
 - I was asking on how to log to a file instead of STDOUT? 
 - How to use correctly "process_count" (when I changed it to 2, I got an DB::Connection refuse)
Thanks

eliasjpr:
Hellow @Xosmond 

> I was asking on how to log to a file instead of STDOUT?
Assuming that you have a directory in your project called `/logs` with a file `production.log`

In your config/application.cr 

```crystal
Amber::Server.configure do |settings|
  if Amber.env.production?
    log_file = File.open("./logs/production.log", "w")
    settings.logger = ::Logger.new(log_file)
  end
end
```

> How to use correctly "process_count" (when I changed it to 2, I got an DB::Connection refuse)

It could be very possible that you have not generated any models or migration yet. Scaffold a migration or model using amber cli `amber g scaffold Posts title:text` then run `amber db create migrate`

Hope this help. If not feel free to chat on gitter https://gitter.im/amberframework/amber or continue the thread here

Xosmond:
@eliasjpr After some debugging, I can see that when I change `process_count` to two and i call `amber watch`, amber starts in development mode instead of production nevertheless `AMBER_ENV` variable

Anybody have an idea on why?

elorest:
Amber watch should never be used for anything but development.

Xosmond:
@elorest Yeah, but I was just making some benchmarks and understanding it.

faustinoaq:
There is something about compiling amber for production:

you need to set `AMBER_ENV` in both, compilation and running phase.

```
AMBER_ENV=production shards build myapp
AMBER_ENV=production ./bin/myapp

# or

export AMBER_ENV=production

shards build myapp
./bin/myapp
```

So, `amber watch` doesn't work in production because it doesn't read `AMBER_ENV` variable when your project gets compiled.

> Thats why I was getting DB::ConnectionRefused because DB url on development file was not for production environment.

Did you solve this? 😅

Xosmond:
@faustinoaq  I'm don't agree with you because of this:

Lets create an app with a basic post migration and then create an migrate the **production database** (using `AMBER_ENV=production amber db create migrate`).

Lets debug a little:
development.yml
```
....
process_count: 1
redis_url: "redis://localhost:6379"
database_url: postgres://localhost:5432/miracle_development
....
```
production.yml
```
....
[...]

PR #502, opened 2018-01-11
[Cluster] Sets the correct environment when Forking

eliasjpr:
Issue: https://github.com/amberframework/amber/issues/488

Current when using cluster mode is not passing the correct Amber environment 
to the forked processes. 

- Ensures forked processes start in the correct environment.

We this change forked processes will start in the correct environment.
screen shot 2018-01-11 at 4 42 59 pm


robacarp:
Shouldn't Amber be forwarding the entire environment?

PR #504, opened 2018-01-12
Bumps sentry version

samueleaton:
for crystal `v0.24.1` support

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -27,7 +27,7 @@ dependencies:
 
   slang:
     github: jeromegn/slang
-    branch: master 
+    branch: master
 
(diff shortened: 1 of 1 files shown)


faustinoaq:
Hi, We embedded sentry since https://github.com/amberframework/amber/pull/458, because we did some custom modifications to sentry for managing some amber processes. 

I just removed sentry dependency from `shards.yml` on https://github.com/amberframework/amber/pull/457

So this PR is not needed now 😅 

@samueleaton Thank you so much for sentry!!! ❤️

elorest:
Thanks for helping out.

Issue #233, opened 2017-09-05
Support HEAD requests


drujensen:
currently HEAD is returning 404
`curl --head http://localhost:3000`

docelic:
The issue seems to be that HEAD is treated as a standalone action and unless an explicit 'head' for a resource exists in routes, 404 is returned.

RFC2616 says for HEAD:
"The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request.:

Based on that, I believe the code would need to work like this:

1) If HEAD route specifically exists, use that route. When data is returned to client, do not send any body, even if route generated some
2) If a specific route does not exist, and if HEADs are enabled by default (config option, default: true), then follow the route identical as GET, but in the end return everything except actual response body
3) If a specific route does not exist, and HEADs are disabled in configuration, keep current behavior

Sounds good?

drujensen:
@docelic that is correct.  The way that Kemal and Kemalyst do this is add a HEAD route for every GET route.  We will need to avoid sending the body back if the request is a HEAD request.

https://github.com/kemalyst/kemalyst/blob/master/src/kemalyst/handler/router.cr#L144

docelic:
Ok submitted a basic PR in https://github.com/amberframework/amber/pull/497.

This needs some little bit of work before merging:

-----------
1) Currently, since HEAD is always added for GETs, it is impossible to create a custom, specific HEAD route if GET action exists as well, because the one added by GET will cause Amber::Exceptions::DuplicateRouteError)

This needs to be solved in any way we want:

1) We delete old route, create new one
2) Or we do not add automatic HEAD on GET if custom HEAD already exists
3) But if we choose #2, what if user first defines a GET, and then his custom HEAD (he'll get above error).
Is #3 that OK? Do we want to force users to pay attention to order of routes? Or we add another field to routes which specifies whether route was automatically or manually created, and if automatically, then those routes are removed in favor of custom/manually-defined ones.

-----------
2) Do we want a config option to control if HEADs are automatically created?

-----------
3) Specs are needed

eliasjpr:
Fixed with https://github.com/amberframework/amber/pull/506

Thanks to @docelic 🙌

PR #506, opened 2018-01-12
[Pipe] Adds support for head requests

eliasjpr:
Issue https://github.com/amberframework/amber/issues/233

Currently head requests are returning 404.

- Does not render response.body if the request is of type HEAD
- Converts route.cr from class to struct this should give a performance
improvement
- Changes :get symbol to "get"

With the changes it will be possible to make HEAD requests and the
response.body should be empty

Example:

 ○ http -v HEAD :3000

Request
----------------------------
HEAD / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: localhost:3000
User-Agent: HTTPie/0.9.9

Response
----------------------------
HTTP/1.1 200 OK
Connection: Keep-Alive
Content-Length: 0
Keep-Alive: timeout=5, max=10000
Set-Cookie: amber.session=eyJfZmxhc2giOiJ7fSJ9--mpOMtKPrJi1R9nRQyYtmlwkX5CY%3D; path=/; HttpOnly
X-Powered-By: Amber

✋ @docelic

### DIFF
--- a/spec/amber/router/pipe/pipeline_spec.cr
+++ b/spec/amber/router/pipe/pipeline_spec.cr
@@ -67,6 +67,13 @@ module Amber
           response.body.should eq "Destroy"
         end
 
+        it "perform HEAD request" do
+          request = HTTP::Request.new("HEAD", "/hello/1")
+          response = create_request_and_return_io(pipeline, request)
+          response.headers["Content-Length"].should eq "0"
+          response.body.empty?.should be_true
+        end
+
         it "initializes context with X-Powered-By: Amber" do
           request = HTTP::Request.new("GET", "/index/faustino")
           response = create_request_and_return_io(pipeline, request)
--- a/spec/amber/router/pipe/pipeline_spec.cr
+++ b/spec/amber/router/pipe/pipeline_spec.cr
@@ -37,37 +37,37 @@ module Amber
           response.body.should eq "Hello World!"
         end
 
-        it "perform GET request" do
+        it "responds to GET request" do
           request = HTTP::Request.new("GET", "/hello")
           response = create_request_and_return_io(pipeline, request)
           response.body.should eq "Index"
         end
 
-        it "perform PUT request" do
+        it "responds to PUT request" do
           request = HTTP::Request.new("PUT", "/hello/1")
           response = create_request_and_return_io(pipeline, request)
           response.body.should eq "Update"
         end
 
-        it "perform PATCH request" do
+        it "responds to PATCH request" do
           request = HTTP::Request.new("PATCH", "/hello/1")
           response = create_request_and_return_io(pipeline, request)
           response.body.should eq "Update"
         end
 
-        it "perform POST request" do
+        it "responds to POST request" do
           request = HTTP::Request.new("POST", "/hello")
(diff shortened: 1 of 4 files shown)


spec/amber/router/pipe/pipeline_spec.cr
```diff
@@ -67,6 +67,13 @@ module Amber
           response.body.should eq "Destroy"
         end
 
+        it "perform HEAD request" do
```
marksiemers: I think this should be `it "responds to HEAD requests"` or `it "provides response for HEAD requests"`

"Performing the request" is more what the client is doing.

spec/amber/router/pipe/pipeline_spec.cr
```diff
@@ -67,6 +67,13 @@ module Amber
...
+          request = HTTP::Request.new("HEAD", "/hello/1")
+          response = create_request_and_return_io(pipeline, request)
+          response.headers["Content-Length"].should eq "0"
+          response.body.empty?.should be_true
+        end
```
marksiemers: Should we check a positive (200) response when expected and a 404 code as well?

This may end up being scope creep for this PR, as these comments could apply to many of these tests.

src/amber/router/route.cr
```diff
@@ -1,5 +1,5 @@
 module Amber
-  class Route
+  struct Route
```
marksiemers: This seems like a big change, and I'm not sure it directly relates to the HEAD requests, does it?

eliasjpr: This was an improvement that I've been wanting to make. This has to do more with the inner workings of crystal. In short, the Route is just holding a struct of data.

For a better understanding of this change read this https://crystal-lang.org/docs/guides/performance.html

spec/amber/router/pipe/pipeline_spec.cr
eliasjpr: In that case we should check for all headers, this was more checking if the pipeline was processing the content.

src/amber/router/route.cr
marksiemers: I've read about the performance gains before, a struct goes on the stack and is immutable (or copied by value anyway).

So I think it will be an improvement, but I ran into issues when working on a caching solution where I tried a struct, but it didn't work because I needed parts to be mutable.

I suppose a route should never change during run time - it just seems like a non-zero risk, and putting it in here means if we rollback, then we rollback the HEAD functionality too.

spec/amber/router/pipe/pipeline_spec.cr
marksiemers: I see, I was muddying the waters between pipeline tests and controller tests. I should have looked closer at the naming.

docelic:
That way would prevent headers from being added, if a GET would set some.

eliasjpr:
@robacarp the head request cannot be hijack since the user my be setting headers during the process of an action. The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request For the given path.

eliasjpr:
@marksiemers removed the `struct` from routes.cr opening a PR just for that.

PR #508, opened 2018-01-13
[Route] Makes Routes a Struct instead of Class

eliasjpr:
This is a small enhancement change to follow Crystal Lang guidelines to
avoid unecessary memory allowcation.

See https://crystal-lang.org/docs/guides/performance.html for details

### DIFF
--- a/src/amber/router/route.cr
+++ b/src/amber/router/route.cr
@@ -1,5 +1,5 @@
 module Amber
-  class Route
+  struct Route
     property :handler, :action, :verb, :resource, :valve, :params, :scope, :controller
 
     def initialize(@verb : String,


marksiemers:
@amberframework/core-team - This could be an easy win for much better performance.

Are we comfortable with the existing test suite making reasonably sure that this won't break anything?

marksiemers:
Maybe this could even improve the TFB results. The amber version will have to be updated there though.

eliasjpr:
This is a small improvement. I would like for us to have a `Performance` milestone and take a look at the Crystal recommendations and apply it everywhere we can in the code. I know this is 🎶  to @elorest ears :)

marksiemers:
> This is a small improvement.

In what way? Are the routes only allocated on the heap once at initial load time, or are new instances allocated when a bunch of requests come in?

Either way, stack memory can be so much faster. Ruby strings that go on the stack are 2x as fast as heap strings: http://patshaughnessy.net/2012/1/4/never-create-ruby-strings-longer-than-23-characters

marksiemers:
bump:

> Are we comfortable with the existing test suite making reasonably sure that this won't break anything?

eliasjpr:
Since we changed to Struct here is being allocated to the stack memory which gets freed. 

Im going to load test locally and observe how the memory behaves with high volume of requests. I guess we should all do this tests.

eliasjpr:
```
± wrk --latency -d 25s -t 30 -c 1k http://localhost:3000/
Running 25s test @ http://localhost:3000/
  30 threads and 1000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   113.21ms   37.83ms 270.56ms   80.15%
    Req/Sec   232.47    125.11     0.97k    75.61%
  Latency Distribution
     50%  128.85ms
     75%  145.06ms
     90%  146.57ms
     99%  155.86ms
  171294 requests in 25.10s, 338.97MB read
  Socket errors: connect 0, read 1177, write 0, timeout 0
Requests/sec:   6823.50
Transfer/sec:     13.50MB
```

eliasjpr:
171,294 requests handled with no timeout

eliasjpr:
logging off
No pipes

elorest:
That's insanely slow...

PR #517, opened 2018-01-13
Prevent server side error with invalid base64 csrf token

veelenga:
### Description of the Change

It is possible to trigger a server side error (error code 500) using some wrong csrf token. Exception happens while decoding binary data using [Base64](https://crystal-lang.org/api/0.24.1/Base64.html) module:

```crystal
Base64.decode("different_token") # => Bytes[118, 39, 223, 122, 183, 167, 183, 251, 104, 145, 233]
Base64.decode("invalid_token")   # => Wrong size (Base64::Error)
                                 #       from Base64@Base64::decode<String>:Slice(UInt8)
```

This PR is about to catch such a case and correctly process invalid token (error code 403)

### DIFF
--- a/spec/amber/router/pipe/csrf_spec.cr
+++ b/spec/amber/router/pipe/csrf_spec.cr
@@ -103,7 +103,7 @@ module Amber
           end
         end
 
-        it "raises a forbidden error for not valid base64 token" do
+        it "raises a forbidden error for an invalid base64 token" do
           csrf = CSRF.new
           request = HTTP::Request.new("PUT", "/")
           context = create_context(request)
--- a/spec/amber/router/pipe/csrf_spec.cr
+++ b/spec/amber/router/pipe/csrf_spec.cr
@@ -102,6 +102,19 @@ module Amber
             csrf.call(context)
           end
         end
+
+        it "raises a forbidden error for not valid base64 token" do
+          csrf = CSRF.new
+          request = HTTP::Request.new("PUT", "/")
+          context = create_context(request)
+
+          context.session["csrf.token"] = "good_token"
+          context.params["_csrf"] = "invalid_token"
+
+          expect_raises Exceptions::Forbidden do
+            csrf.call(context)
+          end
+        end
       end
 
       context "generator" do
--- a/src/amber/router/pipe/csrf.cr
+++ b/src/amber/router/pipe/csrf.cr
@@ -66,14 +66,16 @@ module Amber
 
         def valid_token?(context)
           if request_token(context) && real_session_token(context)
-            decoded_request = Base64.decode(request_token(context).to_s) rescue nil
-            return false if decoded_request.nil? || decoded_request.size != TOKEN_LENGTH * 2
+            decoded_request = Base64.decode(request_token(context).to_s)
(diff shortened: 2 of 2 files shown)


src/amber/router/pipe/csrf.cr
```diff
@@ -66,8 +66,8 @@ module Amber
 
         def valid_token?(context)
           if request_token(context) && real_session_token(context)
-            decoded_request = Base64.decode(request_token(context).to_s)
-            return false unless decoded_request.size == TOKEN_LENGTH * 2
+            decoded_request = Base64.decode(request_token(context).to_s) rescue nil
```
elorest: My only concern is that this adds another check that gets called on every request.

I'd recommend returning on rescue instead of checking for nil.

elorest:
```cr
def valid_token?(context)
  if request_token(context) && real_session_token(context)
    decoded_request = Base64.decode(request_token(context).to_s) rescue return false
    return false if decoded_request.nil? || decoded_request.size != TOKEN_LENGTH * 2 

    unmasked = TokenOperations.unmask(decoded_request)
    session_token = Base64.decode(real_session_token(context))
    return Crypto::Subtle.constant_time_compare(unmasked, session_token)
  end 
  return false
end
```

Or better yet...

```cr
def valid_token?(context)
  if request_token(context) && real_session_token(context)
    decoded_request = Base64.decode(request_token(context).to_s)
    return false if decoded_request.nil? || decoded_request.size != TOKEN_LENGTH * 2 

    unmasked = TokenOperations.unmask(decoded_request)
    session_token = Base64.decode(real_session_token(context))
    return Crypto::Subtle.constant_time_compare(unmasked, session_token)
  end
  return false
rescue e : Base64::Error
  return false
end
```

src/amber/router/pipe/csrf.cr
```diff
@@ -73,7 +73,9 @@ module Amber
             session_token = Base64.decode(real_session_token(context))
             return Crypto::Subtle.constant_time_compare(unmasked, session_token)
           end
-          return false
+          false
+        rescue Base64::Error
```
eliasjpr: 👍

spec/amber/router/pipe/csrf_spec.cr
```diff
@@ -102,6 +102,19 @@ module Amber
             csrf.call(context)
           end
         end
+
+        it "raises a forbidden error for not valid base64 token" do
```
robacarp: it "raises a forbidden error for an invalid base64 token"

src/amber/router/pipe/csrf.cr
```diff
@@ -73,7 +73,9 @@ module Amber
             session_token = Base64.decode(real_session_token(context))
             return Crypto::Subtle.constant_time_compare(unmasked, session_token)
           end
-          return false
+          false
```
eliasjpr: What about making this a guard clause

```
return false unless request_token(context) && real_session_token(context)
```

elorest: I feel like multiple guard clauses in a row are harder to follow, but I'm fine with that.

Issue #522, opened 2018-01-14
Directory db/migrations/ doesn't exist in default app structure


docelic:
Directory db/migrations doesn't exist in default app structure.

This causes commands like 'amber db status' to fail with:

```
Applied At                  Migration
=======================================
Error opening directory 'db/migrations': No such file or directory
```

Is this something we want to fix, by e.g. just creating the empty directory?

faustinoaq:
Maybe we can check if `db/migrations` exists first ?

Something like:

```crystal
if Dir.exists?("db/migrations")
  # execute db status
else
  # print message about running amber db create migrate
end
```

WDYT?

PR #526, opened 2018-01-14
Avoid minor error in 'db status' if there are no migrations (#522)

docelic:
Small change, according to guidelines in https://github.com/amberframework/amber/issues/522

### DIFF
--- a/src/amber/cli/commands/database.cr
+++ b/src/amber/cli/commands/database.cr
@@ -44,7 +44,12 @@ module Amber::CLI
           when "redo"
             Micrate::Cli.run_redo
           when "status"
-            Micrate::Cli.run_status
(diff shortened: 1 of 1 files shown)


eliasjpr:
We should add a test for this

src/amber/cli/commands/database.cr
```diff
@@ -44,7 +44,12 @@ module Amber::CLI
           when "redo"
             Micrate::Cli.run_redo
           when "status"
-            Micrate::Cli.run_status
+            dir = "./db/migrations"
```
eliasjpr: Make `"./db/migrations" ` a constant `MIGRATIONS_DIR="./db/migrations"`

PR #530, opened 2018-01-14
Checks the current path a project name to make sure it doesn't contain a space.

elorest:
### Description of the Change

Checks pwd and project name to determine if the path is valid or will break.
Exits with error if invalid.

Could I get some input on the error message and formatting?
```
12:02:45 Class      | Error: /Users/isaac/workspace/amber/stupid path/name must not contain a space.
Underscores are fine though.
```

### DIFF
--- a/src/amber/cli/commands/new.cr
+++ b/src/amber/cli/commands/new.cr
@@ -24,7 +24,7 @@ module Amber::CLI
         Amber::CLI.color = !options.no_color?
         full_path_name = File.join(Dir.current, args.name)
         if full_path_name =~ /\s+/
-          CLI.logger.error "Path and project name can't contain a space." 
+          CLI.logger.error "Path and project name can't contain a space."
           CLI.logger.error "Replace spaces with underscores or dashes."
           CLI.logger.error "#{full_path_name} should be #{full_path_name.gsub(/\s+/, "_")}"
           exit 1
--- a/src/amber/cli/commands/new.cr
+++ b/src/amber/cli/commands/new.cr
@@ -22,6 +22,11 @@ module Amber::CLI
 
       def run
         Amber::CLI.color = !options.no_color?
+        full_path_name = `pwd`.sub("\n", "") + "/name"
+        if full_path_name.includes?(" ")
+          puts "Error: #{full_path_name} must not contain a space.\nUnderscores are fine though."
+          exit 1
+        end
         name = File.basename(args.name)
         template = Template.new(name, "./#{args.name}")
         template.generate("app", options)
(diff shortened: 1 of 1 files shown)


src/amber/cli/commands/new.cr
```diff
@@ -22,6 +22,11 @@ module Amber::CLI
 
       def run
         Amber::CLI.color = !options.no_color?
+        full_path_name = `pwd`.sub("\n", "") + "/name"
```
marksiemers: Should this be dynamic? `"/name"`

Like `"/#{args.name}"`

elorest: Yes. It totally should. For some reason I thought that was what I had done.

src/amber/cli/commands/new.cr
```diff
@@ -22,6 +22,12 @@ module Amber::CLI
 
       def run
         Amber::CLI.color = !options.no_color?
+        full_path_name = File.join(`pwd`.sub("\n", ""), args.name)
+        if full_path_name.includes?(" ")
+          puts "Error:  Path and project name can't contain a space. Underscores are fine though."
```
eliasjpr: Extra space in `Error: Path`. Replace puts with `CLI.logger.error`

Also maybe instead of `Underscores are fine though.` write something like `Use underscores or dash instead of spaces.`

```
CLI.logger.error "Error: Path and project name can't contain a space." 
CLI.logger.error "Replace spaces with underscores or dashes." (Replace/Use)
```

src/amber/cli/commands/new.cr
```diff
@@ -22,6 +22,12 @@ module Amber::CLI
 
       def run
         Amber::CLI.color = !options.no_color?
+        full_path_name = File.join(`pwd`.sub("\n", ""), args.name)
+        if full_path_name.includes?(" ")
```
epergo: Maybe you could use here a regex, just imagine the user inserts a wonderful tab character.

Using a regex:

`full_path_name =~ /s/`

faustinoaq: Can we change command `pwd` by some other native crystal method?

Something like `Dir.current`

src/amber/cli/commands/new.cr
elorest: Good suggestions.

src/amber/cli/commands/new.cr
elorest: Not sure tabs can even be in a path. Never seen that before...

Dir.current sounds good.

elorest:
@faustinoaq in six months when the new version of Crystal's released we can discuss that then.

PR #531, opened 2018-01-14
[Pipe] Add reload pipe

faustinoaq:
### Description of the Change

Add development pipeline:

```crystal
pipeline :web, :dev do
  plug ...
end

pipeline :dev do
  plug Amber::Pipe::Reload.new if Amber.env.development?
end
```

Updated: (simple way)

```crystal
pipeline :web
  # Other pipes...
  plug Amber::Pipe::Reload.new if Amber.env.development?
end
```

### Alternate Designs

No

### Benefits

Enable `Pipe::Reload` on development environments only

### Possible Drawbacks

No

### DIFF
--- a/src/amber/cli/templates/app/config/routes.cr
+++ b/src/amber/cli/templates/app/config/routes.cr
@@ -1,4 +1,4 @@
-  Amber::Server.configure do |app|
+Amber::Server.configure do |app|
   pipeline :web, :dev do
     # Plug is the method to use connect a pipe (middleware)
     # A plug accepts an instance of HTTP::Handler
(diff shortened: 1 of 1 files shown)


eliasjpr:
We did add the logic inside but something did work as expected @faustinoaq might have more insight on this

eliasjpr:
I suggested the first. I'm fine with either :)

elorest:
As projects grow routes files can become messy. I'd prefer for them not to start out that way if possible.

drujensen:
I agree with @elorest.  I think the extra `dev` pipeline is confusing.  I prefer the one liner though:
```
pipeline :web do
...
  plug Amber::Pipe::Reload.new if Amber.env.development?
end
```

PR #538, opened 2018-01-17
Rename pipe "Last" to "Controller"

docelic:
The pipe name now matches the filename in which it is defined. Also,
it no longer implies that it is the last one in the pipeline.

### DIFF
--- a/src/amber/router/pipe/controller.cr
+++ b/src/amber/router/pipe/controller.cr
@@ -3,7 +3,7 @@ module Amber
     # This is the default last pipe which is inserted automatically
     # in order to call the controller action.
 
-    class Last < Base
+    class Controller < Base
       def call(context : HTTP::Server::Context)
         context.process_request
       end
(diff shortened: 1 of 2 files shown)


docelic:
By the way, over time looks like we might switch to having an explicit Controller pipe in the pipeline in routes.cr, rather than adding it implicitly. (Because, as long as it is implicitly added to the end, it's harder to add any additional pipes after it).

drujensen:
@docelic yes, I would be a fan of that. But I think as of now, the reason it was called `Last` is that this handler is not calling `next()` and is expecting to be the last one.  It will require thinking through how to allow the request to continue.

robacarp:
It's important that Amber does have an "end" of the pipeline. If moving the Controller pipe into the routes file happens, I can get on board with that but Amber should be explicit about "something is wrong, you hit the end of the pipeline and nothing was rendered"

Issue #171, opened 2017-08-12
Provide ability to configure pipes at initialization


drujensen:
Some of the pipes have configuration settings.  Allow these setting to be initialized at creation.

marksiemers:
@drujensen - Do you have an idea how to implement this?

marksiemers:
@drujensen and @elorest - Please confirm that either this is done, or what work is left to complete this feature.

eliasjpr:
There is work still to be done for a couple of pipes

docelic:
https://github.com/amberframework/amber/pull/539

eliasjpr:
This is awesome! Thank you for all the work @docelic

drujensen:
Closing this.  I believe this is now complete.

PR #539, opened 2018-01-17
Ability to configure 'cors' pipe at initialization (#171)

### LINKED ISSUE #540: Improvements needed to 'cors' pipe
In the 'cors' pipe, at the top we have:

        ...
        @allow_headers = "accept, content-type",
        @allow_methods = "GET, HEAD, POST, DELETE, OPTIONS, PUT, PATCH",

These two variables being strings is a problem because later, methods and header names are checked for being "allowed" by doing allow_headers.includes?(word).

Since includes? on Strings includes substrings, this incorrectly makes e.g. header name "acc" be allowed because that's a substring of "accept".

It seems like the easy fix here would be to just turn those two to arrays and the rest would work as expected. (Unless we do not actually want these configurable, in which case they should be constants.)

docelic:
Makes 'cors' pipe configurable at initialization.
Also removes one unused variable.
Specs are included.

Other pipes will be modified and added to this PR.
When all pipes are checked, I will follow up with the final comment.

fixes #540

### DIFF
--- a/src/amber/router/pipe/cors.cr
+++ b/src/amber/router/pipe/cors.cr
@@ -5,12 +5,26 @@ module Amber
       property allow_origin, allow_headers, allow_methods, allow_credentials,
         max_age
 
+      ALLOW_METHODS = %w(GET HEAD POST DELETE OPTIONS PUT PATCH)
+      ALLOW_HEADERS = %w(accept content-type)
+
       def initialize(
         @allow_origin = "*",
-        @allow_headers = "accept, content-type",
-        @allow_methods = "GET, HEAD, POST, DELETE, OPTIONS, PUT, PATCH",
+        @allow_methods = ALLOW_METHODS,
+        @allow_headers = ALLOW_HEADERS,
         @allow_credentials = false,
         @max_age = 0)
+      end
+
+      def initialize(
+        @allow_origin = "*",
+        allow_methods : String = ALLOW_METHODS.join(", "),
+        allow_headers : String = ALLOW_HEADERS.join(", "),
+        @allow_credentials = false,
+        @max_age = 0)
+
+        @allow_methods = allow_methods.split /[\s,]+/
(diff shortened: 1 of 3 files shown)


docelic:
Other pipes look good.

So the only remaining issue is about cors and csrf pipes:
- cors has been modified in the first commit, but regardless of that, it relies on instance variables
- csrf does not depend on instance variables and uses constants instead

Let's figure out whether we want those cors' and csrf's params to be configurable or not.
(If yes, csrf will be modified to use instance vars, if not cors will be modified to use constants.)

src/amber/router/pipe/pipeline.cr
```diff
@@ -6,8 +6,7 @@ module Amber
       getter pipeline
       getter valve : Symbol
 
-      def initialize
-        @valve = :web
+      def initialize( @valve = :web)
```
robacarp: looks like a rogue space here

src/amber/router/pipe/cors.cr
```diff
@@ -5,12 +5,12 @@ module Amber
...
-      def initialize
-        @allow_origin = "*"
-        @allow_headers = "accept, content-type"
-        @allow_methods = "GET, HEAD, POST, DELETE, OPTIONS, PUT, PATCH"
-        @allow_credentials = false
-        @max_age = 0
+      def initialize(
+        @allow_origin = "*",
+        @allow_headers = "accept, content-type",
+        @allow_methods = "GET, HEAD, POST, DELETE, OPTIONS, PUT, PATCH",
```
eliasjpr: Can you please make the following change

```
ALLOW_METHODS = %w(GET HEAD POST DELETE OPTIONS PUT PATCH)
ALLOW_HEADERS = %w(accept content-type)

@allow_headers = ALLOW_HEADERS,
@allow_methods = ALLOW_METHODS
```

This change should address the following issue https://github.com/amberframework/amber/issues/540

PR #541, opened 2018-01-17
Move "X-Powered-By: Amber" header to its own pipe

docelic:
Moves automatically-added "X-Powered-By: Amber" response header into its own pipe.

Can be enabled/disabled at will, with the enabled state being default to not cause any visible differences compared to previous behavior.

### DIFF
--- a/spec/amber/router/pipe/powered_by_amber_spec.cr
+++ b/spec/amber/router/pipe/powered_by_amber_spec.cr
@@ -11,13 +11,13 @@ module Amber
         end
 
         Amber::Server.router.draw :web do
-          options "/test", HelloController, :world
+          options "/poweredbyamber", HelloController, :world
         end
 
         pipeline.prepare_pipelines
 
         it "should contain X-Powered-By in response" do
-          request = HTTP::Request.new("OPTIONS", "/test")
+          request = HTTP::Request.new("OPTIONS", "/poweredbyamber")
           response = create_request_and_return_io(pipeline, request)
 
           response.status_code.should eq 200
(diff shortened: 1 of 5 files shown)


docelic:
Actually, thinking about it, maybe this needs to be a general Headers plug, which adds or removes configured headers. Like:

plug Amber::Pipe::Headers(add: { "X-Powered-By": "Amber" }, delete: ...)

spec/amber/router/pipe/powered_by_amber_spec.cr
```diff
@@ -0,0 +1,29 @@
...
+    describe PoweredByAmber do
+      context "Adds X-Powered-By: Amber to response" do
+        pipeline = Pipeline.new
+
+        pipeline.build :web do
+          plug PoweredByAmber.new
+        end
+
+        Amber::Server.router.draw :web do
+          options "/test", HelloController, :world
```
faustinoaq: Travis is failing:

```
Route: OPTIONS /test is duplicated. (Amber::Exceptions::DuplicateRouteError)
```

`/test` is already in use. Please use another path here, something like `/poweredbyamber`

docelic:
Great, fixed the route.
Given the specific importance of this header/plug, it seems to be we can commit it as is, and then work on Headers plug as part of separate, more general functionality.

src/amber/router/pipe/powered_by_amber.cr
```diff
@@ -0,0 +1,11 @@
+module Amber
+  module Pipe
+    # Adds header "X-Powered-By: Amber" to the response
```
eliasjpr: We can remove the comment since the class name is explicit

drujensen:
@faustinoaq looks like the specs passed.  Can you give your blessing?

PR #545, opened 2018-01-19
[Specs] Spec cleanup

eliasjpr:
- Breaks controller specs into smaller parts
- Removes duplicate settings spec
- Move helper method to Controller helpers


spec/amber/environment/env_spec.cr
```diff
@@ -24,7 +24,7 @@ module Amber::Environment
         amber_env.{{env.id}}.should_not be_nil
       end
     end
-  {% end %}
+    {% end %}
```
marksiemers: Is this meant to be moved over?

spec/amber/controller/respond_width_spec.cr
```diff
@@ -0,0 +1,107 @@
+require "../../../spec_helper"
```
marksiemers: Typo in the filename `spec/amber/controller/respond_width_spec.cr`

spec/amber/environment/env_spec.cr
eliasjpr: Yes, to align the with the opening if. 

I just noticed that this file is named incorrectly

eliasjpr:
@marksiemens they are modules already and those unit/integration specs. Does it make sense? I can improve the specs

PR #548, opened 2018-01-19
Fix conflicting colored_server_name on deploys

### LINKED ISSUE #546: Underscore is not a valid character for DigitalOcean machines.
Hi!

Whenever we try to create a machine on digitalOcean that contains `_` it will fail

https://github.com/amberframework/amber/blob/211dd91b3aac9832a4b42fd8e29e9d7d9f9d2a65/src/amber/cli/commands/deploy.cr#L38

```rb
# We could only remove `|_` from the regexp:
@server_name = "...".gsub(/[^\w\d\-]/, "")
```

Is there any reason to keep underscores on the server name?

gabrielengel:
### Description of the Change

Fixes DigitalOcean's deployment. Caused by a colored string being sent to the docker command.

Also rules off underscores on machine names (not accepted by DO).

fixes #546

### DIFF
--- a/src/amber/cli/commands/deploy.cr
+++ b/src/amber/cli/commands/deploy.cr
@@ -35,7 +35,7 @@ module Amber::CLI
         (spinner = Spin.new(0.1, Spinner::Charset[:arrow2].map { |c| c.colorize(:light_green).to_s })).start
         shard = YAML.parse(File.read("./shard.yml"))
         @project_name = shard["name"].to_s
-        @server_name = "#{project_name}-#{args.server_suffix}".gsub(/[^\w\d\-]|_/, "")
+        @server_name = "#{project_name}-#{args.server_suffix}".gsub(/[^\w\d]/, "-")
         if options.init?
           provision
         else
@@ -47,7 +47,7 @@ module Amber::CLI
       end
 
       private def provision
-        puts "Provisioning server #{server_name}."
+        puts "Provisioning server #{colored_server_name}."
         create_cloud_server
         create_swapfile
         create_deploy_keys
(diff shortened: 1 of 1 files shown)


src/amber/cli/commands/deploy.cr
```diff
@@ -35,7 +35,7 @@ module Amber::CLI
         (spinner = Spin.new(0.1, Spinner::Charset[:arrow2].map { |c| c.colorize(:light_green).to_s })).start
         shard = YAML.parse(File.read("./shard.yml"))
         @project_name = shard["name"].to_s
-        @server_name = "#{project_name}-#{args.server_suffix}".gsub(/[^\w\d\-]|_/, "")
+        @server_name = "#{project_name}-#{args.server_suffix}".gsub(/[^\w\d\-]/, "")
```
robacarp: It seems like this regex should be: `/[^\w\d|_-]/`? 

Actually, this is strange because `"#{project_name}-#{args.server_suffix}"` has a `-` in it, which the gsub is just going to delete.

gabrielengel:
That regexp will delete everything except words, digits and hifens, will it
not?

We could also accept dots, yet not sure other hosts will deal properly.

Em 19 de jan de 2018 7:24 PM, "robacarp" <notifications@github.com>
escreveu:

src/amber/cli/commands/deploy.cr
eliasjpr: I also think it should replace the findings with `-` to make it readable.

gabrielengel:
So "one:weird*string" will become "one-weird-string". Sounds good!

Em 19 de jan de 2018 8:23 PM, "Elias Perez" <notifications@github.com>
escreveu:

eliasjpr:
Issue https://github.com/amberframework/amber/issues/546

gabrielengel:
Ready for review

PR #550, opened 2018-01-20
Add pipelines command

NeverHappened:
### Description of the Change
Added 'pipelines' command. It shows the pipelines and the plugs for them

Also have --no-plugs option if you want to just see the pipelines

fixes #550

### DIFF
--- a/spec/amber/cli/commands/pipelines/pipelines_spec.cr
+++ b/spec/amber/cli/commands/pipelines/pipelines_spec.cr
@@ -36,8 +36,17 @@ module Amber::CLI
           Amber::Pipe::Static
         )
 
+        pipe_plugs = {
+          "web"    => web_default_plugs,
+          "static" => static_default_plugs,
+        }
 
         describe "with the default routes" do
+          it "runs without an error" do
+            MainCommand.run %w(pipelines) { |cmd| output = cmd.out.gets_to_end }
+            output.should_not contain "Good bye :("
+          end
+
           MainCommand.run %w(pipelines) { |cmd| output = cmd.out.gets_to_end }
           output_lines = route_table_rows(output)
 
@@ -60,13 +69,14 @@ module Amber::CLI
             end
           end
 
-          it "maintains the correct order of pipelines and plugs" do
-            in_correct_order = %w(Plug) + %w(web) + web_default_plugs + %w(static) + static_default_plugs
-
-            output_lines
-            .reject { |line| line.empty? }
-            .each_with_index do |line, index|
-              line.should contain (in_correct_order[index])
+          it "saves the order of the plugs in the pipeline" do
+            pipe_plugs.each do |pipe_name, plugs|
+              output_plugs = output_lines.select { |line| line.includes?(pipe_name) }
+              plugs.each_with_index do |plug, index|
+                output_plug = output_plugs[index]
+                (output_plug != nil).should be_true
+                output_plug.should contain plug
+              end
             end
           end
         end
--- a/spec/amber/cli/commands/pipelines/pipelines_spec.cr
+++ b/spec/amber/cli/commands/pipelines/pipelines_spec.cr
@@ -1,0 +1,61 @@
+require "../../../../spec_helper"
+require "../../../../support/helpers/cli_helper"
+
+require "cli/spec"
+
+include CLIHelper
+include Cli::Spec::Helper
+
+module Amber::CLI
+  extend Helpers
+
+  describe "amber pipelines" do
+    begin
+      ENV["CRYSTAL_CLI_ENV"] = "test"
+
+      context "in an `amber new` with default options" do
+        cleanup
+        scaffold_app(TESTING_APP)
+        output = ""
+
+        describe "with the default routes" do
+          MainCommand.run %w(pipelines) { |cmd| output = cmd.out.gets_to_end }
+          output_lines = route_table_rows(output)
+
+          it "outputs the correct headers" do
+            headers = %w(Pipe Plug)
+            headers.each do |header|
+              output.should contain header
+            end
+          end
+
+          it "outputs the default pipeline names" do
+            pipeline_names = %w(web static)
+            pipeline_names.each do |pipeline|
+              output.should contain pipeline
+            end
+          end
+
+          it "outputs the plugs for the pipelines" do
+            plugs = %w(
+              Amber::Pipe::Error
+              Amber::Pipe::Logger
+              Amber::Pipe::Session
+              Amber::Pipe::Flash
+              Amber::Pipe::CSRF
+              Amber::Pipe::Reload
+              Amber::Pipe::Error
+              Amber::Pipe::Static
+            )
+
+            plugs.each do |plug|
+              output.should contain plug
+            end
+          end
+        end
+      end
+    ensure
+      cleanup
+    end
+  end
+end
--- a/spec/amber/cli/commands/pipelines/pipelines_spec.cr
+++ b/spec/amber/cli/commands/pipelines/pipelines_spec.cr
@@ -18,6 +18,25 @@ module Amber::CLI
         scaffold_app(TESTING_APP)
         output = ""
 
+        pipeline_names = %w(web static)
+
+        web_default_plugs = %w(
+          Amber::Pipe::PoweredByAmber
+          Amber::Pipe::Error
+          Amber::Pipe::Logger
+          Amber::Pipe::Session
+          Amber::Pipe::Flash
+          Amber::Pipe::CSRF
+          Amber::Pipe::Reload
+        )
+
+        static_default_plugs = %w(
+          Amber::Pipe::PoweredByAmber
+          Amber::Pipe::Error
+          Amber::Pipe::Static
+        )
+
+
         describe "with the default routes" do
           MainCommand.run %w(pipelines) { |cmd| output = cmd.out.gets_to_end }
           output_lines = route_table_rows(output)
@@ -30,26 +49,24 @@ module Amber::CLI
           end
 
           it "outputs the default pipeline names" do
-            pipeline_names = %w(web static)
             pipeline_names.each do |pipeline|
               output.should contain pipeline
             end
           end
 
           it "outputs the plugs for the pipelines" do
-            plugs = %w(
-              Amber::Pipe::Error
-              Amber::Pipe::Logger
-              Amber::Pipe::Session
-              Amber::Pipe::Flash
-              Amber::Pipe::CSRF
-              Amber::Pipe::Reload
-              Amber::Pipe::Error
-              Amber::Pipe::Static
-            )
-
-            plugs.each do |plug|
+            (web_default_plugs + static_default_plugs).each do |plug|
               output.should contain plug
+            end
+          end
+
+          it "maintains the correct order of pipelines and plugs" do
+            in_correct_order = %w(Plug) + %w(web) + web_default_plugs + %w(static) + static_default_plugs
+
+            output_lines
+            .reject { |line| line.empty? }
+            .each_with_index do |line, index|
+              line.should contain (in_correct_order[index])
             end
           end
         end
--- a/src/amber/cli/commands/pipelines.cr
+++ b/src/amber/cli/commands/pipelines.cr
@@ -14,27 +14,38 @@ module Amber::CLI
         help
       end
 
-      LABELS = %w(Pipe Plug)
+      class BadRoutesException < Exception
+      end
+
+      ROUTES_PATH          = "config/routes.cr"
+      LABELS               = %w(Pipe Plug)
       LABELS_WITHOUT_PLUGS = %w(Pipe)
+
       PIPE_REGEX = /(pipeline)\s+\:(\w+)(?:,\s+\"([^\"]+)\")?/
       PLUG_REGEX = /(plug)\s+([\w:]+)?/
+
+      FAILED_TO_PARSE_ERROR = "Could not parse pipeline/plugs in #{ROUTES_PATH}"
 
       command_name "pipelines"
 
       def run
         parse_routes
         print_pipelines
+      rescue ex : BadRoutesException
+        CLI.logger.error(ex.message.colorize(:red))
+        CLI.logger.error "Good bye :("
+        exit 1
       rescue
-        puts "Error: Not valid project root directory.".colorize(:red)
-        puts "Run `amber pipelines` in project root directory.".colorize(:light_blue)
-        puts "Good bye :("
+        CLI.logger.error "Error: Not valid project root directory.".colorize(:red)
+        CLI.logger.error "Run `amber pipelines` in project root directory.".colorize(:light_blue)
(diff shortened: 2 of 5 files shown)


src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,92 @@
...
+      PIPE_REGEX = /(pipeline)\s+\:(\w+)(?:,\s+\"([^\"]+)\")?/
+      PLUG_REGEX = /(plug)\s+([\w:]+)?/
+
+      command_name "pipelines"
+
+      def run
+        parse_routes
+        print_pipelines
+      rescue
+        puts "Error: Not valid project root directory.".colorize(:red)
```
eliasjpr: Use `CLI.logger.error` instead of puts

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,92 @@
...
+      # TODO: use hash with fixed order?
+      getter result = Hash(String?, Array(String)).new
+      property current_pipe : String?
+
+      class Options
+        bool "--no-color", desc: "Disable colored output", default: false
+        help
+      end
+
+      LABELS     = ["Pipe", "Plug"]
```
eliasjpr: Use `%w(Pipe Plug)` instead

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,92 @@
+require "cli"
+require "shell-table"
+require "../helpers/sentry"
+
+module Amber::CLI
+  class MainCommand < ::Cli::Supercommand
+    class Pipelines < Command
+      # TODO: use hash with fixed order?
```
eliasjpr: It would be nice to print the pipes in the same order they are used and maybe somehow reflect that in the output

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,92 @@
...
+        print_pipelines
+      rescue
+        puts "Error: Not valid project root directory.".colorize(:red)
+        puts "Run `amber pipelines` in project root directory.".colorize(:light_blue)
+        puts "Good bye :("
+        exit 1
+      end
+
+      private def parse_routes
+        lines = File.read_lines("config/routes.cr")
```
eliasjpr: Can we make `"config/routes.cr"` a constant

src/amber/cli/commands/pipelines.cr
marksiemers: Agreed, they should display in order - maybe that is already happening. The test should reflect that.

spec/amber/cli/commands/pipelines/pipelines_spec.cr
```diff
@@ -0,0 +1,61 @@
...
+              Amber::Pipe::Session
+              Amber::Pipe::Flash
+              Amber::Pipe::CSRF
+              Amber::Pipe::Reload
+              Amber::Pipe::Error
+              Amber::Pipe::Static
+            )
+
+            plugs.each do |plug|
+              output.should contain plug
```
marksiemers: I can't immediately think how to do this, but there should be a test to make sure the plugs are in the correct order.

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,92 @@
...
+          when .starts_with?("pipeline")
+            set_pipe(line)
+          when .starts_with?("plug")
+            set_plug(line)
+          end
+        end
+      end
+
+      private def set_pipe(line)
+        match = line.match(PIPE_REGEX)
```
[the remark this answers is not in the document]
marksiemers: See comment below with a complete proposed solution.

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,92 @@
...
+          @current_pipe = match[2]
+
+          if @current_pipe
+            result[@current_pipe] = [] of String
+          end
+        end
+      end
+
+      private def set_plug(line)
+        match = line.match(PLUG_REGEX)
```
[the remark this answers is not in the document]
marksiemers: See comment below with a complete proposed solution.

spec/amber/cli/commands/pipelines/pipelines_spec.cr
NeverHappened: Yeah, I though of that. Just that didn't come up to a solution right away. I'll think about it

src/amber/cli/commands/pipelines.cr
NeverHappened: Yeah, I did it, just forgot to push last updates)

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,92 @@
...
+        if match
+          # TODO: raise if no match?
+          plug = match[2]
+
+          if @current_pipe # TODO: maybe throw exception if no current pipe
+            result[@current_pipe] << plug
+          end
+        end
+      end
```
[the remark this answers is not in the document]
NeverHappened: Okay, I'll try to put more work on error handling.

eliasjpr:
@marksiemers I thought we agree not raising errors with the CLI but printing an error with a precise msg msg

marksiemers:
@eliasjpr - I either forgot that or wasn't there for the convo.

Yeah, not raising an error is fine with me, but I think then reporting a message and existing would go where `raise ...` is now. WDYT?

eliasjpr:
Agree the cli has an exit method or simple log.error will suffice

NeverHappened:
Fixed the stuff you requested. I did the test for ensuring the order. The only problem is that test will fail if you change the default routes. If there is a better way to test please tell me.

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,92 @@
...
+        lines = lines.map { |line| line.strip }
+        lines.each do |line|
+          case line
+          when .starts_with?("pipeline")
+            set_pipe(line)
+          when .starts_with?("plug")
+            set_plug(line)
+          end
+        end
+      end
```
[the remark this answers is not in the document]
eliasjpr: I have no preference honestly

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,98 @@
...
+
+      class BadRoutesException < Exception
+      end
+
+      ROUTES_PATH          = "config/routes.cr"
+      LABELS               = %w(Pipe Plug)
+      LABELS_WITHOUT_PLUGS = %w(Pipe)
+
+      PIPE_REGEX = /(pipeline)\s+\:(\w+)(?:,\s+\"([^\"]+)\")?/
+      PLUG_REGEX = /(plug)\s+([\w:]+)?/
```
robacarp: We've talked elsewhere about using extended regular expressions to add comments, clarity, and maintainability to our regular expressions. Can we do that here?

spec/amber/cli/commands/pipelines/pipelines_spec.cr
```diff
@@ -0,0 +1,78 @@
...
+          end
+
+          it "outputs the plugs for the pipelines" do
+            (web_default_plugs + static_default_plugs).each do |plug|
+              output.should contain plug
+            end
+          end
+
+          it "maintains the correct order of pipelines and plugs" do
+            in_correct_order = %w(Plug) + %w(web) + web_default_plugs + %w(static) + static_default_plugs
```
robacarp: Generally I'm all in for `%w` syntax but this feels weird

eliasjpr: I left a comment previously that caused this. @NeverHappened my apologies I was not explicit on my comments.

src/amber/cli/commands/pipelines.cr
marksiemers: @robacarp - I remember seeing your example in a previous PR, but I've never written an extended regex. Is this the idea?
```crystal
 /
   (plug)    # match and capture the word plug
   \s+       # followed by at least one whitespace character
   ([\w:]+)? # greedy optional match and capture one or more word characters or colons ':'
 /x
```

@NeverHappened - Does this description match what the regex is doing?

This was a fun little exercise, assuming the above is accurate, I would suggest a few changes:
```crystal
 /
   plug\s+  # match 'plug` and at least one whitespace character
   ([\w:]+) # match and capture one or more word characters or colons ':'
 /x
```

marksiemers: I messed around a little more, WDYT of this:
```crystal
PLUG_REGEX = 
  /^         # start at beginning of line
    plug\s+  # match 'plug` and at least one whitespace character
    ([\w:]+) # match and capture one or more word characters or colons ':'
   $         # matches must come before end of line
  /x
```

See https://carc.in/#/r/3fhw

robacarp: @marksiemers that's the right idea! The only functional difference between regular regex and this is that whitespace is basically ignored outside of character classes, and that comments are allowed. I believe it adds a major maintainability component. :]

marksiemers: Indeed, and it makes you think about what you are doing.

I used to teach at a bootcamp and we always told people - you'll never be able to read a regular expression. When you come back to it in 6 months, it will take 20 minutes to figure out - like starting from scratch. If only we had known about this!

eliasjpr: @marksiemers and @robacarp we can probably add a commit with these changes, this way we can have it merge sooner

marksiemers: The pipeline one I'm giving up on for now, but you can see a WIP here: https://carc.in/#/r/3fiv

marksiemers: If this works for the plug, I would say swap this in with this PR, as an example and a reminder:
```crystal
PLUG_REGEX = 
  /^         # start at beginning of line
    plug\s+  # match 'plug` and at least one whitespace character
    ([\w:]+) # match and capture one or more word characters or colons ':'
  $/x        # matches must come before end of line
```

marksiemers:
@NeverHappened - Before merging this I'd like the see the following:
1. (Required) List the pipeline name on each row with each plug (see table below)

1. (Required) Modify tests to cover format shown above

1. (Strongly encouraged) Refactor regex for plugs using extended regex (see comments above)

1. (Optional) Refactor `parse_routes` per suggestion

Sample table:

| Pipeline            | Plug                           |
| -------------- |----------------------|
| web                 | Amber::Pipe::Error     |
| web                 | Amber::Pipe::Logger  | 
| ...                 | ...  | 
| static                 | Amber::Pipe::Error     |
| static                 | Amber::Pipe::Static  | 

@eliasjpr @drujensen @robacarp - Agreed?

drujensen:
@marksiemers I agree with your requirement list.  I think the table output makes it easy to grep as well `amber pipelines | grep web`

@NeverHappened are you ok with the requested changes?

NeverHappened:
Yeah, of course! Thanks for the help :) I'll do the changes as soon as I can.

My only concern is the regular expressions are not 100% robust.
Isn't there a way to wait for the amber to load to fetch real pipes and plugs?

drujensen:
@NeverHappened I believe at one point we implemented the routes that way, but performance wise, we found the RegEx's faster and just as accurate.

marksiemers:
Yeah, I believe loading them from the actual application may take seconds and longer as the app grows bigger. The regex approach is much much faster.

It does seem a little strange to have a second system for reporting the routes, but so far it has worked well.

eliasjpr:
My only concern with the regex is that people can register routes outside of the routes file (This actually has been done) and those routes will not be reported by the command.

NeverHappened:
Yeah, and it's not that uncommon. For example, for projects in rails I often split admin and user routes in separate files.

robacarp:
This would be a bit bigger of a lift, but Crystal files can compile pretty fast if there isn't a big project. It might be possible to make a tiny compilation environment which requires in the routes file and actually evaluates it, and spit out the output. 

As for multiple routes files, I've done that in many many Rails projects and I've always been surprised that Rails didn't provide any sort of helper function for `draw`ing more routes from a different file into the router. Should that be on the roadmap for Amber?

eliasjpr:
You can define routes in a different file other than routes if you follow the same `Amber::Server.configure do |app|` and making sure is require in `config/application.cr`

marksiemers:
I agree that `amber routes` should be handled in a different way. The regex is a bit of a hack.

I also think that is outside the scope of this issue: `Add pipelines command`

A separate issue should be opened for having the CLI evaluate the actual routes, rather than the code.

eliasjpr:
Can we get the final touches to this PR so we can merge it. We would like to release soon

NeverHappened:
Yes, I'll finish it soon

NeverHappened:
Done with what you were asking for!

Note: 
- I didn't sort the pipes in the output. So the order of plugs IS correct, but pipes/plugs can be mixed dependent on the routes.cr file. (Can sort by pipes, if that's critical)
- There is no way to capture contiguous patterns like "api_v1", "api_v2", "api_v2", etc.. That's why It was necessary to capture entire group and split it later

Have some questions, I would be happy if you can answer.
1) why crystal spec ./path/to/pipeline_spec.cr shows a lot more specs than it's really in the file?
2) If the spec fails because of a bad regex, it just didn't execute. I understand that it's not that good, but didn't see why this happens.

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,132 @@
...
+          )         # match and capture all contiguous words
+        /x
+
+      PLUG_REGEX =
+        /^
+          \s*
+          plug        # match 'plug'
+          \s+         # require at least one whitespace character after 'plug'
+          (
+            [\w:]+    # match at least one words with maybe a colon
```
robacarp: @NeverHappened Thank you for doing this, it increases maintainability so much!

src/amber/cli/commands/pipelines.cr
```diff
@@ -0,0 +1,132 @@
...
+          )?
+          (?:
+            [\.\s*\(] # until we reach '.', spaces, or braces
+          )?
+        /x
+
+      FAILED_TO_PARSE_ERROR = "Could not parse pipeline/plugs in #{ROUTES_PATH}"
```
faustinoaq: @NeverHappened Looks like syntax highlighting is broken here, I tried to remove quotes (`'`) in comments and syntax highlighting worked again:

With quotes (`'`) in comments

screenshot_20180131_112412

Without quotes (`'`) in comments:

screenshot_20180131_112500

NeverHappened:
@faustinoaq Fixed

robacarp:
@faustinoaq that's probably an issue with the syntax highlighter rules in your editor

faustinoaq:
> that's probably an issue with the syntax highlighter rules in your editor

Not just my editor, but github diff view itself 👇 

screenshot_20180131_154305

Look at `FAILED_TO_PARSE_ERROR` ^

PR #552, opened 2018-01-20
[Authors] Sets the list of authors

eliasjpr:
Generalizes the list of authors to Amber Core Team and Contributors
this lowers maintenance of this list and still attributes everyone who in some way has contributed.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -3,11 +3,7 @@ name: amber
 version: 0.6.4
 
 authors:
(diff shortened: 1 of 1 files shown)


faustinoaq:
Maybe we can use this if we can add an amber email, something like:

```yml
authors:
  - Amber Team and collaborators - <team@amberframework.org>
```

Issue #553, opened 2018-01-20
[CLI] Redesign Inclusion of DB-related Shards


docelic:
Currently, a new project which defaults to Pg still depends on all DB shards and they are downloaded when one runs 'shards', including the ones not needed/used:

redis
mysql
sqlite3

This should be sorted out and according to @marksiemers' advice, maybe these should be on app level rather than framework level.

marksiemers:
This will require having conditional `require` statements in some places, not sure what else.

Example: https://github.com/amberframework/amber/blob/master/src/amber/cli/commands/database.cr#L4

faustinoaq:
> This will require having conditional require statements in some places, not sure what else

Yeah, a conditional `require` would work here, we should use macros though

```crystal
{% if database == "pg" %}
require "pg"
{% elsif database == "mysql" %}
require "mysql"
{% elsif database == "sqlite" %}
require "sqlite3"
{% end %}
```

eliasjpr:
Was this resolved?

elorest:
@faustinoaq @marksiemers require is a before macro operation. You can't conditionally do requires with macros.

marksiemers:
I don't think this should be closed. The macro solution may not work, but something could.

The require statements are done in the CLI, which does make it a challenge - maybe some kind of dependency injection could work, or somehow offloading the operations and dependencies to the app itself.

It may be a for a far future version, but we shouldn't drop it.

eliasjpr:
👍🏻 I closed it thinking this was addressed in he above PR

faustinoaq:
> require is a before macro operation. You can't conditionally do requires with macros.

I tested some code, and I think we can try something like this:

```crystal
VAR = "no"  # Try changing this to "yes"

{% if VAR == "yes" %}

require "tempfile"

pp Tempfile.new("foo").path

{% elsif VAR == "no" %}

require "colorize"

pp "foo".colorize(:green)

{% end %}
```

https://carc.in/#/r/3k9q

Or maybe we can use some kind of ECR template to generate db-related files, WDYT?

marksiemers:
Based on some playing in carc.in a macro will change what is required: https://carc.in/#/r/3khf

eliasjpr:
@marksiemers and @faustinoaq can you guys take a shot at addressing this in a PR? I think this can be an easy win :)

marksiemers:
[...]

PR #556, opened 2018-01-20
move to development dependencies

drujensen:
### Description of the Change

Addresses issue #553 

We no longer require the db drivers because we removed building amber `on the fly` for docker.  We should investigate what other libraries are `development_dependencies` with this change.

### Benefits

Reduces shard dependencies and install time.

### Possible Drawbacks

None

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -54,6 +54,7 @@ dependencies:
     github: askn/spinner
     version: ~> 0.1.1
 
+development_dependencies:
   pg:
     github: will/crystal-pg
     version: ~> 0.14.1
--- a/shard.yml
+++ b/shard.yml
@@ -34,6 +34,7 @@ dependencies:
     github: stefanwille/crystal-redis
     version: ~> 1.9.0
 
+development_dependencies:
   cli:
     github: amberframework/cli
     version: ~> 0.7.0
@@ -54,7 +55,6 @@ dependencies:
     github: askn/spinner
     version: ~> 0.1.1
 
-development_dependencies:
   pg:
     github: will/crystal-pg
     version: ~> 0.14.1
@@ -67,7 +67,6 @@ development_dependencies:
     github: crystal-lang/crystal-sqlite3
     version: ~> 0.9.0
 
-development_dependencies:
   ameba:
     github: veelenga/ameba
     version: ~> 0.4.0


faustinoaq:
I think We can remove [this line](https://github.com/amberframework/amber/pull/556/files#diff-12a20bd44c6d0361697e8dce350c26d2R70)

shard.yml
```diff
@@ -54,6 +54,7 @@ dependencies:
     github: askn/spinner
     version: ~> 0.1.1
 
+development_dependencies:
```
eliasjpr: There is two branches of development dependencies I we can group all Dev deos into one

drujensen:
I moved more shards to `development_dependencies`:

I believe we depend on `radix`, `kilt` and `slang`.  I think `redis` can be moved if we only add it to the project shard when someone generates a WebSocket.

elorest:
@eliasjpr Also wrote a redis store for sessions. But they could add it in if they need that.

shard.yml
```diff
@@ -34,6 +34,7 @@ dependencies:
     github: stefanwille/crystal-redis
     version: ~> 1.9.0
 
+development_dependencies:
   cli:
```
faustinoaq: So, I need to remove `--production` flag [here](https://github.com/amberframework/amber-aur/blob/master/PKGBUILD#L15), Is the cli a development dependency only?

drujensen: @faustinoaq Yes, the CLI should be used for development only.  

Is the `amber` command required for heroku deployment?  If so, can we use the docker image instead?

drujensen: @faustinoaq just remembered to run the migrations in production, we need the amber cli.  Lets hold off on this PR until we figure out how to do deployments without requiring to build `amber`.

drujensen:
@faustinoaq  I tested and you're right, the heroku build pack requires all of the shards and failed:
```
remote: -----> Amber CLI not compiled... -----> Installing amber CLI... Fetching https://github.com/luislavena/radix.git
remote: Fetching https://github.com/jeromegn/kilt.git
remote: Fetching https://github.com/jeromegn/slang.git
remote: Fetching https://github.com/stefanwille/crystal-redis.git
remote: Fetching https://github.com/amberframework/cli.git
remote: Fetching https://github.com/mosop/optarg.git
remote: Fetching https://github.com/mosop/callback.git
remote: Fetching https://github.com/mosop/string_inflection.git
remote: Fetching https://github.com/amberframework/teeplate.git
remote: Fetching https://github.com/juanedi/micrate.git
remote: Fetching https://github.com/crystal-lang/crystal-db.git
remote: Fetching https://github.com/jwaldrip/shell-table.cr.git
remote: Fetching https://github.com/askn/spinner.git
remote: Fetching https://github.com/will/crystal-pg.git
remote: Fetching https://github.com/crystal-lang/crystal-mysql.git
remote: Fetching https://github.com/veelenga/ameba.git
remote: Error shard name (ameba) doesn't match dependency name (sqlite3)
remote:  !     Push rejected, failed to compile Amber app.
remote: 
remote:  !     Push failed
remote: Verifying deploy...
remote: 
remote: !	Push rejected to drusblog.
remote: 
```

drujensen:
@faustinoaq @eliasjpr @elorest I would like for us to have a discussion regarding heroku buildpack and see if we can remove the need to build amber `on the fly`.  

Maybe we can remove the buildpack and support deployment to heroku via `amber deploy`?

eliasjpr:
Honestly we just need a flag on the require that we can call from the build pack to exclude Sqlite from being require in the `cli/commands/database.yml` so we can do something similar the way we run the build tests. 

```
 crystal build --no-debug --stats -D --no-sqlite
```

drujensen:
Closing this PR.  I think we may want to leave the shards in so we can build on the fly.  It looks like this is needed for other reasons as well.

PR #559, opened 2018-01-21
ECR template fixes

drujensen:
### Description of the Change

There were several small bugs found in the ECR templates. 

Note: ECR templates do not support implicit `do end` blocks.  This causes issues with some of the helper methods we have.  A work around was to inline the block.  We can't do this for the form helper though.

### DIFF
--- a/src/amber/cli/templates/app/src/views/layouts/application.ecr.ecr
+++ b/src/amber/cli/templates/app/src/views/layouts/application.ecr.ecr
@@ -12,7 +12,7 @@
     <div class="masthead">
       <div class="container">
         <nav class="nav">
-          <%="<"%>%= render(partial: "layouts/_nav.<%= @language %>") %>
+          <%="<"%>%= render(partial: "layouts/_nav.ecr") %>
         </nav>
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


src/amber/cli/templates/app/src/views/layouts/application.ecr.ecr
```diff
@@ -12,7 +12,7 @@
     <div class="masthead">
       <div class="container">
         <nav class="nav">
-          <%="<"%>%= render(partial: "layouts/_nav.<%= @language %>") %>
+          <%="<"%>%= render(partial: "layouts/_nav.ecr") %>
```
eliasjpr: Would this work for `slang` templates?

drujensen: This is in the ecr template so we can be specific here.

src/amber/cli/templates/auth/src/views/{{name}}/edit.ecr.ecr
```diff
@@ -8,11 +8,11 @@
   </ul>
 <%="<"%>%- end %>
 
-<%="<"%>%= form(action: "/profile", method: :patch) do %>
+<form action="/profile" method="patch">
```
elorest: Frustrating how ECR is less complete than slang.

PR #563, opened 2018-01-21
Dj/fixes to docker

drujensen:
### Description of the Change

A couple bugs in the `Dockerfile`:

1. The tag has a `v` so the image version requires this.

2. If you run `amber w` before running `docker-compose up`, the `node_modules` from the current OS gets copied to the image.  We need to remove these so that the linux OS based node_modules is properly installed.


KevinSjoberg:
For good measure, you could also replace `ADD . /app` with `COPY . /app`. `COPY` is preferred as long as you're not dealing with tar-archives or urls.

drujensen:
ready for review again.

PR #564, opened 2018-01-21
Add client IP pipe

### LINKED ISSUE #534: Improve client IP detection
As part of handling a request, Amber may need to know the IP of the client.

If Amber server sits behind a proxy, then the client IP seen is the address of the proxy, and the proxy passes the real client IP in some HTTP header, such as X-Forwarded-For.
So, that other header must be consulted to get the real client IP for most purposes (access control, display of the IP address, logging, etc.).

Amber does currently have some support for this, and it does it in the following way:
(excerpts from src/amber/router/context.cr):

```crystal
IP_ADDRESS_HEADERS = %w(REMOTE_ADDR CLIENT_IP X_FORWARDED_FOR
        X_FORWARDED X_CLUSTER_CLIENT_IP FORWARDED)

  def client_ip
    headers = request.headers
    val = {} of String => String
    IP_ADDRESS_HEADERS.find { |header|
      dashed_header = header.tr("_", "-")
      val = headers[header]? || headers[dashed_header]? || headers["HTTP_#{header}"]? || 
            headers["Http-#{dashed_header}"]?
    }
    val
  end
```

This approach has the following negative consequences:

1. If the Amber server is not behind a proxy, then these various headers should not be looked up at all. (Looking them up (and taking them without discrimination) only allows users to spoof their IP address - this is even noted in the comments above the method)
1.  If the Amber server is behind a proxy, then the proxy must do extra work to delete all these possible headers from incoming requests (a total of 24 combinations per current code, in the worst-case scenario) and only set one real header in which it will store the actual client IP
1. Even then, that header is by current Amber's code almost required to be named CLIENT_IP, because naming it any other way would cause more loops / checks to be ran on each request
1. At least some local variables are created and some tr() replacements are run in each invocation of the client_ip() method
1. It seems to me that client_ip should not be a method at all. Real client IP should be determined once and then stored somewhere in a variable or header as fixed information
1. This implementation does not take into account that, in the future, support for PROXY protocol may be added to Amber or Crystal (https://www.haproxy.com/blog/haproxy/proxy-protocol/)

My suggestion is to improve the situation as follows:

1. Allow Amber server configuration to specify the names of headers which will be checked for real client IP
1. No transformations or variations of these names are to be generated during runtime (if a user wants 24 variations of header name, they simply need to list 24 variations in step #1)
1. Default this value to empty list (no headers checked, no overhead in runtime, no conflict with PROXY protocol later)
1. If some header names are specified, then perform efficient client IP detection once, and store the value in some fixed variable or header

Time permitting, I will work on implementing this. Comments welcome, thanks.

docelic:
This PR adds better support for client IP detection in Amber.

(Some discussion about it is in https://github.com/amberframework/amber/issues/534.)

Item left for todo for later when Crystal provides access to remote IP address is to add parameter "from", or "source", or something similar, and only honor the client IP headers when requests are coming from trusted IPs (IPs of legitimate load balancers).

-------------
Comments/notes from commits:

- Remove current unconditional search for real client IP 
- Add Amber::Pipe::ClientIp
- It looks into configured header names (default: "X-Forwarded-For"),
  finds the first header that exists, and takes its last value
- Saves the value into context.client_ip as Socket::IPAddress.new(addr, 0). (The port value is currently unused and just set to 0 as required by initialize.)
- Covered by specs
- Along the way, sort in alphabetic order methods on the controller
  that are delegated to context for easier lookup and modification
- Add Amber::Pipe::ClientIp to template config/routes.cr
- By default, add it in disabled state
- Even though ClientIp has default header ("X-Forward-For"), do mention
  name explicitly in the template, and also show an array to be as
  easy and intuitive as possible for users to understand what the header
  name is and that multiple headers are accepted
--------------
People who were previously relying on built-in client IP detection and now upgrade Amber should add the following to their pipeline in config/routes.cr:

plug Amber::Pipe::ClientIp.new(%w(Header-Name1 Header-Name2))

And then replace eventual calls to `context.client_ip` with `context.client_ip.address`.

 fixes #534


src/amber/pipes/client_ip.cr
```diff
@@ -0,0 +1,34 @@
...
+      def initialize(header : String = "X-Forwarded-For")
+        @headers = [header]
+      end
+      def initialize( @headers : Array(String))
+      end
+
+      def call(context : HTTP::Server::Context)
+        @headers.each do |header|
+          if addresses = context.request.headers.get?(header)
+            context.client_ip = Socket::IPAddress.new(addresses[-1], 0)
```
drujensen: I think you want `addresses[0]`.  The first is usually the `client_ip`.  Subsequent IP addresses are the proxies.

spec/amber/router/pipe/client_ip_spec.cr
```diff
@@ -0,0 +1,47 @@
...
+          pipeline = Pipeline.new
+          pipeline.build :web do
+            plug ClientIp.new
+          end
+          pipeline.prepare_pipelines
+
+          ip_address1 = "102.168.35.88"
+          ip_address2 = "102.168.35.99"
+          headers = HTTP::Headers.new
+          headers["X-Forwarded-For"] = [ip_address1, ip_address2]
```
drujensen: shouldn't this be a comma delimited list if IP's?

spec/amber/router/pipe/client_ip_spec.cr
```diff
@@ -0,0 +1,47 @@
...
+          pipeline.prepare_pipelines
+
+          ip_address1 = "102.168.35.88"
+          ip_address2 = "102.168.35.99"
+          headers = HTTP::Headers.new
+          headers["X-Forwarded-For"] = ip_address1
+          headers["X-Client-IP"] = ip_address2
+          request = HTTP::Request.new("GET", "/client_ip_address", headers)
+          response = create_request_and_return_io(pipeline, request)
+          response.body.should contain ip_address2
```
drujensen: probably should look for `ip_address1` instead, right?

src/amber/pipes/client_ip.cr
docelic: You're right about this one. I've misunderstood the documentation.

spec/amber/router/pipe/client_ip_spec.cr
docelic: HTTP::Headers supports initialization via an Array so I've done it this way. It might in the end result in a comma-separated list being sent, but I've purposely left this decision to HTTP::Header's built-in mechanism.

spec/amber/router/pipe/client_ip_spec.cr
docelic: The ip_address2 is correct. If you look at the ClientIp plug initialization, you will see X-Forwarded vs. X-Forwarded-For. Therefore, the header X-Client-IP (and ip_address2) will be found and used.

This is reflected in the name of the test as well to an extent, it says "gets client IP from first custom header found".

drujensen: oops.  I put this comment on the wrong line.  I meant to put this on line 25 above. ^^^

spec/amber/router/pipe/client_ip_spec.cr
```diff
@@ -0,0 +1,47 @@
...
+          end
+          pipeline.prepare_pipelines
+
+          ip_address1 = "102.168.35.88"
+          ip_address2 = "102.168.35.99"
+          headers = HTTP::Headers.new
+          headers["X-Forwarded-For"] = [ip_address1, ip_address2]
+          request = HTTP::Request.new("GET", "/client_ip_address", headers)
+          response = create_request_and_return_io(pipeline, request)
+          response.body.should contain ip_address2
```
drujensen: should this one be ip_address2?

docelic: Oh right, after the change to addresses[0] it should be. Will correct later today.

eliasjpr:
Failing specs.
 
1) Amber::Pipe::ClientIp Gets client IP from headers gets last client IP from default header
     Failure/Error: response.body.should contain ip_address2
       Expected:   "102.168.35.88"
       to include: "102.168.35.99"

docelic:
Indeed, after the change from addresses[-1] to addresses[0] (as per @drujensen's comment) I've missed to update the test. Will do later today.

spec/amber/router/pipe/client_ip_spec.cr
```diff
@@ -0,0 +1,47 @@
...
+
+        it "gets client IP from first custom header found" do
+          pipeline = Pipeline.new
+          pipeline.build :web do
+            plug ClientIp.new(["X-Unmatched", "X-Client-IP"])
+          end
+          pipeline.prepare_pipelines
+
+          ip_address1 = "102.168.35.88"
+          ip_address2 = "102.168.35.99"
```
robacarp: I think it would help future devs understand the context of this test is the names `ip_address1` and `ip_address2` were more meaningful. Does `client_ip_address` and `proxy_ip_address` make sense?

src/amber/cli/templates/app/config/routes.cr
```diff
@@ -3,6 +3,7 @@ Amber::Server.configure do |app|
     # Plug is the method to use connect a pipe (middleware)
     # A plug accepts an instance of HTTP::Handler
     plug Amber::Pipe::PoweredByAmber.new
+    #plug Amber::Pipe::ClientIp.new(["X-Forwarded-For"])
```
robacarp: I like that this is disabled out of the box and also configured for the most common use case. Can we add a comment here or elsewhere that with some instructions about when this code should be activated?

Something like:

> Allows Amber to read the correct client information from headers set by http load balancers and proxies. Should remain disabled unless this application is going to be deployed behind a load balancer, otherwise it presents a security risk.

src/amber/cli/templates/app/config/routes.cr
```diff
@@ -15,6 +16,7 @@ Amber::Server.configure do |app|
   # All static content will run these transformations
   pipeline :static do
     plug Amber::Pipe::PoweredByAmber.new
+    #plug Amber::Pipe::ClientIp.new(["X-Forwarded-For"])
```
robacarp: This probably doesn't need to be here, I don't think the static pipeline would need this under normal circumstances.

src/amber/controller/base.cr
```diff
@@ -15,9 +15,27 @@ module Amber::Controller
...
+      :put?,
+      :request,
+      :requested_url,
+      :request_handler,
+      :response,
+      :route,
+      :session,
+      :valve,
+      :websocket?,
+    to: context
```
robacarp: 👍 thank you for cleaning this up.

src/amber/pipes/client_ip.cr
```diff
@@ -0,0 +1,34 @@
...
+    # Handler to determine real client IP address when Amber is behind a
+    # reverse proxy and save it to context.client_ip.
+    #
+    # The proxy should pass client IP in a header such as "X-Forwarded-For".
+    # In case of multiple values in a header, the first one is used.
+    #
+    # Crystal does not currently make remote IP available, so it is impossible
+    # to accept the header value only when coming from trusted proxy IPs.
+    # For that and other benefits, it is suggested to always run Amber apps
+    # behind proxies such as HAProxy or Nginx.
```
robacarp: One nitpick here. In the first sentence you label these reverse proxies and in the last sentence just proxies. Unless I'm missing something, those should both be the same.

src/amber/pipes/client_ip.cr
```diff
@@ -0,0 +1,34 @@
...
+          end
+        end
+        call_next(context)
+      end
+    end
+  end
+end
+
+class HTTP::Server::Context
+  property client_ip : Socket::IPAddress?
```
robacarp: This should go [in the extensions folder](https://github.com/amberframework/amber/tree/master/src/amber/extensions) instead of being hidden away in here.

src/amber/pipes/client_ip.cr
```diff
@@ -0,0 +1,34 @@
...
+      def initialize(header : String = "X-Forwarded-For")
+        @headers = [header]
+      end
+      def initialize( @headers : Array(String))
+      end
+
+      def call(context : HTTP::Server::Context)
+        @headers.each do |header|
+          if addresses = context.request.headers.get?(header)
+            context.client_ip = Socket::IPAddress.new(addresses[0], 0)
```
robacarp: The `x-forwarded-proto` is similarly important (and risky). Should it be included here as well?

src/amber/pipes/client_ip.cr
```diff
@@ -0,0 +1,34 @@
...
+    class ClientIp < Base
+      def initialize(header : String = "X-Forwarded-For")
+        @headers = [header]
+      end
+      def initialize( @headers : Array(String))
+      end
+
+      def call(context : HTTP::Server::Context)
+        @headers.each do |header|
+          if addresses = context.request.headers.get?(header)
```
robacarp: @amberframework/core-team @amberframework/contributors This is a little out of the scope of this pull request, but HTTP headers are technically [case ignorant](http://httpwg.org/specs/rfc7230.html#header.fields). Should we be doing something somewhere to downcase and internally standardize all http headers? It might help avoid [writing downcase everywhere](https://github.com/rack/rack/search?utf8=%E2%9C%93&q=downcase&type=) and I think coming out of the gate with a strong opinion on this might have some long term benefits.

robacarp:
I believe this will fix #534? I'm not sure if this works in a comment or just in the original pull request text: fixes #534

docelic:
Thanks for the input. Some of it I was already aware of (mentioned in my comment to #534) and will change and commit later today.

Re. #534, yes, this PR should fix it (it reworks/improves client_ip detection, which #534 was about).

spec/amber/router/pipe/client_ip_spec.cr
docelic: Sure, can do.

src/amber/cli/templates/app/config/routes.cr
docelic: No idea where a good position would be. I thought a description spanning more than one line wouldn't be appropriate to be placed above the pipe. You have any ideas where to put it, except in official docs?

src/amber/cli/templates/app/config/routes.cr
docelic: Yes and no. If you are running behind a load balancer and want the exact same behavior that you would have if you were receiving requests directly, you would enable this.

Also I would think it makes sense for the default behavior to be uniform, which would call for enabling (or leaving disabled) ClientIp pipe in both places. If individual people know specific reasons why they want to enable in one and not another, that should be their skilled and circumstantial decision.

src/amber/pipes/client_ip.cr
docelic: Right, commit with all of the changes coming later today.

src/amber/pipes/client_ip.cr
docelic: Could you clarify please? From what I see, Amber doesn't consult x-forwarded-proto header anywhere, so it seems that there is no risk?
Are you simply suggesting to add the value of x-forwarded-proto to Context as well?

src/amber/cli/templates/app/config/routes.cr
docelic: @robacarp I am somewhat leaning towards thinking that usage notes for each pipe (or any other piece of functionality) should be found in the documentation, rather than as comments in user visible/modifiable parts of configuration.

(It would certainly be more useful to have them in-place, agreed, but then various notes could be added to each pipe or feature, and it would be unreadable.)

Maybe once we document this in the documentation, just a link to the docs can be placed next to the pipe initialization.

Any opinions?

robacarp: Yeah, I agree. I saw your comment on the pipe after I wrote this, but I still think that it should have a note attached to it. Something terse which conveys that this thing is dangerous if you play with it outside of the right circumstances and points to the Guide (better?) or source code (at least) with more complete documentation on when and how it should be used.

@marksiemers Just because you said something in gitter about having written some of the guides, do you have any opinion on where this documentation should live?

src/amber/cli/templates/app/config/routes.cr
robacarp: I can't see myself ever using it, but I also don't see any harm to leaving it here. My thought is: if it's serving a static file, why is there any reason to be looking at the client ip address. But I suppose that isn't all encompassing.

src/amber/pipes/client_ip.cr
robacarp: Hm, that's interesting. I suppose my app is redirecting to https by nginx and not Amber. It's possible that Amber isn't doing anything with `x-forwarded-proto` at all. Since that seems to be the case, doing anything with it should be done in a separate PR. Carry on!

src/amber/cli/templates/app/config/routes.cr
marksiemers: If it is about a pipe in a pipeline, it should be an addition to this document: https://amberframework.org/guides/getting-started/routing/pipelines.md#pipelines

There is a balance here - as that document may get too large and have to be split up, but for now I think we can just add content to it.

The actual change will happen here: https://github.com/amberframework/online-docs/blob/bec7377b2e980df73f5dfdd253aa5a227a679700/getting-started/routing/pipelines.md

src/amber/cli/templates/app/config/routes.cr
docelic: @robacarp - true, the client IP is mostly irrelevant there. But it comes handy for logging, and also for the element of least surprise. Otherwise a developer could make 1 request in their browser (e.g. to reload the page) and see two different IPs being reported in Amber.

docelic:
Just a reminder here in case it went unnoticed.

Last commit here (984082d ) addresses all discussed issues, except one we can't do anything about.
(For those not interested in further discussion, please stop here. For those wanting to go into details, please read on.)
------------
So, here's the explanation of "one thing we can't do anything about":

Elias linked this nice article as source of good information regarding client_ip detection:
http://blog.gingerlime.com/2012/rails-ip-spoofing-vulnerabilities-and-protection

In it, the "best" method suggested for determining the real client IP involves knowing which IPs belong to trusted load balancers, and then taking the rightmost IP which is not on the trusted list as the client IP.

In other words, if we receive header X-Forwarded-For:"1.1.1.1,2.2.2.,3.3.3.3,4.4.4.4", and we know that 4 and 3 are our trusted load balancers, then 2 is the real client IP.

(That method is suggested as best because clients can pre-populate the beginning of X-Forward-For header with rogue data, and so if leftmost IP is taken, it can easily be an arbitrary/spoofed IP sent by the client.)

We can't implement that method exactly because Crystal does not give us access to remote_ip (the IP of the other side of the TCP connection). So we don't know whether the requests we are receiving are coming from trusted LB IPs or not.

Also, even if we could check, it would require that we 1) know and maintain lists of trusted LB IPs on our app server (which I believe is a wrong approach), and 2) it would involve looping over two arrays.

My alternative (which this PR implements, and which works regardless of missing remote_ip in Crystal) is to always take the leftmost IP, but with two simple additional rules (which in most cases will hold true with no additional work):

1) The application must either not have any LBs in front, or all requests must be going through at least one LB. (I.e. the app must choose one approach and should not be configured to receive production/public requests both through LBs and directly)

2) The first LB in a chain of LBs (or the only LB, if there is only one) needs to delete all existing values for X-Header-For, and then set the header value anew.

With these two principles in place and using the ClientIp pipe, taking the leftmost IP as client_ip is straightforward and correct.

src/amber/cli/templates/app/config/routes.cr
robacarp: > Otherwise a developer could make 1 request in their browser (e.g. to reload the page) and see two different IPs being reported in Amber, one for dynamic content and one for static content.

And that would mess with logging. Gotcha!

robacarp:
@docelic thanks for the summary. This is clearly a more contended feature than I would have expected, but it seems like everyone is coming at it with good intentions. This is the sort of code that, if not provided by Amber, ends up getting version controlled via StackOverflow comment threads and syndicated by copy-pasta. The world of web simply lives behind proxies and load balancers. 

I think this is a good tool, and like all good tools it has the potential to be used wrong.

drujensen:
@docelic agree with @robacarp on the detailed summary. 👍  

I think its a good practice for the LB/Proxy to strip headers that are not needed, so requiring the LB/Proxy to prevent an erroneous ip chain is an acceptable solution here.  Probably need to document this requirement though.

PR #568, opened 2018-01-22
Migration improvements (#554)

### LINKED ISSUE #554: Prevent Index out of bounds error when running `amber db migrate` with no migrations
### Description

When running `amber db migrate` and there are no migration files, you get a runtime, `Index out of bounds` error.

This could be fixed in micrate, but also, we can prevent it in the CLI.

This is particularly a problem with deploying to heroku, as the buildpack automatically runs migrations.

### Steps to Reproduce

Create a new amber project and run `amber db create migrate`

**Expected behavior:** [What you expect to happen]

No run-time error and maybe a warning about no migrations being present.

**Actual behavior:** [What actually happens]

Runtime error: `Index out of bounds`

100%

### Additional Information

Likely this is the best place to check if any migration files exist: https://github.com/amberframework/amber/blob/master/src/amber/cli/commands/database.cr#L42

Issue with micrate: https://github.com/juanedi/micrate/issues/26

Micrate is no longer actively maintained, but amberframework may take over.

KevinSjoberg:
### Description of the Change

Prevents crashing from happening when a user run `amber db migrate` without having any migrations present.

### Alternate Designs

None. Although, this should eventually be handled by [micrate](https://github.com/juanedi/micrate).

### Benefits

The user get textual feedback instead of an `IndexError`.

### Possible Drawbacks

None.

fixes #554

### DIFF
--- a/src/amber/cli/commands/database.cr
+++ b/src/amber/cli/commands/database.cr
@@ -52,17 +52,17 @@ module Amber::CLI
             `crystal db/seeds.cr`
             Micrate.logger.info "Seeded database"
           when "migrate"
-            Micrate::Cli.run_up
+            begin
+              Micrate::Cli.run_up
+            rescue e : IndexError
+              exit! "no migrations to run"
+            end
           when "rollback"
             Micrate::Cli.run_down
           when "redo"
             Micrate::Cli.run_redo
           when "status"
-            if Dir.exists?(MIGRATIONS_DIR)
-              Micrate::Cli.run_status
-            else
-              exit! "Directory #{MIGRATIONS_DIR} does not exist. Please run `amber db create migrate` in project root directory."
-            end
+            Micrate::Cli.run_status
           when "version"
             Micrate::Cli.run_dbversion
           else
--- a/src/amber/cli/commands/database.cr
+++ b/src/amber/cli/commands/database.cr
@@ -55,7 +55,7 @@ module Amber::CLI
             begin
               Micrate::Cli.run_up
             rescue e : IndexError
-              exit! "no migrations to run"
+              exit! "No migrations to run in #{MIGRATIONS_DIR}."
             end
           when "rollback"
             Micrate::Cli.run_down


src/amber/cli/commands/database.cr
```diff
@@ -52,17 +52,19 @@ module Amber::CLI
             `crystal db/seeds.cr`
             Micrate.logger.info "Seeded database"
           when "migrate"
-            Micrate::Cli.run_up
+            # TODO(kevinsjoberg): Remove begin/rescue as Micrate properly
```
robacarp: @KevinSjoberg I like your ownership but I'd rather not start dropping TODO comments around the code. If Micrate fixes this bug, the code can be removed later but it's not likely to break anything if that happens and it lives on.

marksiemers:
@amberframework/core-team - Did we come to a conclusion as to whether or not the migration folder should be created with `amber new` ?

It makes sense to me. If you want an app without a database, then perhaps creating one without `amber new` is the way to go.

Or should we have an option flag `amber new -d none` ?

src/amber/cli/commands/database.cr
```diff
@@ -52,17 +52,19 @@ module Amber::CLI
             `crystal db/seeds.cr`
             Micrate.logger.info "Seeded database"
           when "migrate"
```
eliasjpr: We should probably cover all off these commands `migrate,rollback,redo,status,versions` since all of them depend on the directory. No?

src/amber/cli/commands/database.cr
KevinSjoberg: @robacarp makes sense. I'll remove that.

src/amber/cli/commands/database.cr
KevinSjoberg: @eliasjpr my first commit, ensure new apps always has the `db/migrations` directory. Is that enough?

KevinSjoberg:
@marksiemers nothing has been discussed AFAIK. I just thought it made sense since it's shipping with a database adapter.

src/amber/cli/commands/database.cr
```diff
@@ -52,17 +52,19 @@ module Amber::CLI
...
-            Micrate::Cli.run_up
+            # TODO(kevinsjoberg): Remove begin/rescue as Micrate properly
+            # handles empty migration directories.
+            begin
+              Micrate::Cli.run_up
+            rescue e : IndexError
+              exit! "no migrations to run"
+            end
```
marksiemers: I would prefer an if statement over expecting an `IndexError` to be raised.
```crystal
if Dir.empty?(MIGRATIONS_DIR)
  exit! "no migrations to run"
else
  Micrate::Cli.run_up
end
```
Perhaps some error handling would also be good in addition to the conditional check.

marksiemers: A general Exception is rescued at the bottom of the method, so I think a conditional here will do.

KevinSjoberg: I'd agree if this was an expected error. It's not, it's a bug. If we're going to add proper error handling, we'd want to check if the directory exists and contains any migrations. Since the app template ships with a `.keep` in `db/migrations` directory we would want to exclude that during our conditional checks as well.

But if we think we should handle this properly in Amber, I'd suggest we add proper error handling for all migration related commands. 🙂

src/amber/cli/commands/database.cr
```diff
@@ -52,17 +52,17 @@ module Amber::CLI
             `crystal db/seeds.cr`
             Micrate.logger.info "Seeded database"
           when "migrate"
-            Micrate::Cli.run_up
+            begin
+              Micrate::Cli.run_up
+            rescue e : IndexError
+              exit! "no migrations to run"
```
eliasjpr: Just capitalize `"No migrations to run."`

Maybe we should state something more specific as the error `"No migrations to run in #{MIGRATIONS_DIR}."`

marksiemers:
The build tests are failing. @KevinSjoberg if you run `crystal spec -D run_build_tests` locally, it should reproduce.

It will take a while - 1 to 5 minutes

KevinSjoberg:
@marksiemers I'm not entirely sure this my fault. Looking at the Travis test ouput it looks like some sort of `animals` relation is missing.

marksiemers:
@KevinSjoberg - It looks like something must have gone wrong on that specific Travis build.

For a little background, the build tests actually use the CLI tools to build a brand new amber app, generate a bunch of things and run `crystal spec` on that generated app.

I was concerned because in the build tests, the `animals` relation is created with a migration, and it is the first table created, so I thought perhaps changing `amber db migrate` broke the ability to run migrations (one of the reasons we have the build tests).

All looks fine now though.

src/amber/cli/commands/database.cr
```diff
@@ -55,7 +55,7 @@ module Amber::CLI
             begin
               Micrate::Cli.run_up
             rescue e : IndexError
-              exit! "no migrations to run"
+              exit! "No migrations to run in #{MIGRATIONS_DIR}."
```
KevinSjoberg: The reason behind writing the error in lower case was to match the output given from Micrate. The error messages returned by Micreate are all lowercase.

Issue #476, opened 2018-01-04
Amber Watch - Configurable


drujensen:
### Description

This request is to change the `amber watch` command to be configurable.  This will remove the dependency on npm and allow for other services to be added if wanted.

The configuration will be stored in the `.amber.yml` file:
```yaml
type: app
database: sqlite
language: slang
model: granite
watch:
  server:
    files: ["src/**/*.cr", "src/**/*.ecr"]
    command: shards build
  client:
    files: "public/**/*"
    command: ["npm install", "npm run watch"]
  db:
    files: "db/**/*"
    command: amber db migrate
  jobs:
    files: "src/jobs/*"
    command: crystal run src/sidekiq.cr
```

faustinoaq:
Interesting, maybe we can use https://github.com/f/guardian to solve this issue, WDYT?

eliasjpr:
The fix for this feature is so simple that 

```
file = "./.amber.yml"
watchers =  YAML.parse_all(File.read(file))["watchers"]
if File.exists? file
  watchers.each do |program|
    @files << 
    create_process(program, path)
  end
else
  Amber::CLI.log "#{".amber.yml".colorize(:red)} does not exists!"
  exit 1
end
```

eliasjpr:
I agree with @elorest less dependencies to manage

faustinoaq:
I think we should add this feature to `config/enviroments/development.yml` because `amber watch` is only useful in development phase so would be good to have this kind of settings in this place, WDYT?

Something like:

```yml
secret_key_base: eV7pl_XuPgROPHGmuy0TSMC2yQ3OCInvRg4pxmj5nUA
port: 3000
name: forum

logging:
  severity: debug
  colorize: true
  filter:
    - password
    - confirm_password
  skip:
    -
  context:
    - request
    - session
    - headers
    - cookies
    - params

watch:
  server:
    files:
      - "src/**/*.cr"
      - "src/**/*.ecr"
    command: shards build
  client:
    files: "public/**/*"
    command:
      - "npm install"
      - "npm run watch"
  db:
    files: "db/**/*"
    command: amber db migrate
  jobs:
    files: "src/jobs/*"
    command: crystal run src/sidekiq.cr

host: 0.0.0.0
port_reuse: true
process_count: 1
# ssl_key_file:
# ssl_cert_file:
redis_url: "redis://localhost:6379"
database_url: sqlite3:./db/forum_development.db

session:
  key: amber.session
  store: signed_cookie
[...]

PR #574, opened 2018-01-23
Asset helper w/ modified webpack config

JoshJuncker:
### Description of the Change

Modified webpack config for development and production.  It now produces a manifest which is used by a new Asset helper that allows you to reference fingerprinted assets from controllers/views via asset_path("path_to_asset"). Code is also added to cause a compilation failure when trying to reference assets that don't exist.  When one doesn't exist, the compilation error will do it's best to suggest a different file if there is one similar.

### Alternate Designs

Webpack seems to be the standard and used already in Amber, so I decided to modify its config.  I'm not sure what other design there would be for the asset helper.

### Benefits

- Assets will be fingerprinted by default allowing browsers to cache them.
- You can now safely and easily reference assets in controllers and views via asset_path(path_to_asset) and dynamic_asset_path(path_to_file)

### Possible Drawbacks

- It adds an extra default dependency to the controller base which wouldn't be needed for api controllers.
- ~~It is currently a breaking change to older projects which don't have a config/webpack/manifest.json file.~~


src/amber/cli/templates/app/config/webpack/common.js
```diff
@@ -19,48 +22,61 @@ let config = {
...
         exclude: /node_modules/,
         use: [
-          'file-loader?name=/images/[name].[ext]'
+          {
+            loader: 'file-loader',
+            options: {
+              name: 'images/[name]-[hash].[ext]'
+            }
+          },
+          // {
```
drujensen: Can you remove the commented out section?

src/amber/version.cr
```diff
@@ -1,3 +1,3 @@
 module Amber
-  VERSION = "0.6.4"
+  VERSION = "0.6.5"
```
drujensen: this is updated by us during release

JoshJuncker: Good to know.  Thanks!

src/amber/cli/templates/app/package.json.ecr
```diff
@@ -12,16 +12,20 @@
     "test": "echo \"No test specified\" && exit 1"
   },
   "devDependencies": {
-    "babel-core": "^6.26.0",
-    "babel-loader": "^7.1.2",
-    "babel-preset-env": "^1.6.1",
+    "@babel/core": "^7.0.0-beta.38",
```
faustinoaq: Why we are using beta here?

> My hope was to keep the web pack and npm packages to a bare minimum. This adds two more npm packages and a plugin that is a beta version. Is there a way to do this without the extra packages?

@drujensen ^ https://github.com/amberframework/amber/pull/301#issuecomment-335983603

faustinoaq: BTW, this doen't add 2 packages but 9 package dependencies 😅

JoshJuncker: @faustinoaq I tried to closely match the npm dependencies that come default in rails/webpacker.  However, the amber.js webpack config was breaking with the latest version of babel-core/babel-loader/babel-preset-env (at least on my machine), and I couldn't find anything to fix it other than upgrading as shown in the babel-loader docs on github.  Feel free to modify the babel dependencies to a different configuration.  I will not be offended :).

JoshJuncker: Also, my experience with customizing webpack has been terrible.  Coming to Amber, I would have much preferred a nice setup that does for me everything that Sprockets does in rails. That's just how I feel, so I decided to configure it that way.  Also, these dependencies aren't really dependencies because they are only in the generated template, not in the actual compiled code.  Thus, anyone can customize their webpack however much they want.  It just means that new projects come with a nice default.  Those are my thoughts, but I'm also new to Amber and I know my thoughts are just opinion :).

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
```diff
@@ -66,7 +66,7 @@ h6, .h6 {
 
 #logo {
   background-color: #f4994b;
-  background-image: url(/dist/images/logo.png);
+  background-image: url(/dist/images/logo-c321a05deb5e6e097a40ecffbf90de45.png);
```
robacarp: There are a bunch of rendered files in here, compressed/mapped, etc. I would think we don't need to track those in git since the original sources will be there. The compiled assets will just get compiled on the first run of `amber w`, right?

src/amber/cli/templates/app/src/views/layouts/application.ecr.ecr
```diff
@@ -6,7 +6,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" />
-    <link rel="stylesheet" href="/dist/main.bundle.css">
+    <link rel="stylesheet" href="#{asset_path("stylesheets.css")}">
```
robacarp: This is going to be a nice feature.

src/amber/controller/helpers/asset.cr
```diff
@@ -0,0 +1,44 @@
+module Amber::Controller::Helpers
+  module Asset
+    ASSET_MANIFEST = {} of String => String
+
+    {{ run "../../run_macros/generate_asset_helpers" }}
+
+    macro asset_path(path)
+      {% if path.is_a?(StringLiteral) %}
+        {% if Amber::Controller::Helpers::Asset::ASSET_MANIFEST[path] %}
+          {{ "/dist" + Amber::Controller::Helpers::Asset::ASSET_MANIFEST[path] }}
```
robacarp: `/dist` is consistent with the current implementation and is a fine default, but should this be something that can be overridden?

src/amber/controller/helpers/asset.cr
```diff
@@ -0,0 +1,44 @@
...
+
+          ▸ If you're using a variable, switch to a literal string
+          ▸ If you can't use a literal string, use the 'dynamic_asset' method instead
+
+        ERROR %}
+      {% end %}
+    end
+
+    def dynamic_asset_path(path)
+      fingerprinted_path = Amber::Controller::Helpers::Asset::ASSET_MANIFEST[path]?
```
robacarp: That's a long constant! My first instinct would be to try and trim it down to just `MANIFEST`. Is the `Amber::Controller::Helpers::Asset::` explicit prefix necessary since it is in the module already? 

I think it might also help to make a getter for this, so if people want to inspect the manifest they don't have to go so deep.

src/amber/controller/helpers/asset.cr
```diff
@@ -0,0 +1,44 @@
...
+        {% raise <<-ERROR
+        \n
+        The 'asset' macro requires a literal string like "my-logo.png", instead got: #{path}
+
+        Try this...
+
+          ▸ If you're using a variable, switch to a literal string
+          ▸ If you can't use a literal string, use the 'dynamic_asset' method instead
+
+        ERROR %}
```
robacarp: I'm not experienced enough with Crystal macros to have come up with this myself, and I think this is a fantastic error message. 🌮 ❗️

src/amber/run_macros/generate_asset_helpers.cr
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  private def raise_missing_manifest_error
+    puts "Manifest at #{AssetManifestBuilder::MANIFEST_PATH} does not exist".colorize(:red)
+    puts "Make sure you have compiled your assets".colorize(:red)
+  end
+end
+
+begin
+  AssetManifestBuilder.build_with_retry
```
robacarp: Just to make sure I understand the flow here:

- Developer runs `crystal run src/app.cr` (or other build command)
- If assets aren't built, this script will wait up to `MAX_RETRIES`x`RETRY_AFTER`=5s for them
- If no assets are present still, this will abort the compile
- Otherwise, this provides a compiled list, provided by webpack, of static assets

Is that right?

src/amber/run_macros/missing_asset.cr
```diff
@@ -0,0 +1,20 @@
+require "colorize"
+require "json"
+require "levenshtein"
+
+missing_asset = ARGV.first
+manifest_path = File.expand_path("./config/webpack/manifest.json")
+manifest_file = File.read(manifest_path)
+manifest = JSON.parse(manifest_file)
+
+best_match = Levenshtein::Finder.find missing_asset, manifest.map(&.to_s), tolerance: 4
```
robacarp: Cute, I'll be curious to see how helpful this is in practice 😃

src/amber/cli/templates/app/package.json.ecr
drujensen: nice catch @faustinoaq.  Let's see if we can get this to work without the `beta`.

src/amber/run_macros/generate_asset_helpers.cr
```diff
@@ -0,0 +1,58 @@
+require "json"
+require "colorize"
+
+class AssetManifestBuilder
+  MANIFEST_PATH = File.expand_path("./config/webpack/manifest.json")
```
robacarp: One last comment. 

I don't know anything about webpack, is manifest.json a standardized format or is it specific to webpack? If so, that's okay, but should this class be `WebpackAssetManifestBuilder`?

Actually, now that I think about it, this class is more of a `Parser` than a `Builder`, isn't it?

robacarp:
In general, I like this. I have more questions than I do requested changes. Thanks for jumping in.

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
drujensen: @robacarp  We were doing this because we didn't have `npm` integrated with `amber w` but I think we still want this.  When you first startup, it takes a while for npm to compile and run.  wdyt?

src/amber/run_macros/generate_asset_helpers.cr
JoshJuncker: Essentially yes.  And I have to give credit for this stuff to Lucky Framework.  But they seemed to borrow heavily from Amber, so I think it's fair :).

src/amber/controller/helpers/asset.cr
JoshJuncker: I'm up for moving it into configuration.  I don't know where/how to do configuration settings though.  Point me in the right direction and I can go from there.

src/amber/cli/templates/app/package.json.ecr
```diff
@@ -12,16 +12,20 @@
...
+    "@babel/preset-env": "^7.0.0-beta.38",
```
drujensen: Do we need this?

src/amber/run_macros/generate_asset_helpers.cr
JoshJuncker: Yes.  I would say it is more of a parser.  A relic of grabbing it from Lucky ;)

src/amber/cli/templates/app/package.json.ecr
JoshJuncker: So here is the dilemma.  If you look at https://github.com/babel/babel-preset-env, you will notice that the repo is archived.  If you go to the new place and follow instructions, it installs the beta.  It also seems from the babel-loader readme that it wants you to use the beta as well.  Do we want to go with the beta as the older way seems to be on its way out?

src/amber/controller/helpers/asset.cr
docelic: You should add a new setting to src//amber/environment/settings.cr . Then it becomes available for use in .yml files and in Amber::Server.configure {} block.

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
robacarp: Since this pull auto retries, I don’t think it creates a race condition anymore.

src/amber/cli/templates/app/package.json.ecr
JoshJuncker: @faustinoaq I'm looking at this commit.  https://github.com/amberframework/amber/commit/76420ea0f6f2575a96c7722e8a14a476e625b728#diff-fbea2c0ca78b494fd0f7e012198847d4.  I can't get npm run build to succeed with this set of babel dependencies when including "import Amber from 'amber'" in the main.js.  The error is: ERROR in ../amber/assets/js/amber.js
Module build failed: Error: Couldn't find preset "env" relative to directory "/Users/joshjuncker/Workspace/amber/assets/js".  This doesn't happen with the newer beta versions.  And these older versions are being archived.  Look up babel-preset-env and you'll see that it is archived.  Can you help me figure out how to make it work?  Or if we just decide to use the beta.  Either way, I just can't figure it out :)

faustinoaq: Oh, you're right, official webpack docs are using beta versions https://webpack.js.org/loaders/babel-loader/#install Maybe we don't need to worry  about this 😅

drujensen: seriously??!!  The JS world is such a mess.  Let's make sure Crystal and Amber don't head down this path.

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
drujensen: ok, so for ~5 seconds the styles will not show but once the `npm` finishes, the auto reload will kick in and fix it.  Is that correct?

What if they don't have `npm` installed?

src/amber/cli/templates/app/package.json.ecr
faustinoaq: Sorry, we can use webpack 3.x babel-loader 7.x | babel 6.x Can we use this:

```
npm install babel-loader babel-core babel-preset-env
```

JoshJuncker: @faustinoaq I've followed that exactly and get the error referenced in my comment above.  Are you able to get it to work?

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
robacarp: @drujensen the way I read this code, it won't even compile until there is a manifest.json file available, so the amber server won't even be available. There will simply be a delay until npm renders the manifest. 

What about on a slower computer (think rasbpi), or a project with a huuuuuuuuge asset bundle, maybe npm doesn't finish building manifest.json until after the 5s timeout is hit. @JoshJuncker Any idea if thats a risk?

As for a dependency on npm, Amber ships by default using webpack to build assets. If you don't have npm running, I'm not sure the default project structure even makes sense. `src/assets` files don't ever get copied out to `/public` and so the static handler never serves them. Shipping precompiled static assets is a nice wedge to make the developer onboarding easier, especially before `amber w` was rewritten to spawn npm (what happens during `amber w` if npm insn't installed?)

eliasjpr:
CAN we reduce he wait time by spawning a process in the background or that’s not feasible?

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
faustinoaq: > what happens during amber w if npm insn't installed?

@robacarp Currently Amber just outputs `--: npm: command not found`, and works without problem: (without webpack)

ezgif-4-8e25e1f244

src/amber/cli/templates/app/package.json.ecr
faustinoaq: > Are you able to get it to work?

Yes, I do, I'm using non-beta version and I can't reproduce your issue:

```
➜  app npm install  
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

up to date in 7.772s
➜  app npm ls --depth=0
app@0.1.0 /home/main/Projects/app
├── babel-core@6.26.0
├── babel-loader@7.1.2
├── babel-preset-env@1.6.1
├── clean-webpack-plugin@0.1.17
├── compression-webpack-plugin@1.1.3
├── css-loader@0.28.9
├── extract-text-webpack-plugin@3.0.2
├── file-loader@1.1.6
├── node-sass@4.7.2
├── require-context@1.0.2
├── sass-loader@6.0.6
├── style-loader@0.19.1
├── webpack@3.10.0
├── webpack-manifest-plugin@1.3.2
└── webpack-merge@4.1.1
➜  app npm run watch

> app@0.1.0 watch /home/main/Projects/app
> webpack -w --config config/webpack/development.js

clean-webpack-plugin: /home/main/Projects/app/public/dist has been removed.

Webpack is watching the files…
```

Also works well if I use beta dependencies that you're using:

```
➜  app rm -rf node_modules 
➜  app npm install        

> node-sass@4.7.2 install /home/main/Projects/app/node_modules/node-sass
> node scripts/install.js

Cached binary found at /home/main/.npm/node-sass/4.7.2/linux-x64-59_binding.node

> uglifyjs-webpack-plugin@0.4.6 postinstall /home/main/Projects/app/node_modules/uglifyjs-webpack-plugin
> node lib/post_install.js


> node-sass@4.7.2 postinstall /home/main/Projects/app/node_modules/node-sass
> node scripts/build.js

Binary found at /home/main/Projects/app/node_modules/node-sass/vendor/linux-x64-59/binding.node
Testing binary
Binary is fine
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

added 681 packages in 39.883s
➜  app npm ls --depth=0   
app@0.1.0 /home/main/Projects/app
├── @babel/core@7.0.0-beta.38
├── @babel/preset-env@7.0.0-beta.38
├── babel-loader@8.0.0-beta.0
├── clean-webpack-plugin@0.1.17
├── compression-webpack-plugin@1.1.3
├── css-loader@0.28.9
├── extract-text-webpack-plugin@3.0.2
├── file-loader@1.1.6
├── node-sass@4.7.2
├── require-context@1.0.2
├── sass-loader@6.0.6
├── style-loader@0.19.1
├── webpack@3.10.0
├── webpack-manifest-plugin@1.3.2
└── webpack-merge@4.1.1
➜  app npm run watch      

> app@0.1.0 watch /home/main/Projects/app
> webpack -w --config config/webpack/development.js

clean-webpack-plugin: /home/main/Projects/app/public/dist has been removed.

Webpack is watching the files…
```

I'm using the same `common.js` as described in this PR

I don't know why your webpack build is failing, the presets `env` (non-beta) and `@babel/preset-env` (beta) works just fine for me 😅

JoshJuncker: @faustinoaq Are you including Amber in your main.js?  It isn't the common js reference that is failing.  It is because I include amber in my main js and amber.js has its own webpack config with it.  It works fine if I don't include amber.

src/amber/controller/helpers/asset.cr
faustinoaq: Yeah, this error message looks amazing:

```
expanding macro
in lib/amber/src/amber/controller/helpers/asset.cr:12: Error executing run (exit code: 1): ../../run_macros/missing_asset "stylesheet.css"

stdout:

    "stylesheet.css" does not exist in the manifest
    Did you mean "stylesheets.css"?
```

Awesome! ✨

src/amber/cli/templates/app/package.json.ecr
faustinoaq: > Are you including Amber in your main.js?

Yes, this is my `main.js` : (Other files like `run_macros/` dir and Assets Helper also included, I'm using this PR)

```js
var context = require.context('../images', true) // Necessary for asset helper.  Has webpack import all the images regardless of use in js/css.
var context = require.context('../fonts', true) // Necessary for asset helper.  Has webpack import all the fonts regardless of use in js/css.
import Amber from 'amber'

console.log("it works!");
```

But I'm having another problem, caching system is not working well, changes aren't reflected in the browser, it always shows the first compiled assets, it watches and compiles files but browser keep unchanged (fetching cached file):

vokoscreen-2018-01-25_09-35-11

Maybe we need to tweak `Amber::Pipe::Reload` too 😅

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
robacarp: @faustinoaq thanks for the nice demo.

I guess I'm a little divided. I don't really want Amber to depend on npm, but right now it does and I think acknowledging that would be more helpful to newcomers than trying to disguise it.

src/amber/controller/helpers/asset.cr
JoshJuncker: Do you have a suggestion for this?  I'm not exactly sure what you are intending? :)

robacarp: If line 36 and others can be written like this instead, I think it would clean up the code a lot.

```crystal
      fingerprinted_path = ASSET_MANIFEST[path]? 
```

Additionally, a method to get the manifest without using a constant would be nice for people wanting to do some reflection on it:

```
def manifest
  ASSET_MANIFEST
end
```

src/amber/controller/helpers/asset.cr
JoshJuncker: How do I use a config variable inside of a macro?  I can't get it to compile.

JoshJuncker: Nevermind.  I think I figured it out :).

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
faustinoaq: > I don't really want Amber to depend on npm

@robacarp That's why #476 is important. Making `amber watch` configurable will allow us to use npm alternatives.

JoshJuncker: If you want compiling/precompiling assets built in, is there a way around not depending on something to do that compiling?  Whether it be npm with webpack or if someone makes all the tools needed in crystal, it will still be a dependency.  The way it is structured right now, you can leave npm out.  I removed the necessity of having a manifest.json.  It will only require it if you use the asset helpers.  But since the default app template does use asset_path for including the base javascript and css, I have included the compiled assets and manifest so the project won't fail to run on the first amber watch.  If they want to remove it, they definitely  can.  I guess I would call it a soft dependency? :).  
The hard dependency is as @faustinoaq mentioned in amber watch.  As far as I know, npm and webpack are being run in amber watch no matter what.  I could be wrong though as I'm new... :).  
My opinion on the matter is to have a good default for compiling assets similar to sprockets.  For now, I think npm with webpack is a good option as it seems to be the best current standard.

damianham:
I am not really in favour of this change.  It adds complexity for a benefit that can be achieved with the Etag header and given that @drujensen suggests it won't even compile until there is a manifest.json file available, so the amber server won't even be available.... it doesn't sound like a good idea.  The benefit of having asset_path helpers do not seem to me to be of any substantial value and suffers from hide-it-itous (yeah.... that's a word that doesn't exist... I know).  As long as it is a feature I can turn on if I want to then OK, but as a default... not for me.  On a recent Rails contract we had issues with the asset pipeline - it is mature yet still can cause problems.  I like things simple, straight forward and I don't want a framework to hide anything from me.  Furthermore if I want to set an image URL in a stylesheet I want to use the image filename as it is, e.g. dist/images/logo.png and not be forced to use asset_path("dist/images/logo.png").

robacarp:
@damianham  I don't think this would prevent you from hardcoding a path in your stylesheet or elsewhere. You only need to do that if you want to use the manifest signature in the filename.

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
drujensen: @JoshJuncker I agree with you and I think we should leave in the compiled assets in the repo so the startup time is quick.  Also, if they don't have npm installed, it will still work, albeit they won't be able to make changes.  :-)

We will remove the dependency on npm in amber watch.  It should be configurable so they can replace it with yarn or any other favorite js toolset.

npm is a good starting point and should handle everything that sprockets provided.  This should make it just a little bit easier for a Rails project to be ported over.

src/amber/cli/templates/app/package.json.ecr
drujensen: I prefer we go with non-beta versions.  We will leave updating it as an exercise for the developer.

drujensen:
@damianham I agree that ETag headers is probably a better way to implement breaking cache.  I think we could provide a new header when assets are compiled, right?

@JoshJuncker What are your thoughts on using an ETag header instead of using a hash for every asset filename?

JoshJuncker:
@drujensen My preference is to use fingerprinting with far future expires header.  My reasoning for that over ETag is the client doesn't have to make an additional request for every asset.  With ETag, each asset triggers a request to the server, it is just the server doesn't respond with the entire file if it hasn't been modified according to the ETag.  With fingerprinting and using the expires header, the client doesn't need to make any requests to the server.
That is at least my understanding and opinion on the matter. I am by no means an http caching expert.  So correct me if I am wrong :).

paulcsmith:
Hey! I love that you're adding this feature. Just one request, could you mention that this was inspired by/is using code from Lucky?

* https://github.com/luckyframework/lucky/blob/master/src/run_macros/missing_asset.cr
* https://github.com/luckyframework/lucky/blob/master/src/run_macros/generate_asset_helpers.cr
* https://github.com/luckyframework/lucky/blob/master/src/lucky/asset_helpers.cr

JoshJuncker:
@paulcsmith Most definitely!  I’m actually a bit embarrassed I haven’t put it in there already.  I just added acknowledgements. And thank you for the inspiration :).

src/amber/controller/helpers/asset.cr
JoshJuncker: Thanks to Lucky on that one :)

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
elorest: > @robacarp That's why #476 is important. Making amber watch configurable will allow us to use npm alternatives.

Whatever we use I think that it should be default and supported. Not just a random example of "this is how you could do it.."

> @robacarp I don't really want Amber to depend on npm

It would be very nice if we could do all of this in crystal rather than relying on any sort of node dependancy. Obviously we could, I just don't think any of currently have time to replicate sass, scss, es6 compilation and auto reloading... although honestly the last one is probably the easiest with crystal.

elorest:
@drujensen @damianham ETag doesn't solve most of the problems that finger printing does. For example it requires quite a bit more back and forth and still relies on a expiration. So images, js or css could have changed minutes to days ago and the website will still load the old one because the caches haven't expired. 

With fingerprints since the file name is different it will immediately load the new file. Otherwise it can use the cached version forever.

Unfortunately it coupling of npm and crystal make this a little bit hard as there are multiple things that have to work together for this to work. 

We should probably talk about this more.

src/amber/cli/templates/app/public/dist/stylesheets-4bc3cc12d2780fec8bb311892094977b.css
marksiemers: Yeah, ES6 compilation - or trans-piling as I've heard it called. I don't think anyone wants to implement that.

damianham:
This can now be implemented as a recipe

faustinoaq:
Hi, I'm already working on some code to make webpack optional, see: #693  and #476 

So, I still think `Amber::Controller::Helpers::Asset` (based on Lucky) is very useful.

Maybe we can simplify and implement this in some agnostic way,

Just checking files inside `public` dir, WDYT?

BTW, this PR is too complex now, so we can't update it with master branch :sweat_smile: 

screenshot_20180512_184738

faustinoaq:
Hi @JoshJuncker since last time you opened this awesome PR amber has changes a lot.

Now we're trying to implement a new `amber watch` configuration to allow devs use other tools like yarn, parcel, rollup, etc, etc.

I still think `asset_path("path_to_asset")` is a nice tool and the Lucky helper message by @paulcsmith looks very nice, so I gonna open a new PR and implement a new version of this compatible with most recent changes.

Thanks you so much for collaborating with amber!

Tracking issue: https://github.com/amberframework/amber/issues/881

PR #578, opened 2018-01-24
Rename extensions

docelic:
- Rename Amber::Extensions::[Type]Extension into Amber::Extensions::[Type]
  (This makes it uniform with other things in Amber, e.g. pipes are named
   Amber::Pipe::[Name], not Amber::Pipe::[Name]Pipe.)
- Rename files accordingly as well
- Add test for HTTP extensions

### DIFF
--- a/spec/amber/extensions/http_spec.cr
+++ b/spec/amber/extensions/http_spec.cr
@@ -1,0 +1,13 @@
+require "../../../spec_helper"
+
+module Amber::Extensions
+  describe HTTPServerContext do
+    it "client_ip should exist and be nil" do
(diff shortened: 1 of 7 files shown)


spec/amber/extensions/http_spec.cr
```diff
@@ -0,0 +1,13 @@
+require "../../../spec_helper"
+
+module Amber::Extensions
+  describe HTTPServerContext do
+    it "client_ip should exist and be nil" do
```
veelenga: I don't think we need to test for the existence of a particular method. Crystal is compiled and guarantees the method exists for that class. Maybe just `"client_ip should be nil"` is enough?

PR #579, opened 2018-01-25
Unique user creation

gabrielengel:
Sign up was allowing multiple user entries with the same email. Small fix for avoiding that.


drujensen:
Should we put the unique constraint in the database instead?  That will avoid the possible race condition.

gabrielengel:
Will that raise an apropriate response to the user?

Em 25 de jan de 2018 8:35 PM, "Dru Jensen" <notifications@github.com>
escreveu:

eliasjpr:
We should do both the dB constraint and the validation fix.

faustinoaq:
I didn't realize my PR (https://github.com/amberframework/granite-orm/pull/118 ) makes granite validators different from amber validators 😅 

```crystal
# granite
validate :email, "You can't use outlook.com" do
  is_outlook? = email.not_nil!.ends_with? "outlook.com"
  !is_outlook?
end

# amber
validate :email, "already in use", -> (user : User) do # notice the proc parameters
  existing = User.find_by :email, user.email
  !existing
end
```

robacarp:
@faustinoaq I think ideally, both Amber and Granite validators would support both syntaxes

drujensen:
@gabrielengel @eliasjpr Any database error is added to the `errors` array so if you add it to both the db and the model, you will get two errors.  I am fine with either way.  This is fine as long as the DB is only accessed from the model.  For prototyping this is probably easier to understand and find.

PR #583, opened 2018-01-26
Moved wildcard route to bottom by default

elorest:
### Description of the Change

Moves while card static route to the bottom. I've helped multiple people over the last 6 months fix an issue where their route was getting caught by static pipe instead of the route they defined. 

This results in 405 errors. In Rails having a wildcard up top would have even more issues.


eliasjpr:
A couple of questions: 

- Is there a way we can add a spec to test this case? - Also what would stop someone from moving the wildcard back on top?

elorest:
Same thing as in rails. They know it will break things.

elorest:
Nothing stops them but we shouldn't generate code that could randomly fail.

elorest:
Its likely that we should just address this after we fix the router. All of my manual tests pass for this but for some reason the build specs fail.

robacarp:
This change was included in #432

PR #586, opened 2018-01-27
Add handler of */* requested response

### LINKED ISSUE #584: Incorrect handling of Accept header
### Description

Using following code in controller
```crystal
  def example_action
    ...
    respond_with do
      text("...")
      json("...")
    end
  end
```
Will act like expected when looking in browser, but when using `curl` or `wget` or another HTTP client, will always return HTTP code 406: Not Acceptable
The reason for this is incorrect handling of Accept HTTP Header: HTTP clients set this to "*/*", indicating that they can handle of types of data returned by server. Amber server can't understand this and returns error 406

Workaround for this was permanently setting this header to my browser's, but this is not valid fix.

**Expected behavior:** Return valid response
**Actual behavior:** Incorrect handling of HTTP Header "Accept" when set to "*/*"

### Versions

Using latest crystal and amber versions:
```
Crystal 0.24.1 (2017-12-22)

LLVM: 4.0.0
```
```
Amber CLI (amberframework.org) - v0.6.4
```

grig191:
### Description of the Change

Adding just a single line to response helpers:

```crystal
@requested_responses = @available_responses.keys if @requested_responses.includes?("*/*")
```

Makes all types of responses available if "*/*" is present in requested responses, so fixing issue #584: interacting with APIs built in amber by other clients like `wget` or `curl` or etc is now available

### Benefits

Fixes #584, so our loved amber is a bit less buggy now and we can use it when creating, for example, JSON APIs

### DIFF
--- a/spec/amber/controller/respond_with_spec.cr
+++ b/spec/amber/controller/respond_with_spec.cr
@@ -14,6 +14,14 @@ module Amber::Controller
         context.response.status_code.should eq 200
       end
 
+      it "respond_with html as default option with */* header" do
+        expected_result = "<html><body><h1>Elorest <3 Amber</h1></body></html>"
+        context.request.headers["Accept"] = "*/*"
+        ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "text/html"
+        context.response.status_code.should eq 200
+      end
+
       it "respond_with html" do
         expected_result = "<html><body><h1>Elorest <3 Amber</h1></body></html>"
         context.request.headers["Accept"] = "text/html"
@@ -25,6 +33,14 @@ module Amber::Controller
       it "responds with json" do
         expected_result = %({"type":"json","name":"Amberator"})
         context.request.headers["Accept"] = "application/json"
+        ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "application/json"
+        context.response.status_code.should eq 200
+      end
+
+      it "responds with json having */* at end" do
+        expected_result = %({"type":"json","name":"Amberator"})
+        context.request.headers["Accept"] = "application/json;*/*"
         ResponsesController.new(context).index.should eq expected_result
         context.response.headers["Content-Type"].should eq "application/json"
         context.response.status_code.should eq 200
--- a/spec/amber/controller/respond_with_spec.cr
+++ b/spec/amber/controller/respond_with_spec.cr
@@ -40,7 +40,7 @@ module Amber::Controller
 
       it "responds with json having */* at end" do
         expected_result = %({"type":"json","name":"Amberator"})
-        context.request.headers["Accept"] = "application/json;*/*"
+        context.request.headers["Accept"] = "application/json,*/*"
         ResponsesController.new(context).index.should eq expected_result
         context.response.headers["Content-Type"].should eq "application/json"
         context.response.status_code.should eq 200
@@ -116,6 +116,14 @@ module Amber::Controller
         context.request.path = "/response/1.texas"
         ResponsesController.new(context).index.should eq expected_result
         context.response.headers["Content-Type"].should eq "application/json"
+        context.response.status_code.should eq 200
+      end
+
+      it "responds html as default with invalid extension but having */* at end" do
+        expected_result = "<html><body><h1>Elorest <3 Amber</h1></body></html>"
+        context.request.headers["Accept"] = "unsupported/extension,*/*"
+        ResponsesController.new(context).index.should eq expected_result
+        context.response.headers["Content-Type"].should eq "text/html"
         context.response.status_code.should eq 200
       end
     end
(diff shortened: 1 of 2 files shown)


src/amber/controller/helpers/responders.cr
```diff
@@ -52,6 +52,7 @@ module Amber::Controller::Helpers
         raise "You must define at least one response_type." if @available_responses.empty?
         # NOTE: If only one response is requested don't return something else.
         @requested_responses << @available_responses.keys.first if @requested_responses.size != 1
+        @requested_responses = @available_responses.keys if @requested_responses.includes?("*/*")
```
elorest: Make this line and the one above work together and add tests.

grig191:
I don't think there is a reason to write new tests, because if we change 

https://github.com/amberframework/amber/blob/963248270f1b7589ed869222d107712d3656489f/spec/amber/controller/respond_with_spec.cr#L7 to:

```ruby
request.headers["Accept"] = "*/*"
```
Running test on current source will fail:
```ruby
Failures:

  1) Amber::Controller::Base #respond_with respond_with html as default option
     Failure/Error: ResponsesController.new(context).index.should eq expected_result

       Expected: "<html><body><h1>Elorest <3 Amber</h1></body></html>"
            got: "Response Not Acceptable."

     # spec/amber/controller/respond_with_spec.cr:12
```
But they work as expected on my branch, so current tests is testing both new and old functionality.

src/amber/controller/helpers/responders.cr
```diff
@@ -50,9 +50,12 @@ module Amber::Controller::Helpers
 
       private def select_type
         raise "You must define at least one response_type." if @available_responses.empty?
-        # NOTE: If only one response is requested don't return something else.
-        @requested_responses << @available_responses.keys.first if @requested_responses.size != 1
-        @requested_responses = @available_responses.keys if @requested_responses.includes?("*/*")
+        # NOTE: If */* is requested - make all available, but if only one response is requested don't return something else
+        if @requested_responses.includes?("*/*")
+          @requested_responses = @available_responses.keys
```
marksiemers: The issue here is that if a request header is something like: `application/json;*/*;` then this will likely return `text/html` - which does not seem ideal.

I think we need to come to an agreement on the requirements for this and maybe even psuedo-code for how things should work.

marksiemers: What about something like below? And if the requested response is not found in the available responses, shouldn't it be a 406?

```crystal
@requested_responses.find do |resp|
  @available_responses.keys.includes?(resp) || resp.includes?("*/*")
end
```

src/amber/controller/helpers/responders.cr
```diff
@@ -50,8 +50,12 @@ module Amber::Controller::Helpers
 
       private def select_type
         raise "You must define at least one response_type." if @available_responses.empty?
-        # NOTE: If only one response is requested don't return something else.
-        @requested_responses << @available_responses.keys.first if @requested_responses.size != 1
+        # NOTE: If */* is requested - make all available, but if only one response is requested don't return something else
+        if @requested_responses.includes?("*/*")
```
elorest: Both of these cases should just be `@available_responses.keys.first` since it will be matching against it's self.

marksiemers: Agreed, given the description above.

Just to confirm, this means if someone is rolling their own headers and they request `application/json;application/xml;text/plain;` and the server does not support any of those, the response will be `text/html`. And to increase performance on 100% of requests, we don't care about this unlikely case, correct?

elorest: I don't care about it since someone shouldn't be requesting json from an endpoint that only has html. In the past when the first 2 PR's were open on this it was discussed before.

marksiemers: I'm not disagreeing or suggesting a change, just making sure I have an understanding.

grig191:
[Changed](https://github.com/grig191/amber/blob/0a3a1baa8b779493b7135d663e027c4f36cb4867/src/amber/controller/helpers/responders.cr#L54) as mentioned by @elorest, now it is working as was supposed to
Also added few tests, here is diff:
https://github.com/grig191/amber/commit/0a3a1baa8b779493b7135d663e027c4f36cb4867#diff-3a53b42bc1e41a22d1031383203f29e5
`request.headers["Accept"] = ""` is now empty again

spec/amber/controller/respond_with_spec.cr
```diff
@@ -30,6 +38,14 @@ module Amber::Controller
         context.response.status_code.should eq 200
       end
 
+      it "responds with json having */* at end" do
+        expected_result = %({"type":"json","name":"Amberator"})
+        context.request.headers["Accept"] = "application/json;*/*"
```
marksiemers: This looks good. Can we get one more test where it asks for an unsupported format, then `*/*`, perhaps `"application/ndjson;*/*;`

grig191: Just a moment

grig191: And it is failing

marksiemers: hmm, that sounds like it is on the client side. @amberframework/core-team - any idea if this would be in amber or in crystal's HTTP client?

grig191: @marksiemers It was my fault, because separator between formats is `,`, not the `;`. I updated it with https://github.com/amberframework/amber/pull/586/commits/8d57895880035ada6ae8dccb4f81fc92f91e971a

marksiemers:
@elorest - Want to take another look?

Issue #585, opened 2018-01-27
[CLI] Allow piped content for 'amber exec'


docelic:
Currently, it seems it is not possible to read code from STDIN, so this doesn't work:

echo "script contents" | amber exec
amber exec < /from/file

Also, there seems to be a (rather unusual) behavior where the argument to exec is checked whether it exists as a file, and if not it is executed as an in-place code.

For example:

amber exec test  --> if file 'test' exists, executes it
amber exec test  --> if file 'test' does not exist, tries to execute code "test"

This is quite non-standard, and interpretation of filename as code should not be happening. It should be removed in favor of the mentioned support for reading from STDIN. (I.e. if exec is running interactively, editor is opened; if not running interactively, STDIN is read.)

sam0x17:
Totally agree on the STDIN, but I think the filename check is extremely slick. Wish more utilities worked this way.

sam0x17:
@docelic I'm assuming the reason this feature (the filename check) might be "undesirable" would be that it is technically possible for a file to exist in the current directory with the same name as some valid bit of crystal code, making it ambiguous/non-obvious whether the file would be executed or the line of crystal code would be executed. There are two mitigations I could see to reduce the chances of this happening, though I would argue the ease-of-use gains of not disabling this far outweigh the benefits:

1.  (100% mitigation / ambiguity is impossible): require filenames to always contain a `/` character, meaning for files in the current directory you have to do `./file`. This makes it impossible for there to be a case where valid crystal code is a also a filename that exists. Users of Node.js will be familiar with this requirement e.g. requiring files in the same directory etc.
2. Require `.cr` to be at the end of the downcased filename. While it is possible to have a class like `MyFile` with a class method `cr`, where `MyFile.cr` also exists in the current directory, this would be uncommon to the point where the example is essentially contrived.
3. Specify filenames OR code as a command-line arg. In this case I would advocate `amber exec [crystal code here]` having precedent, so we would have that syntax plus `amber exec -i filename.cr`. This is less exciting, but eliminates any ambiguity, for sure.

Regardless of what we do, the docs are pretty clear about what will happen even in the ambiguous cases (if the file exists, that will execute, otherwise attempt to execute as crystal code).

Thoughts?

robacarp:
Yeah, the command is trying to be clever and I think it actually has some potential to cause unintended consequences as a result. Looking for a `.cr` isn't really that different. If something needs to be done, a `-f` flag seems appropriate for reading from a file.

[...]

PR #587, opened 2018-01-27
Fix unterminated parenthesized expression

faustinoaq:
### Description of the Change

Currently this error happens:

```
➜  tareas amber x '                  
quote> Tareas.all.each do |t|
quote>   puts t
quote> end
quote> '
Error in line 1: while requiring "./tmp/1517093214408_console.cr"

Syntax error in tmp/1517093214408_console.cr:1: unterminated parenthesized expression

puts (
     ^


Exe
```

### Alternate Designs

I can use `{}` but I think `do..end` should be possible

```
➜  tareas amber x '
Tareas.all.each { |t| 
  puts t
}
'
05:47:58 Server     | (INFO) SELECT tareas.id, tareas.nombre, tareas.descripcion, tareas.lista FROM tareas : []
#<Tarea:0x555c32339f00>
nil
Exec
```

### Benefits

No code discrimination for `do..end`, also not confusing errors like `unterminated parenthesized expression`

### Possible Drawbacks

No

### DIFF
--- a/src/amber/cli/commands/exec.cr
+++ b/src/amber/cli/commands/exec.cr
@@ -30,15 +30,6 @@ module Amber::CLI
         system("cp #{_filename} #{@filename}") if _filename
       end
 
-      def code(args)
-        <<-CRYSTAL
-        result = begin
-          #{args.code}
-        end
-        puts result
-        CRYSTAL
-      end
-
       def run
         Dir.mkdir("tmp") unless Dir.exists?("tmp")
 
@@ -46,7 +37,7 @@ module Amber::CLI
           prepare_file
           system("#{options.editor} #{@filename}")
         else
-        File.write(@filename, code(args))
+        File.write(@filename, "result = begin\n#{args.code}\nend\nputs result.inspect\n")
         end
 
         result = ""
--- a/src/amber/cli/commands/exec.cr
+++ b/src/amber/cli/commands/exec.cr
@@ -37,7 +37,7 @@ module Amber::CLI
           prepare_file
           system("#{options.editor} #{@filename}")
         else
-        File.write(@filename, "result = begin\n#{args.code}\nend\nputs result.inspect\n")
+        File.write(@filename, "result = (\n#{args.code}\n)\nputs result.inspect\n")
         end
 
         result = ""
@@ -48,8 +48,12 @@ module Amber::CLI
           result = `crystal eval '#{eval_cmd.join(" ")}'`
         end
 
-        File.write(@filename.sub("console.cr", "console_result.log"), result) unless result.blank?
-        puts result, "Exec"
+        STDOUT.puts result
+        if result.blank?
+          STDOUT.puts "Info: if `unexpected token: )` then probably a `end` token is missing"
+        else
+          File.write(@filename.sub("console.cr", "console_result.log"), result)
+        end
       end
     end
   end
--- a/src/amber/cli/commands/exec.cr
+++ b/src/amber/cli/commands/exec.cr
@@ -37,7 +37,7 @@ module Amber::CLI
           prepare_file
           system("#{options.editor} #{@filename}")
         else
-          File.write(@filename, "puts (#{args.code}).inspect")
+        File.write(@filename, "result = begin\n#{args.code})\nend\nputs result.inspect\n")
         end
 
         result = ""
--- a/src/amber/cli/commands/exec.cr
+++ b/src/amber/cli/commands/exec.cr
@@ -37,7 +37,7 @@ module Amber::CLI
           prepare_file
           system("#{options.editor} #{@filename}")
         else
-        File.write(@filename, "result = (\n#{args.code}\n)\nputs result.inspect\n")
+          File.write(@filename, "result = (\n#{args.code}\n)\nputs result.inspect\n")
         end
 
         result = ""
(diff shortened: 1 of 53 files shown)


faustinoaq:
> I'm fine with this fix although sending multiple lines to amber exec as an arg seems like bad form.

I do this because looks similar to ICR 😅

robacarp:
Relevant discussion #585

src/amber/cli/commands/exec.cr
```diff
@@ -37,7 +37,7 @@ module Amber::CLI
           prepare_file
           system("#{options.editor} #{@filename}")
         else
-          File.write(@filename, "puts (#{args.code}).inspect")
+        File.write(@filename, "result = begin\n#{args.code}\nend\nputs result.inspect\n")
```
robacarp: This would probably be more readable if it was broken out into a heredoc:

```
code = <<-CRYSTAL
result = begin
  #{args.code}
end
puts result
CRYSTAL

File.write(@filename, code)
```

eliasjpr: I would even go a little further and make it its own method

```crytal
def code(args)
  <<-CRYSTAL
  result = begin
    #{args.code}
  end
  puts result
  CRYSTAL
end
```

marksiemers:
This has a legit test failure:
```shell
Failures:
  1) amber exec outside of project executes outside of project but without including project
     Failure/Error: File.read(logs.last?.to_s).should eq expected_result
       Expected: ":hello\n"
            got: "hello\n"
     # spec/amber/cli/commands/exec_spec.cr:50
Finished in 25.49 seconds
403 examples, 1 failures, 0 errors, 0 pending
```
I'm assuming this `#{args.code}` does an implicit `to_s` and perhaps `:hello.to_s # => "hello"`

src/amber/cli/commands/exec.cr
```diff
@@ -30,14 +30,23 @@ module Amber::CLI
         system("cp #{_filename} #{@filename}") if _filename
       end
 
+      def code(args)
+        <<-CRYSTAL
+        result = begin
+          #{args.code}
```
marksiemers: If we want to maintain functionality as before (for symbols), this may need string concatenation instead of interpolation. I'm not totally sure that will fix it though.

drujensen:
Looks like there is a breaking test:
```
  1) amber exec outside of project executes outside of project but without including project
     Failure/Error: File.read(logs.last?.to_s).should eq expected_result
       Expected: ":hello\n"
            got: "hello\n"
     # spec/amber/cli/commands/exec_spec.cr:50
```

src/amber/cli/commands/exec.cr
faustinoaq: The `code` method doesn't work

Explanation: https://github.com/amberframework/amber/pull/587#issuecomment-361470033

elorest:
Why does this end it's response with `Exec`. That's an issue. I started looking at it but don't want to have merge conflicts with this PR.

elorest:
It's frustrating that `amber exec` now returns a new line with `Exec` on it. This is because of puts being overwritten. For amber exec it needs to output with a normal puts instead of a logging puts.

faustinoaq:
> It's frustrating that `amber exec` now returns a new line with Exec on it. This is because of puts being overwritten. For amber exec it needs to output with a normal puts instead of a logging puts

I can try something, let me use `Helper.run`

> I'm fine with this but and like the simplicity of the solution although I don't really see the point of multiple line input from the CLI when you could use vim/nano etc.

I think is preference, I like the idea of using multi-line eval in the terminal, looks very similar to ICR:

Also some shells like `zsh` allow you to edit all lines without losing focus

```shell
➜  ~ crystal eval '
quote> class Foo
quote>   def initialize(@bar : String)
quote>   end
quote> end
quote> 
quote> foo = Foo.new("bar")
quote> puts typeof(foo)
quote> puts foo.@bar
quote> '
Foo
bar
➜  ~ # If I repeat the command I can edit any line:
➜  ~ crystal eval '
class Foo
  def initialize(@bar : String)
  end
end

foo = Foo.new("bar")
puts typeof(foo)
puts foo.@bar
'
```

elorest:
Ok thanks for looking into that exec thing. 

> I think is preference, I like the idea of using multi-line eval in the terminal, looks very similar to ICR:
> Also some shells like zsh allow you to edit all lines without losing focus

I understand the syntax and also use zsh. Although I just don't understand how it could possibly be easier to write/edit large code blocks in the shell vs an editor. :sweat_smile: 

`amber exec -b 1` will give you a copy of the same thing you just ran as well. Anyway lets merge this.

faustinoaq:
> Anyway lets merge this.

@elorest Wait one commit more, I'm working in something, testing a code regarding to this PR😅

faustinoaq:
> Wait one commit more, I'm working in something, testing a code regarding to this PR

@amberframework/core-team  This PR is ready for final review! 👍

faustinoaq:
> It's frustrating that amber exec now returns a new line with Exec on it. 

@elorest I'm using `STDOUT.puts`, so `Exec` newline isn't returned anymore 🎉

elorest:
Thanks a lot @faustinoaq

faustinoaq:
**Please: No merge yet**

I have another enhancement for this 👍

Some commits more 😅

faustinoaq:
Amber (v0.6.5) vs this PR

vokoscreen-2018-02-03_10-42-53

@amberframework/contributors ^ Oh yeah! Just in Time output 🎉 😄

elorest:
Pretty cool.

PR #588, opened 2018-01-27
Fix validate in model specs

faustinoaq:
### Description of the Change

This PR fixes model spec issues caused by https://github.com/amberframework/granite-orm/pull/117 and https://github.com/amberframework/granite-orm/pull/118

### Alternate Designs

No, but sometimes Travis randomly fails, we should take care of that 😅 

### Benefits

`validate` in Model specs will pass Travis after this fix

### Possible Drawbacks

No

### DIFF
--- a/src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
+++ b/src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
@@ -12,12 +12,12 @@ class <%= class_name %> < Granite::ORM::Base
   timestamps
 <% end -%>
 
-  validate :email, "is required", -> (<%= @name %> : <%= class_name %>) do
-    (email = <%= @name %>.email) ? !email.empty? : false
+  validate :email, "is required" do
+    !email.to_s.blank?
   end
 
-  validate :password, "is too short", -> (<%= @name %> : <%= class_name %>) do
-    <%= @name %>.password_changed? ? <%= @name %>.valid_password_size? : true
+  validate :password, "is too short" do
+    password_changed? ? valid_password_size? : true
   end
 
   def password=(password)


faustinoaq:
@elorest `@email ? @email.blank? : false` will fail:

```crystal
class Foo
  @baz : String?
  def bar
    @baz ? @baz.blank? : false
  end
end

p Foo.new.bar
```

Outputs:

```
Error in line 8: instantiating 'Foo#bar()'

in line 4: undefined method 'blank?' for Nil (compile-time type is (String | Nil))

Rerun with --error-trace to show a complete error trace.
```

So we can use `(baz = @baz) : baz.blank? : false` or `@baz.not_nil!.blank?` or `@baz.to_s.blank?`

https://carc.in/#/r/3gst

elorest:
Yeah that thing again... So why is it an instance var to begin with though?

faustinoaq:
> So why is it an instance var to begin with though?

Because this:

```crystal
s1 = WebSite.new(name: "")
s1.valid? # => false
s1.errors # => [Error(:name, "Name cannot be blank") ]

s2 = WebSite.new(name: "Foo Bar")
s2.valid? # => true
s2.errors # => empty
```

See: https://github.com/amberframework/granite-orm/pull/118

faustinoaq:
There we go:

Before granite was using something like this:

```crystal
class Bar
  macro inherited
    @@validations = {} of String => Proc(self, Bool)
  end
  
  macro validate(text, block)
    @@validations[{{text}}] = {{block}}
  end
  
  def errors
    errors = [] of String
    @@validations.each do |key, proc|
      errors.push key unless proc.call(self)
    end
    errors
  end
end

class Foo < Bar
  property name : String?
  validate "text1", -> (foo : Foo) do
    foo.name == "test1"
  end
end

f1 = Foo.new
f1.name = "test2"
puts f1.errors

f2 = Foo.new
f2.name = "test1"
puts f2.errors
```

Then, after https://github.com/amberframework/granite-orm/issues/110 I implemented https://github.com/amberframework/granite-orm/issues/117 (breaking change)

Now granite `validate` looks like this:

```crystal
class Bar
  macro inherited
    @validations = {} of String => Proc(Bool)
  end
  
  macro validate(text)
    def validate!
    @validations[{{text}}] = ->{{{yield}}}
    end
  end
  
  def errors
    validate!
    errors = [] of String
    @validations.each do |key, proc|
      errors.push key unless proc.call
    end
    errors
  end
end

class Foo < Bar
  property name : String?
  validate "text1" do
    @name == "test1"
  end
end

f1 = Foo.new
f1.name = "test2"
puts f1.errors

f2 = Foo.new
f2.name = "test1"
puts f2.errors
```

Maybe I did wrong, WDYT about returning old syntax? @drujensen 

```crystal
validate :name, "Name cannot be blank", -> (u : User) { !user.name.not_nil.blank? }
# vs
validate :name, "Name cannot be blank" { !name.not_nil!.blank? }
```

Also I think we can use default `""` (empty) value for string fields to avoid nil check

src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
```diff
@@ -12,17 +12,22 @@ class <%= class_name %> < Granite::ORM::Base
...
 
-  validate :email, "already in use", -> (user : User) do
-    existing = User.find_by :email, user.email
-    !existing
+  validate :password, "is too short" do
+    if password_changed?
+      valid_password_size?
+    else
+      true
+    end
```
marksiemers: WDYT about this style?
```crystal
return true unless password_changed?
valid_password_size?
```

src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
```diff
@@ -12,17 +12,22 @@ class <%= class_name %> < Granite::ORM::Base
...
   end
 
-  validate :password, "is too short", -> (<%= @name %> : <%= class_name %>) do
-    <%= @name %>.password_changed? ? <%= @name %>.valid_password_size? : true
+  validate :email, "already in use" do
+    existing = User.find_by :email, email
+    !existing
```
marksiemers: Same question:
```crystal
return true unless User.find_by(:email, email)
false
```

src/amber/cli/templates/auth/src/models/{{name}}.cr.ecr
```diff
@@ -12,17 +12,22 @@ class <%= class_name %> < Granite::ORM::Base
   timestamps
 <% end -%>
 
-  validate :email, "is required", -> (<%= @name %> : <%= class_name %>) do
-    (email = <%= @name %>.email) ? !email.empty? : false
+  validate :email, "is required" do
+    blank = email.to_s.blank?
+    !blank
```
marksiemers: WDYT about this style?
```crystal
return true unless email.to_s.blank?
false
```

spec/amber/cli/commands/generator_spec.cr
```diff
@@ -192,7 +192,6 @@ module Amber::CLI
           File.read("./src/handlers/authenticate.cr").should contain camel_case
           File.read("./src/handlers/authenticate.cr").should contain snake_case
           File.read("./src/models/admin_user.cr").should contain class_definition_prefix
-          File.read("./src/models/admin_user.cr").should contain snake_case
```
marksiemers: What changed to make this no longer necessary/valid?

faustinoaq: Granite models aren't using `snake_case` anymore, see https://github.com/amberframework/granite-orm/pull/117

This was a **breaking change**, next time we should release a major (0.X.0) version 😅

drujensen: oops.  shoot.  your right. I suck.  my bad.

drujensen:
@faustinoaq lets see if we can support both syntaxes for the validation in Granite.  That will fix this and avoid a breaking change.

faustinoaq:
> lets see if we can support both syntaxes for the validation in Granite. That will fix this and avoid a breaking change.

Oh, then let'me fix this PR 😉

faustinoaq:
**Note:** Latest change depend on https://github.com/amberframework/granite-orm/pull/130

drujensen:
This is now fixed in Granite.

PR #593, opened 2018-01-29
Make valid? return false on missing field

docelic:
When one calls params.valid?, before this patch the following happens:

1) If required param was provided, but does not pass validation, return value is false (this is correct)
2) If required param was not provided at all, valid? raises an exception (this is incorrect, should return false without raising)

With this patch, calling params.valid? returns false as expected.

Tests are also fixed to behave accordingly. (It is interesting that the description of an existing test was correct, but the expectation in it needed to be fixed.)

Also some excess "../" were fixed.

### DIFF
--- a/src/amber/validators/params.cr
+++ b/src/amber/validators/params.cr
@@ -13,8 +13,14 @@ module Amber::Validators
       @predicate = block
     end
 
-    def apply(params : Amber::Router::Params)
-      raise Exceptions::Validator::InvalidParam.new(@field) unless params.has_key? @field
+    def apply(params : Amber::Router::Params, raise_on_error = true)
+      unless params.has_key? @field
+        if raise_on_error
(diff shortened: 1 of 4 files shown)


src/amber/validators/params.cr
```diff
@@ -13,8 +13,14 @@ module Amber::Validators
       @predicate = block
     end
 
-    def apply(params : Amber::Router::Params)
-      raise Exceptions::Validator::InvalidParam.new(@field) unless params.has_key? @field
+    def apply(params : Amber::Router::Params, raise_on_error = true)
```
eliasjpr: Instead of adding a flag I would recommend to create a RequiredRule class, this way we avoid the `raise_on_error` config argument which I am not fond about. 

See https://gist.github.com/eliasjpr/4d89efe9f2b3cf37d1ad40137fc8b3e5

PR #594, opened 2018-01-30
Test Errno by value rather than description

docelic:
This patch tests errno by value rather than by textual description of the error.

### DIFF
--- a/src/amber/server/server.cr
+++ b/src/amber/server/server.cr
@@ -74,7 +74,7 @@ module Amber
           server.listen(settings.port_reuse)
           exit
         rescue e : Errno
-          if e.message == "accept: Too many open files"
+          if e.errno == Errno::E2BIG
             logger.error e.message
             logger.info "Restarting server..."
             sleep 1


src/amber/server/server.cr
```diff
@@ -74,7 +74,7 @@ module Amber
           server.listen(settings.port_reuse)
           exit
         rescue e : Errno
-          if e.message == "accept: Too many open files"
+          if e.errno == Errno::E2BIG
```
elorest: Is E2BIG equivalent to "accept: Too many open files"?

docelic:
@elorest please check if this looks reasonable since it is a change to your original code. Thanks.

elorest:
@docelic please review my comment above and advise before merging.

src/amber/server/server.cr
docelic: Yes, I looked it up in Crystal sources.

elorest:
Ok could probably be this same PR when you add that.

PR #595, opened 2018-01-30
Test Errno by code rather than description

docelic:
Errno::EMFILE corresponds to errno 24, which corresponds to the current 'accept: Too many open files'.

### DIFF
--- a/src/amber/server/server.cr
+++ b/src/amber/server/server.cr
@@ -74,7 +74,7 @@ module Amber
           server.listen(settings.port_reuse)
           exit
         rescue e : Errno
-          if e.message == "accept: Too many open files"
+          if e.errno == Errno::EMFILE
             logger.error e.message
             logger.info "Restarting server..."
             sleep 1


docelic:
@elorest @drujensen Here's the correct PR. Thanks!

src/amber/server/server.cr
```diff
@@ -74,7 +74,7 @@ module Amber
           server.listen(settings.port_reuse)
           exit
         rescue e : Errno
-          if e.message == "accept: Too many open files"
+          if e.errno == Errno::EMFILE
```
elorest: Did you see my question about whether `rescue e : Errno::EMFILE` would work so we wouldn't need to have an if statement?

docelic: @elorest Yes, I would think it wouldn't work because Errno::EMFILE is a constant, not a class?

elorest: Yeah looks like it returns an Int32. Have you tested this?

docelic: @elorest Yes, I've brought down the limit with 'ulimit -n' which got the condition to trigger and it started doing server restart in a loop.

PR #598, opened 2018-01-31
Update year

faustinoaq:
### Description of the Change

This PR updates year to 2018

### Alternate Designs

Not applicable

### Benefits

Updated LICENSE date

### Possible Drawbacks

Not applicable


docelic:
How about 2017-2018?
(The year should be the year of the first publication of the work, and then multiple years may be noted if the work goes through substantial revisions.)

PR #604, opened 2018-01-31
Upgrades newly generated projects to Quartz 0.5.0

robacarp:
### Description
Updates the `amber generate mailer` command to generate mailers which match the recent upgrades to Quartz

### Possible Drawbacks
🌯


robacarp:
@marksiemers good catch on the slang text template. 

Re: `.txt.ecr` vs `.text.ecr`, I find the latter more readable, but I don't have strong feelings on file extensions. I'm happy to rename if that's what we want to set as the precedent.

robacarp:
@elorest @drujensen @marksiemers updated for .txt. extensions instead of .text.

PR #608, opened 2018-02-01
Simplify app template

### LINKED ISSUE #607: Simplify default app template
### Description

Currently default app template has some lines I remove/modify all time in my projects without any issue

1. https://github.com/amberframework/amber/blob/master/src/amber/cli/templates/app/config/application.cr.ecr#L3

```crystal
require "./initializers/**"
require "amber"
require "../src/handlers/**"
require "../src/models/**"
require "../src/controllers/application_controller"
require "../src/controllers/**"
```

I use this:

```crystal
require "amber"

require "./initializers/*"
require "../src/models/*"
require "../src/handlers/*"
require "../src/controllers/*"
```

My questions

- Why we use `**` (I know is for subdirectories but we are using just one level, so `*` works just fine)
- Why we use that order, I tested it on different orders and seems that crystal is able to solve them without problems.

2. https://github.com/amberframework/amber/blob/master/src/amber/cli/templates/app/config/application.cr.ecr#L127

This line uncommented doesn't work:

```
settings.logger = ::Logger.new(STDOUT)
```

I use this:

```
settings.logger = Amber::Environment::Logger.new(STDOUT)
```

3. https://github.com/amberframework/amber/blob/master/src/amber/cli/templates/app/config/initializers/database.cr.ecr#L2

We are requiring model/orm dependencies here but They're already required by `src/models` so I just remove them without any issue.

faustinoaq:
### Description of the Change

Fixes #607 

### Alternate Designs

Keep app template without changes

### Benefits

Simple code without extra requires

### Possible Drawbacks

¿Misleading code?

### DIFF
--- a/src/amber/cli/templates/app/config/application.cr.ecr
+++ b/src/amber/cli/templates/app/config/application.cr.ecr
@@ -1,7 +1,7 @@
-require "./initializers/**"
 require "amber"
-require "../src/controllers/application_controller"
-require "../src/controllers/**"
+
+require "./initializers/*"
+require "../src/controllers/*"
 
 # About Application.cr File
 #
@@ -124,7 +124,7 @@ Amber::Server.configure do |settings|
   # Log: Is the logger that will be used for Amber and it defaults to ::Logger.new(STDOUT).
   # You can supply a custom logger.
   #
-  # settings.logger = ::Logger.new(STDOUT)
+  # settings.logger = Amber::Environment::Logger.new(STDOUT)
   #
   #
 end
--- a/src/amber/cli/templates/app/config/initializers/database.cr.ecr
+++ b/src/amber/cli/templates/app/config/initializers/database.cr.ecr
@@ -1,6 +1,4 @@
 <% if @model == "granite" -%>
-require "granite_orm/adapter/<%= @database %>"
-
 Granite::ORM.settings.database_url = Amber.settings.database_url
 Granite::ORM.settings.logger = Amber.settings.logger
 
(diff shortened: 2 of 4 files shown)


marksiemers:
Doesn't the `**` require files recursively and `*` only gets files directly in that folder?

If that is the case, I think leave the double stars.

faustinoaq:
@marksiemers Sorry for my misunderstanding, but Why we need require files recursively? 😅 

Are amber projects using subfolders inside `src/controllers` or `src/models` directories?

marksiemers:
They could:

```
models
-> concerns
    -> taggable.cr
-> article.cr
-> comment.cr
```
```
controllers
-> admin
  -> user_controller.cr
-> api
  -> article_controller.cr
-> static_controller.cr
```

faustinoaq:
@marksiemers Thank you! for your comment! I updated this PR and the related issue 👍

src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -124,7 +126,7 @@ Amber::Server.configure do |settings|
   # Log: Is the logger that will be used for Amber and it defaults to ::Logger.new(STDOUT).
   # You can supply a custom logger.
   #
-  # settings.logger = ::Logger.new(STDOUT)
+  # settings.logger = Amber::Environment::Logger.new(File.open("<%= @name %>.log", "w"))
```
robacarp: The comment above probably doesn't need to say `and it defaults to ::Logger.new(STDOUT)`. How about something that's less likely to fall out of date as things change:

> settings.logger is the logger that Amber and other capable shards in the project will use instead of writing directly to STDOUT. Supply a custom logger to write to syslog, etc.

faustinoaq: Done! 👍

PR #611, opened 2018-02-02
Remove 'require' for ORM shard

docelic:
The 'require' is no longer needed in model template because it is
already added to config/initializers/database.cr at the time of
app creation.

### DIFF
--- a/spec/support/fixtures/cli_fixtures.cr
+++ b/spec/support/fixtures/cli_fixtures.cr
@@ -55,8 +55,6 @@ module CLIFixtures
 
   def expected_post_model
     <<-MODEL
-    require "granite_orm/adapter/pg"
-
     class Post < Granite::ORM::Base
       adapter pg
       table_name posts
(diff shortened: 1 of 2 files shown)


drujensen:
there is a spec failing. can you take a peek?

faustinoaq:
I approve this 👍 

However spec need to be updated 👇 

```
Failures:
  1) amber generate model generates Post model
     Failure/Error: File.read(generated_model).should eq expected_post_model
       Expected: "require \"granite_orm/adapter/pg\"\n\nclass Post < Granite::ORM::Base\n  adapter pg\n  table_name posts\n\n  belongs_to :user\n\n  # id : Int64 primary key is created for you\n  field title : String\n  field body : String\n  field published : Bool\n  field likes : Int32\n  timestamps\nend\n"
            got: "class Post < Granite::ORM::Base\n  adapter pg\n  table_name posts\n\n  belongs_to :user\n\n  # id : Int64 primary key is created for you\n  field title : String\n  field body : String\n  field published : Bool\n  field likes : Int32\n  timestamps\nend\n"
     # spec/amber/cli/commands/generator_spec.cr:72
Finished in 28.43 seconds
413 examples, 1 failures, 0 errors, 0 pending
```

docelic:
I missed the failing spec because for some reason that spec did not run here. (As mentioned previously I believe that some specs silently aren't running (but do run if they are called with 'crystal spec spec/file.cr'). I didn't go debug why yet.)

Issue #15, opened 2017-04-22
[Suggestion] Use full class name in routes, instead of symbol


fridgerator:
The current route definition `get "/", :hello, :index, :api` is a little unreadable with 3 symbols in a row, and is not immediately obvious what each symbol does.

Using the full class name, in my opinion, makes the route more readable and obvious what the route is doing:
`get "/", HelloController, :index, :api`

PR #618, opened 2018-02-02
Quartz secure smtp upgrades

robacarp:
### Description of the Change

Small updates to the quartz initializer. See [#15 on quartz](https://github.com/amberframework/quartz-mailer/pull/15) for more information.

blocked by https://github.com/amberframework/quartz-mailer/pull/15

I also added an example config for connecting to sendgrid, changed the logger progname for the mailer logger.

**I updated the version number in the generated shard.yml, but that's a wild guess as to what the version number will be.**

### Benefits

- better example code
- allows authenticated smtp connections
- correct verbiage ~ssl~->**tls**

### Possible Drawbacks

- This is technically a breaking change.

### DIFF
--- a/src/amber/environment/settings.cr
+++ b/src/amber/environment/settings.cr
@@ -10,7 +10,7 @@ module Amber::Environment
       skip: Array(String?),
       context: Array(String?))
 
-    alias Value = String | Int32 | Bool | Nil
+    alias SettingValue = String | Int32 | Bool | Nil
 
     struct SMTPSettings
       property host = "127.0.0.1"
@@ -20,7 +20,7 @@ module Amber::Environment
       property password = ""
       property tls = false
 
-      def self.from_hash(settings = {} of String => Value) : self
+      def self.from_hash(settings = {} of String => SettingValue) : self
         i = new
         i.host     = settings["host"]?     ? settings["host"].as String     : i.host
         i.port     = settings["port"]?     ? settings["port"].as Int32      : i.port
@@ -80,9 +80,9 @@ module Amber::Environment
       ssl_key_file: {type: String?, default: nil},
       ssl_cert_file: {type: String?, default: nil},
       smtp: {
-        type:    Hash(String, Value),
+        type:    Hash(String, SettingValue),
         getter:  false,
-        default: Hash(String, Value){
+        default: Hash(String, SettingValue){
           "enabled" => false,
         },
       }
--- a/src/amber/environment/settings.cr
+++ b/src/amber/environment/settings.cr
@@ -10,8 +10,9 @@ module Amber::Environment
       skip: Array(String?),
       context: Array(String?))
 
+    alias Value = String | Int32 | Bool | Nil
 
-    struct SmtpSettings
+    struct SMTPSettings
       property host = "127.0.0.1"
       property port = 1025
       property enabled = false
@@ -19,7 +20,7 @@ module Amber::Environment
       property password = ""
       property tls = false
 
-      def self.from_hash(settings = {} of String => String | Int32 | Bool | Nil) : self
+      def self.from_hash(settings = {} of String => Value) : self
         i = new
         i.host     = settings["host"]?     ? settings["host"].as String     : i.host
         i.port     = settings["port"]?     ? settings["port"].as Int32      : i.port
@@ -50,10 +51,10 @@ module Amber::Environment
       @logger ||= LoggerBuilder.new(STDOUT, logging).logger
     end
 
-    @smtp_settings : SmtpSettings?
+    @smtp_settings : SMTPSettings?
 
-    def smtp : SmtpSettings
-      @smtp_settings ||= SmtpSettings.from_hash @smtp
+    def smtp : SMTPSettings
+      @smtp_settings ||= SMTPSettings.from_hash @smtp
     end
 
     YAML.mapping(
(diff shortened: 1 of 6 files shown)


robacarp:
@marksiemers I would like to push these settings into the env config, just haven’t gotten there yet. 

As far as tests, I don’t know that either. I haven’t spent much time looking at the tests for the generators. My guess based on the other failures I’ve seen in other pulls is, it’s probably a “does this command generate <string>” where <string> was just copied from the output of the command when I ran it manually :/

elorest:
I would like to see configuration for this added to `./config/environments/env.yml`

faustinoaq:
> blocked by amberframework/quartz-mailer#15

@robacarp amberframework/quartz-mailer#15 was already merged, Should we release quartz-mailer `v0.5.1`?

robacarp:
@faustinoaq yes, that needs to happen. But I also want to add the config file stuff before this goes in.

src/amber/environment/settings.cr
```diff
@@ -10,6 +10,27 @@ module Amber::Environment
...
+      property tls = false
+
+      def self.from_hash(settings = {} of String => String | Int32 | Bool | Nil) : self
+        i = new
+        i.host     = settings["host"]?     ? settings["host"].as String     : i.host
+        i.port     = settings["port"]?     ? settings["port"].as Int32      : i.port
+        i.enabled  = settings["enabled"]?  ? settings["enabled"].as Bool    : i.enabled
+        i.username = settings["username"]? ? settings["username"].as String : i.username
+        i.password = settings["password"]? ? settings["password"].as String : i.password
+        i.tls      = settings["tls"]?      ? settings["tls"].as Bool        : i.tls
```
robacarp: @eliasjpr @elorest @drujensen if you have some ideas on how to make this cleaner, I'm open to suggestions.

eliasjpr: @robacarp Why not follow the same pattern as the LoggingType. Create an alias for a SmtpType with a `NamedTupple(host: String, port: Int32, enabled: Bool, username: String, Password: String, tls: Bool)`

```crystal
alias SmtpType = NamedTupple(
  host: String, 
  port: Int32, 
  enabled: Bool, 
  username: String, 
  password: String, 
  tls: Bool
)

property smtp : SmtpType? 

# Yaml mapping
smtp: { type:  SmtpType, getter: false, default: nil }
```

```yaml
...
smtp:
  host: "smtp.postmarkapp.com", 
  port: 25, 
  enabled: true, 
  username: "59df0bfb-7995-493c-bc6d-68sdf7g87f6d", 
  password: "59df0bfb-7995-493c-bc6d-68sdf7g87f6d", 
  tls: true
...
Or it can simply not be defined
```

This way you can avoid to have the smtp Struct

robacarp: @eliasjpr I initially tried to do that but that meant that I had to add _all_ the smtp fields or none. This way allows several different shapes of configurations:

Full: 

```yaml
smtp:
  host: "smtp.postmarkapp.com"
  port: 25
  enabled: true
  username: "59df0bfb-7995-493c-bc6d-68sdf7g87f6d"
  password: "59df0bfb-7995-493c-bc6d-68sdf7g87f6d"
  tls: true
```

Mailcatcher:

```yaml
smtp:
  host: "localhost"
  port: 1025
  enabled: true
```

Etc. Initially, I built an `SmtpType` following the pattern of Logger but it didn't allow optional sub-fields.

But it's also possible there was something else going on crashing the yaml builder.

eliasjpr:
I am ready to merge this if everyone else is

robacarp:
@eliasjpr if you’re fine with the way the env params are defined, good to go.

src/amber/environment/settings.cr
```diff
@@ -10,6 +10,27 @@ module Amber::Environment
       skip: Array(String?),
       context: Array(String?))
 
+
+    struct SmtpSettings
```
marksiemers: Should we treat struct naming the same as class naming?
> Acronyms
> In class names, acronyms are all-uppercase. For example, HTTP, and LibXML.
> In method names, acronyms are all-lowercase. For example #from_json, #to_io.

Source: https://crystal-lang.org/docs/conventions/coding_style.html

If so, it should be `SMTPSettings`, right?

src/amber/environment/settings.cr
```diff
@@ -10,6 +10,27 @@ module Amber::Environment
...
+      property host = "127.0.0.1"
+      property port = 1025
+      property enabled = false
+      property username = ""
+      property password = ""
+      property tls = false
+
+      def self.from_hash(settings = {} of String => String | Int32 | Bool | Nil) : self
```
marksiemers: This is used enough, I think it should have an alias:
`alias SettingValue = String | Int32 | Bool | Nil`

robacarp: good idea. Since aliases are namespaced, `Value` is `Settings::Value` so that seems good enough, right?

src/amber/environment/settings.cr
robacarp: 👍

src/amber/environment/settings.cr
```diff
@@ -10,16 +10,17 @@ module Amber::Environment
       skip: Array(String?),
       context: Array(String?))
 
+    alias Value = String | Int32 | Bool | Nil
```
marksiemers: `Value` is the name of an abstract struct: https://crystal-lang.org/api/0.24.1/Value.html

Since this is namespaced I don't think it will technically matter, but I think calling it something more specific will help differentiate it.

`SettingValue` may not be the most elegant but I think it is semantically correct.

robacarp: It's borderline [Smurf code](https://blog.codinghorror.com/new-programming-jargon/) but I think I agree. Thanks @marksiemers

marksiemers:
Yay for Travis working the first time, and under 3 minutes. Thanks @eliasjpr

Issue #619, opened 2018-02-02
Amber should keep environment vars on forked processes


docelic:
As reported by @maxwells, it appears environment vars from the main process aren't passed onto the forks if multiple processes are used.

This is in turn causing variables like DATABASE_URL to not be passed through, and the application might not behave as intended.

PR #620, opened 2018-02-02
Passes ENV along to child processes

elorest:
### Description of the Change

Fixes https://github.com/amberframework/amber/issues/619

Passes ENV module along to child processes after turning it into a hash so that a couple other params could be merged in.

Please suggest effective way to write tests for this, since specs can't set ENV vars outside of process.

### DIFF
--- a/src/amber/server/server.cr
+++ b/src/amber/server/server.cr
@@ -46,7 +46,7 @@ module Amber
       env_hash = ENV.to_h
       if Cluster.master? && thread_count > 1
         while (thread_count > 0)
-          Cluster.fork ({"id" => thread_count.to_s}.merge(env_hash))
+          Cluster.fork (env_hash.merge({"id" => thread_count.to_s}))
           thread_count -= 1
         end
         sleep
--- a/src/amber/server/server.cr
+++ b/src/amber/server/server.cr
@@ -44,10 +44,7 @@ module Amber
     def run
       thread_count = settings.process_count
       if Cluster.master? && thread_count > 1
-        while (thread_count > 0)
-          Cluster.fork
-          thread_count -= 1
-        end
+        thread_count.times{ Cluster.fork }
         sleep
       else
         start
--- a/src/amber/server/server.cr
+++ b/src/amber/server/server.cr
@@ -43,9 +43,10 @@ module Amber
 
     def run
       thread_count = settings.process_count
+      env_hash = ENV.to_h
       if Cluster.master? && thread_count > 1
         while (thread_count > 0)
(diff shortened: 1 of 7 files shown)


src/amber/server/server.cr
```diff
@@ -43,9 +43,10 @@ module Amber
 
     def run
       thread_count = settings.process_count
+      env_hash = ENV.to_h
       if Cluster.master? && thread_count > 1
         while (thread_count > 0)
-          Cluster.fork ({"id" => thread_count.to_s})
+          Cluster.fork (env_hash.merge({"id" => thread_count.to_s}))
```
eliasjpr: Can we do this in the cluster class instead?

eliasjpr: And I think crystal classes has a dup/clone method see https://github.com/amberframework/amber/blob/3374da08814340e50c4deb47daeaa8cb785311d8/src/amber/server/cluster.cr#L5

eliasjpr:
Also I don’t recommend to use merge since is appending to the env object an is key

elorest:
I don't understand what you mean by:

> Also I don’t recommend to use merge since is appending to the env object an is key

src/amber/server/server.cr
```diff
@@ -45,7 +45,7 @@ module Amber
       thread_count = settings.process_count
       if Cluster.master? && thread_count > 1
         while (thread_count > 0)
```
drujensen: WDYT?
```
(0...thread_count).each do |cnt|
  Cluster.fork
end
```

src/amber/server/server.cr
```diff
@@ -44,10 +44,7 @@ module Amber
     def run
       thread_count = settings.process_count
       if Cluster.master? && thread_count > 1
-        while (thread_count > 0)
-          Cluster.fork ({"id" => thread_count.to_s})
-          thread_count -= 1
-        end
+        thread_count.times{ Cluster.fork }
```
faustinoaq: Small formatting issue here: _missing space_ `thread_count.times { Cluster.fork }`

elorest: I ran `crystal tool format` on it before committing and this is how it made it...

 It looks very strange to me to put a space between the method and `()` and `{}`? When did we start doing that?

faustinoaq: I think is just a convention, by default `crystal tool format` _adds_ an space in between `method {` (braces) and _ignore_ spaces between `method (` (parenthesis)

faustinoaq: But anyway, this PR looks good, let's merge it 👍

Issue #622, opened 2018-02-03
Test cleanup shouldn't delete test folder


elorest:
### Description

Currently test cleanup deletes test folder. This is bad since tmp isn't owned by or only used by tests.

PR #623, opened 2018-02-03
Fixed issue where running tests deleted tmp folder.

elorest:
### Description of the Change

Fixes https://github.com/amberframework/amber/issues/622

Also used file utils which should make it easier if we ever have to support Amber in windows or something.

### DIFF
--- a/spec/support/helpers/cli_helper.cr
+++ b/spec/support/helpers/cli_helper.cr
@@ -1,3 +1,5 @@
+require "file_utils"
+
 module CLIHelper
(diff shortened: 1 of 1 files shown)


spec/support/helpers/cli_helper.cr
```diff
@@ -6,7 +8,9 @@ module CLIHelper
 
   def cleanup
     Dir.cd CURRENT_DIR
-    `rm -rf ./tmp/`
+    if Dir.exists?(TESTING_APP)
```
drujensen: good catch!

Issue #624, opened 2018-02-03
[Specs] Tests fail on master.


elorest:
### Description

Crystal spec on master fails.

### Steps to Reproduce

1. run `crystal spec`

**Expected behavior:** Tests pass.

**Actual behavior:** Tests fail.

```sh
Failures:

  1) Amber::Pipe::Error returns status code 500 for all other exceptions

       Index out of bounds
       Array(Tuple(String, Int32, Int32, Int32))@Indexable(T)#at<UInt32>:Tuple(String, Int32, Int32, Int32)
       Array(Tuple(String, Int32, Int32, Int32))@Indexable(T)#[]<UInt32>:Tuple(String, Int32, Int32, Int32)
       Debug::DWARF::LineNumbers#register_to_matrix<Debug::DWARF::LineNumbers::Sequence, Debug::DWARF::LineNumbers::Register>:Nil
       Debug::DWARF::LineNumbers#read_statement_program<Debug::DWARF::LineNumbers::Sequence>:Nil
       Debug::DWARF::LineNumbers#decode_sequences<UInt64>:Nil
       Debug::DWARF::LineNumbers#initialize<IO::FileDescriptor+, UInt64>:Nil
       Debug::DWARF::LineNumbers::new<IO::FileDescriptor+, UInt64>:Debug::DWARF::LineNumbers
       Exception+@Exception#backtrace?:(Array(String) | Nil)
       Exception+@Exception#inspect_with_backtrace<String::Builder>:Nil
       Exception+@Exception#inspect_with_backtrace:String
       Amber::Controller::Error#internal_server_error:String
       Amber::Pipe::Error#call<HTTP::Server::Context>:(Bool | HTTP::Headers | HTTP::Server::Context | IO+ | Int32 | Nil)
       create_request_and_return_io<Amber::Pipe::Error, HTTP::Request>:HTTP::Client::Response
       ~procProc(Nil)@spec/amber/router/pipe/error_spec.cr:15
       it<String, String, Int32, Int32, &Proc(Nil)>:(Array(Spec::Result) | Nil)
       ~procProc(Nil)@spec/amber/router/pipe/error_spec.cr:5
       Spec::RootContext::describe<String, String, Int32, &Proc(Nil)>:Spec::Context+
       describe<Amber::Pipe::Error:Class, String, Int32, &Proc(Nil)>:Spec::Context+
       __crystal_main
       _crystal_main<Int32, Pointer(Pointer(UInt8))>:Nil
[...]

PR #626, opened 2018-02-03
[BUGFIX] Refixed error controller where ex#inspect_with_backtrace raises new exception.

elorest:
### Description of the Change

Fix for https://github.com/amberframework/amber/issues/624

`amber spec` is broken because `inspect_with_backtrace` currently fails if backtrace is empty.
It's been this way ever since 0.24.1. This was one of the things @drujensen and I figured out over new years when we were trying to get the release with new crystal support finished before midnight. 540cde6

Looks like it was changed back on the 25th with a PR that cleaned up the output: 3fcf593e

I've fixed it so that if there's an exception on printing the backtrace it will just print the message.

### DIFF
--- a/src/amber/controller/error.cr
+++ b/src/amber/controller/error.cr
@@ -13,9 +13,9 @@ module Amber::Controller
 
     def internal_server_error
       response_format("ERROR: #{@ex.inspect_with_backtrace}")
-      # NOTE: As I commented in December 540cde6
-      # #inspect_with_backtrace will fail in some situtions which breaks the tests.
-      # it was added back with 3fcf593e
+      # IMPORTANT: #inspect_with_backtrace will fail in some situtions which breaks the tests.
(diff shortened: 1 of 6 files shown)


src/amber/controller/error.cr
```diff
@@ -13,6 +13,12 @@ module Amber::Controller
 
     def internal_server_error
       response_format("ERROR: #{@ex.inspect_with_backtrace}")
+      # IMPORTANT: #inspect_with_backtrace will fail in some situtions which breaks the tests.
+      # Even if you call @ex.callstack you'll notice that backtrace is nil.
+      # #backtrace? is supposed to be safe but it exceptions anyway.
+      # Please don't remove this without verifying that crystal core has been fixed first.
```
robacarp: Is there an issue on crystal-Lang about this that can be left here in the comment?

marksiemers: @robacarp - It is fixed in crystal master.

PR #627, opened 2018-02-03
[TEST] pushing this to see if it passes in travis everytime.

elorest:
### Description of the Change

Just a travis conf test to see how tests pass if only run once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,13 @@ services:
 before_install:
   - docker build -t garnet-spec .
 
-env:
-  - TEST_SUITE=./spec/build_spec.cr
-  - TEST_SUITE=./spec/amber
+# env:
+#   - TEST_SUITE=./spec/build_spec.cr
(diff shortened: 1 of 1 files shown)


faustinoaq:
> Build tests should be run completely in isolation because it is a full regression test to make sure everything in a generated app and amber cli works.

Yeah, Maybe that's the reason why `restart` button in Travis CI works after trying twice, when one of the jobs has finished.

elorest:
@eliasjpr But both of the tests have to do a lot of setup which is redundant. Also it since `crystal spec` is expected to work it would probably be best to run the same set of tests in the CI as people should be running at home. We obviously already have the abilty to run individual tests if we need too.

faustinoaq:
@eliasjpr I agree with @elorest the test made in this PR are taking just 5 min 4 sec

screenshot_20180207_180851

I think a 5 min test is very acceptable

PR #628, opened 2018-02-03
[Formatting] Applies crystal formatting

eliasjpr:


src/amber/cli/templates/app/config/routes.cr
```diff
@@ -3,7 +3,7 @@ Amber::Server.configure do |app|
     # Plug is the method to use connect a pipe (middleware)
     # A plug accepts an instance of HTTP::Handler
     plug Amber::Pipe::PoweredByAmber.new
-    #plug Amber::Pipe::ClientIp.new(["X-Forwarded-For"])
+    # plug Amber::Pipe::ClientIp.new(["X-Forwarded-For"])
```
docelic: I generally think that this is bad decision on the part of Crystal's formatter.

Specifically:
1) Lines which have a space (like "# something") should be textual comments which don't need to / cannot be uncommented
2) And lines with no space (like "#something") should be lines of code that actually can be uncommented

faustinoaq:
> Can we add a step to CI which fails if crystal fmt changes anything?

Perhaps ameba can do that, @veelenga WDYT?

PR #634, opened 2018-02-04
Fix warning on make install

docelic:
Fixes a warning on 'make install' which is printed if the previous binary
does not exist.

Functionally the previous and current behavior are almost identical, with
the benefit that the irrelevant error message is now silently ignored
rather than too verbosely printed and potentially confusing to the user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ run:
 	$(AMBER)
 
 install: build | $(INSTALL_DIR)
-	@-rm $(AMBER_SYSTEM)
+	@rm -f $(AMBER_SYSTEM)
 	@cp $(AMBER) $(AMBER_SYSTEM)
 
 link: build | $(INSTALL_DIR)


docelic:
The -f will return 0 instead of 1 if file doesn't exist, so using - to continue execution in case of failure isn't required.

robacarp:
My preference is to leave the leading hyphen. The point of using the force flag is to silence an error and keep going, but that could be overlooked by a future maintainer. Leaving a leading hyphen makes that explicit. 

But I won’t block a merge for that preference.

docelic:
@robacarp the effect of a hyphen was exactly that - to continue in case of an error. But this was wrong for two reasons:

1) It printed a nasty error (with no indication that it is just a warning), so to the end user it was appearing as if the whole make target had errored out, which it didn't
2) It was continuing execution even in the case of serious errors like file existing, but user unable/not having permission to delete it

Now, with "-" removed and with -f added, the behavior is as follows:

1) If the error is just file missing (so there is nothing to delete), no warning is printed at all
2) If the error is more serious (like file present, but user not able to delete it), then this warning is printed and execution stops

I believe this is now better/more correct behavior than previously. 
If you want that we continue execution even if 'rm' fails (e.g. somehow expecting that rm failed but that we'll still be able to overwrite the destination when we do a cp), then let me know and I can sure add the "-" back. Thanks!

Issue #592, opened 2018-01-29
[Router] Allow multiple values for a single parameter key


maiha:
### Description

`Amber::Router::ParamsParser` can't handle multiple values for a parameter.

- https://github.com/amberframework/amber/blob/master/src/amber/router/params_parser.cr#L146-L147
```crystal
      HTTP::Params.parse(values) do |key, value|
        params[key] = value
```

This always overwrites previous values where `params` is a `Amber::Router::Params < Hash(String, String)` .

### Steps to Reproduce

```crystal
require "amber"

request  = HTTP::Request.new("GET", "/?foo=bar&foo=baz")
response = HTTP::Server::Response.new(IO::Memory.new)
context  = HTTP::Server::Context.new(request, response)
context.params["foo"] # => "baz"
```

**Expected behavior:**

We should provide another method to **fetch all** values like `Rails`, `Crystal`, `Kemal` and any other libraries.

**Crystal**
```crystal
require "http/params"

params = HTTP::Params.parse("foo=bar&foo=baz")
params.fetch("foo")     # => "bar"
params.fetch_all("foo") # => ["bar", "baz"]
```

**Kemal**
```crystal
require "kemal"

request  = HTTP::Request.new("GET", "/?foo=bar&foo=baz")
params = Kemal::ParamParser.new(request)
params.query["foo"]           # => "bar"
params.query.fetch_all("foo") # => ["bar", "baz"]
```

### Versions

- `Amber CLI (amberframework.org) - v0.6.4`
- `Crystal 0.24.1 (2017-12-22)`

**Amber** is a quite cool design, and I love it. 😄 
I thinks it is time to polish it for production use! If no one has started this problem, I will try it.

Best regards,

elorest:
Interesting... Feel free to take a stab at it.

eliasjpr:
I believe this can be as simple as just appending to the param key @elorest 

@maiha thank you for the thorough submission.

robacarp:
I don't recall anything like this from Rails, and I don't see anything in their codebase that shows anything like `params.fetch_all`. My experience with it says later parameters simply overwrite so previous parameters aren't available. Unless you're thinking of [hash and array parameters](http://guides.rubyonrails.org/action_controller_overview.html#hash-and-array-parameters)?

Apparently [there is no standard](https://www.owasp.org/images/b/ba/AppsecEU09_CarettoniDiPaola_v0.8.pdf) (9th slide) for what to do when passing multiple parameters with the same name.

[...]

PR #637, opened 2018-02-07
Refactors Params into Module, Addresses Issue  #592

eliasjpr:
Refactors Params into Modules

Issue  #592

> As a developer I want to parse params and return an array of values
for a given param key.

The changes here introduces a Parser module that handles parsing query
params. It organizes each parser into its own class.

List of changes:
- Adds Params Specs
- Adss params class
- Moves Amber::Router::Files::File to Amber::Router::File
- Allows to parse form data from request body
- Allows for parsing form Multipart form data
- Allows for JSON request body parsing
- Adds a Request.cr file 
- Removes the original ParamsParser in favor of the new Parser
- Cleans context spec 
- Params are parsed only once per request (Adds parse state to memoize)
- Removes redundant Request Handler Method now is request.route
- Repurpose invalid_route? to valid_route? (the logic in the method is easier to read and understand)
- [Controller] Error fixes error controller (This was also addressed by @marksiemers in another PR) I made the change here because it was hindering development.

These changes allows for lazy loading params in instances where  there is no need to override method, and matching Radix route, with this change we can work on allowing Pipelines to run before  matching a route.

Thanks @elorest for the pairing session for shedding some light on the module and Params struct


src/amber/controller/error.cr
```diff
@@ -43,5 +37,13 @@ module Amber::Controller
         message
       end
     end
+
+    private def exception_error_message
```
eliasjpr: If we merge @marksiemers PR the changes here can be discarded

src/amber/router/request.cr
```diff
@@ -0,0 +1,53 @@
...
+    matched_route.payload? || router.socket_route_defined?(self)
+  end
+
+  def process_websocket
+    router.get_socket_handler(self)
+  end
+
+  def matched_route
+    radix_route = router.match_by_request(self)
+    radix_route.params.each { |k,v| @params.store.add(k, v) }
```
eliasjpr: This merges the radix route params with the params object.

src/amber/router/router.cr
```diff
@@ -84,10 +84,6 @@ module Amber
         end
       end
 
-      private def merge_params(params, context)
```
eliasjpr: This method was never used

src/amber/cli/templates/error/src/handlers/error.cr.ecr
```diff
@@ -6,7 +6,7 @@ module Amber
     class Error < Base
       def call(context : HTTP::Server::Context)
         begin
-          raise Amber::Exceptions::RouteNotFound.new(context.request) if context.invalid_route?
+          raise Amber::Exceptions::RouteNotFound.new(context.request) unless context.valid_route?
```
docelic: Nice one!

spec/amber/router/params_spec.cr
```diff
@@ -0,0 +1,84 @@
...
+        it "returns params" do
+          headers["Content-Type"] = "application/x-www-form-urlencoded"
+          body = "name=John Doe"
+          request = HTTP::Request.new("POST", "/", headers, body)
+          request.params["name"].should eq "John Doe"
+        end
+
+        it "parses body params with charset" do
+          headers["Content-Type"] = "application/x-www-form-urlencoded; charset=UTF-8"
+          body = "\x68\x65\x6C\x6C\x6F\x3D\x77\x6F\x72\x6C\x64"
```
elorest: Use `Bytes[104, 101, 108, 108, 111, 61, 119, 111, 114, 108, 100]` instead of broken string. @RX14 and <Oprypin> are pretty adamant that non utf8 strings should not be used.

src/amber/router/params.cr
```diff
@@ -1,20 +1,14 @@
...
-  class Params < Hash(String, String)
-    alias KeyType = String | Symbol
-    property files = {} of String => Amber::Router::Files::File
+  struct Params
+    property files = {} of String => File
+    property store : HTTP::Params = HTTP::Params.new(Hash(String, Array(String)).new)
 
-    def json(key : KeyType)
-      JSON.parse(self[key]?.to_s)
+    forward_missing_to @store
```
elorest: There was some stuff we had talked about here that I was going to do later. I'll do that as a different PR after this is merged.

src/amber/controller/error.cr
elorest: I do think this should be fixed in a different PR.

spec/amber/router/params_spec.cr
```diff
@@ -0,0 +1,84 @@
...
+module Amber::Router
+  describe Parser do
+    headers = HTTP::Headers.new
+
+    context "when parsing query string" do
+      it "returns query params" do
+        request = HTTP::Request.new("GET", "/?test=test")
+        request.params["test"].should eq "test"
+      end
+
```
docelic: @eliasjpr it seems that here a test is missing to show what happens when request.params["test"] is called but contains multiple values.

eliasjpr: See next test below. Let me know if that answers your question👇

docelic: In the next one you call fetch_all(), not params[]. I am saying that it seems a test is missing where you do fill it with multiple params, but then call params[] instead of fetch_all.

elorest: I don't get this. `request.params["test"].should eq "test2"` in that case. Is that a thing that needs retested?

eliasjpr: I see what you're saying, I have updated the test below

spec/amber/router/params_spec.cr
```diff
@@ -0,0 +1,81 @@
...
+          body = "\x68\x65\x6C\x6C\x6F\x3D\x77\x6F\x72\x6C\x64"
+          request = HTTP::Request.new("POST", "/", headers, body)
+          request.params["hello"].should eq "world"
+        end
+      end
+
+      context "when parsing multipart" do
+        it "parses files from multipart forms" do
+          headers["Content-Type"] = "multipart/form-data; boundary=fhhRFLCazlkA0dX"
+          body = "--fhhRFLCazlkA0dX\r\nContent-Disposition: form-data; name=\"_csrf\"\r\n\r\nPcCFp4oKJ1g-hZ-P7-phg0alC51pz7Pl12r0ZOncgxI\r\n--fhhRFLCazlkA0dX\r\nContent-Disposition: form-data; name=\"title\"\r\n\r\ntitle field\r\n--fhhRFLCazlkA0dX\r\nContent-Disposition: form-data; name=\"picture\"; filename=\"index.html\"\r\nContent-Type: text/html\r\n\r\n<head></head><body>Hello World!</body>\r\n\r\n--fhhRFLCazlkA0dX\r\nContent-Disposition: form-data; name=\"content\"\r\n\r\nseriously\r\n--fhhRFLCazlkA0dX--"
```
robacarp: I think a heredoc would make this easier to read, even if the `\r` is preserved:

```
--fhhRFLCazlkA0dX\r
Content-Disposition: form-data; name="_csrf"\r
\r
PcCFp4oKJ1g-hZ-P7-phg0alC51pz7Pl12r0ZOncgxI\r
--fhhRFLCazlkA0dX\r
Content-Disposition: form-data; name="title"\r
\r
title field\r
--fhhRFLCazlkA0dX\r
Content-Disposition: form-data; name="picture"; filename="index.html"\r
Content-Type: text/html\r
\r
<head></head><body>Hello World!</body>\r
\r
--fhhRFLCazlkA0dX\r
Content-Disposition: form-data; name="content"\r
\r
seriously\r
--fhhRFLCazlkA0dX--
```

Related, The spec seems to indicate that it's not proper but the HTTP body doesn't _need_ the `\r`. Do we handle parsing the raw http, or is Crystal? Do we need a test around parsing a request body with only `\n`?

spec/amber/router/params_spec.cr
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      context "when parsing JSON body" do
+        it "parses json hash" do
+          headers["Content-Type"] = "application/json"
+          body = %({ "test": "test", "address": { "city": "New York" }})
+          request = HTTP::Request.new("POST", "/", headers, body)
+
+          request.params["test"].should eq "test"
+          request.params.json("address")["city"].should eq "New York"
```
robacarp: Is `.json` on a params object new? That's nice.

spec/amber/router/params_spec.cr
```diff
@@ -0,0 +1,81 @@
...
+          request = HTTP::Request.new("GET", "/fake/action/123/john")
+
+          request.route
+
+          request.params["id"].should eq "123"
+          request.params["name"].should eq "john"
+        end
+      end
+    end
+  end
```
robacarp: I think it would be good to add a test for how different parameters can override each other, so it's documented to not change accidentally.

eg, what happens when this route: `/objects/:id/:name` is hit with this url: `/objects/23/foo?name=bar`?

src/amber/router/file.cr
```diff
@@ -0,0 +1,26 @@
...
+    getter file : Tempfile
+    getter filename : String?
+    getter headers : HTTP::Headers
+    getter creation_time : Time?
+    getter modification_time : Time?
+    getter read_time : Time?
+    getter size : UInt64?
+
+    def initialize(upload)
+      @filename = upload.filename
```
robacarp: Is `upload.filename` direct from end user form submission? Is naming a file on the disk by a string provided from some unknown source a dangerous idea? Could this be changed so that the tempfile has a uuid name, but the `Router::File#filename` method still returns the name that was sent in the request?

src/amber/router/params.cr
```diff
@@ -1,20 +1,14 @@
...
+
+    def json(key)
+      JSON.parse(store[key]?.to_s)
```
robacarp: I have a few questions about this.

- This seems like it's going to run once for each `parms.json`. Can it make a memo somewhere and only do that once?
- Is the idea here that a part of the params would be json but the rest wouldn't? I'm not sure I understand.
- Can this failure case be handled more gracefully? I think there should be a more gentle way to signal to a framework user that the json payload is invalid than raising an anonymous exception. Is that something that can be added?

src/amber/router/request.cr
```diff
@@ -0,0 +1,55 @@
+require "./params"
+require "./parser"
+require "./file"
+
+class HTTP::Request
```
robacarp: This is out of the scope of this pull, I think, but monkeypatching `HTTP::Request` makes me uncomfortable. I'd much rather create a wrapper object around it and delegate into `HTTP::Request` as needed. Doing this is hostile to having more than one http framework in the same project.

src/amber/router/parsers/form_data.cr
```diff
@@ -0,0 +1,19 @@
+module Parsers
```
robacarp: All of these parsers should be nested under the `Amber::Router::Parser` module instead of just `Parsers` to minimize leakage outside of the namespace.

src/amber/router/parser.cr
```diff
@@ -0,0 +1,48 @@
...
+      end
+    end
+
+    private def multipart
+      return unless content_type.try &.starts_with? MULTIPART_FORM
+      Parsers::Multipart.parse(@params, self)
+    end
+
+    private def json
+      return unless content_type.try &.starts_with? APPLICATION_JSON
```
robacarp: why `starts_with` instead of just `==`?

src/amber/router/parser.cr
```diff
@@ -0,0 +1,48 @@
...
+    MULTIPART_FORM   = "multipart/form-data"
+    APPLICATION_JSON = "application/json"
+    @parsed = false
+
+    def params
+      return @params if @parsed
+      parse_url_params
+      form_data
+      json
+      multipart
```
robacarp: The order here is really delicate, and I think it might be off. Does this meansthat url parameters, which are defined in a `routes.cr` file will get overwritten by form data, json, and multipart data sources? That feels like an attack surface to me.

src/amber/router/parser.cr
eliasjpr: I actually thought a regex here would be better. But you're also correct. Thinking `regex` vs `==`...

src/amber/router/parsers/form_data.cr
eliasjpr: Yes missed that. Good observation

src/amber/router/request.cr
eliasjpr: Please provide some examples. The reason why I went this route was to avoid initializing an extra class on every request it seems more of an overhead to accomplish the same, the only method that get overriden is the `method` everything else is Amber specific. I wish there was a way to just inherit from HTTP::Request and have HTTP::Context take the new request object but that's like rewriting your HTTP library again..

src/amber/router/file.cr
docelic: @robacarp @eliasjpr  Right, it would be nice to use a tempfile because what if the original filename is maliciously reported as e.g. "../../../../../tmp/myfile"? It would allow one to save it to an unintended directory.

But more importantly - why is a temp file being saved at all, instead of the file contents persisting in memory until the user decides to save them somewhere (or not save them at all)?

eliasjpr: The reason to write the file to disk is to offload the server memory. Think what would happened if your crystal app is running on 1gb of ram and and you upload a 2gb file? That will use all the ram available to the server. Also the temp directory is the default behavior for most web servers I do see some good reasons why rename the uploaded file on server

docelic: @eliasjpr yes, but the file contents already are in memory. The data transfer of file contents won't begin when you save the file, but it has already been completed and the whole contents are in memory. Or I am wrong on that?

spec/amber/router/params_spec.cr
elorest: In HTTP blocks are separated by `\r\n\r\n`

src/amber/router/params.cr
elorest: @eliasjpr and I have already talked about changes I'm going to make to this class to bring it back in line with what is currently in Amber. We can address this there too.

src/amber/router/request.cr
elorest: I agree with @eliasjpr here. HTTP::Request is initialized by HTTP::Server, so if we were to create a wrapper object we'd have to monkey patch http server as well.

src/amber/router/parser.cr
elorest: Why is @params and instance variable. Seems sort of weird since it's defined outside of this scope.

src/amber/router/parsers/multipart.cr
```diff
@@ -0,0 +1,15 @@
+module Parsers
+  module Multipart
+    def self.parse(params : Amber::Router::Params, request : HTTP::Request)
+      HTTP::FormData.parse(request) do |upload|
+        next unless upload
+        filename = upload.filename
+        if filename.is_a?(String) && !filename.empty?
+          params.files[upload.name] = Amber::Router::File.new(upload: upload)
+        else
+          params[upload.name] = upload.body.gets_to_end
```
elorest: This needs to be `params.add(upload.name, upload.body.gets_to_end)` or else it will overwrite params that have already been defined instead of appending.

src/amber/router/request.cr
```diff
@@ -0,0 +1,55 @@
...
+  include Amber::Router::Parser
```
elorest: Why include all of this into request when we could just call `Amber::Router::Parser.params` as a class method and have it generate params. Which we could save here in the memo @params. That way call chains are partitioned, variable definitions stay within scope and code follows SOLID principles.

src/amber/router/file.cr
eliasjpr: I don't think keeping file in memory is advisable or good practice. Reasons that I see is:

- One form can have multiple files to upload
- File uploads can be streamed (Chunked uploads)
- https://github.com/rack/rack/blob/master/lib/rack/multipart/uploaded_file.rb#L14

src/amber/router/request.cr
eliasjpr: I had it that way :)

src/amber/router/request.cr
```diff
@@ -0,0 +1,58 @@
...
+    else                    @method
+    end
+  end
+
+  def requested_method
+    @requested_method ||= params[METHOD]? || headers[OVERRIDE_HEADER]? || @method
+  end
+
+  def params
+    @params ||= Amber::Router::Parser.params(self)
```
elorest: Awesome. Sorry for the confusion. I never meant for you to change that.

src/amber/router/parser.cr
```diff
@@ -0,0 +1,60 @@
...
+    TYPE_EXT_REGEX   = Amber::Support::MimeTypes::TYPE_EXT_REGEX
+    URL_ENCODED_FORM = "application/x-www-form-urlencoded"
+    MULTIPART_FORM   = "multipart/form-data"
+    APPLICATION_JSON = "application/json"
+    @parsed = false
+    @request : HTTP::Request
+    @params : Amber::Router::Params = Amber::Router::Params.new
+
+    def self.params(request : HTTP::Request)
+      new(request).params
```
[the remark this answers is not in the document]
elorest: @parsed doesn't matter as request won't call `Amber::Router::Parser` unless there is no memo for @params.

spec/amber/router/params_spec.cr
robacarp: Yeah, fetch all is fine, but I'm asking about the regular fetch.

What happens when you call `params["name"]` in a `/objects/:id/:name` action that was triggered with a url like this: `/objects/23/foo?name=bar`

I think there should be a test defining what happens when a parameter is shadowed, because I don't expect the majority of use cases to use list fetch with `fetch_all`.

src/amber/router/params.cr
robacarp: Does that mean that this will be a breaking change now, and also later when your changes merge in? Why not just address the issue and not have any broken state?

spec/amber/router/params_spec.cr
elorest: Apparently with `HTTP::Params` fetch returns the first one added rather than the last one. Where as fetch_all returns the array.

src/amber/router/file.cr
elorest: Writing file to tempfile is pretty standard. In regard to the name of the file Tempfile already handles adding a random string for the filename. https://crystal-lang.org/api/0.23.0/Tempfile.html

In terms of storing, upload.body in memory, I don't think that would be very useful, and I'd be afraid of memory usage increasing out of control as unclosed connections built up. As it currently stands the request has this already but storing the parsed multipart file in memory would over double the memory per request.... WDYGT?

src/amber/router/params.cr
elorest: It's in master so it won't be released as a breaking change. As it stands I guess it is a tiny bit. For instance before this params[] could take a symbol or string as key. This removes the functionality until I add it back.

elorest: > This seems like it's going to run once for each parms.json. Can it make a memo somewhere and only do that once?

This is to handle the case where json params go multiple levels deep. That can't be store in the normal params as JSON::TYPE so we store the nested values as a string that is parsed if you need it.

It would be possible to create `@json_params : Hash(String, JSON::Type)` which could then be set when `.json `is called.

```cr
def json(key)
  @json_params[key] ||= JSON.parse(store[key]?)
end
```

spec/amber/router/params_spec.cr
robacarp: Well, that breaks my mind, but the standard is decidedly mute on the subject, so I suppose that's just what the crystal folks picked when they wrote it.

elorest: In most cases it's bad practice to have multiple params the same key anyway. One exception being selects with multiple values.

elorest: In most cases it's bad practice to have multiple params the same key anyway. One exception being selects with multiple values.

robacarp: Sorry, was confused about the line we were talking about.

src/amber/router/file.cr
robacarp: If tempfile is already randomizing, we should just not use the file name at all. It’s not worth the potential risk.

elorest: This isn't actually even a new change it just shows that way because @eliasjpr moved a bunch of files around and renamed them. It's useful to store the filename so that when doing something with it in the controller we know what it was originally called.

src/amber/router/file.cr
```diff
@@ -0,0 +1,26 @@
...
+      @file = Tempfile.new(filename)
```
elorest: @robacarp on this line however it could make sense to just have `@file = Tempfile.new`.

eliasjpr:
im closing this PR. It was brought to my attention that the work here brings too many unnecessary changes and it is not backward compatibly. I thought a refactor was necessary given the structure of params.

PR #642, opened 2018-02-11
[Params] Refactors Params into Modules

eliasjpr:
Refactors Params into Modules

Issue  #592

> As a developer I want to parse params and return an array of values
for a given param key.

The changes here introduces a Parser module that handles parsing query
params. It organizes each parser into its own class.

List of changes:
- Adds Params Specs
- Adss params class
- Moves Amber::Router::Files::File to Amber::Router::File
- Allows to parse form data from request body
- Allows for parsing form Multipart form data
- Allows for JSON request body parsing
- Adds a Request.cr file 
- Removes the original ParamsParser in favor of the new Parser
- Cleans context spec 
- Params are parsed only once per request (Adds parse state to memoize)
- Removes redundant Request Handler Method now is request.route
- Repurpose invalid_route? to valid_route? (the logic in the method is easier to read and understand)
- [Controller] Error fixes error controller (This was also addressed by @marksiemers in another PR) I made the change here because it was hindering development.

These changes allows for lazy loading params in instances where  there is no need to override method, and matching Radix route, with this change we can work on allowing Pipelines to run before  matching a route.

Thanks @elorest for the pairing session for shedding some light on the module and Params struct


src/amber/router/params.cr
```diff
@@ -1,20 +1,104 @@
...
+
+    def json(key : Types::Key)
       JSON.parse(self[key]?.to_s)
     rescue JSON::ParseException
       raise "Value of params.json(#{key.inspect}) is not JSON!"
     end
 
-    def []=(key : KeyType, value : V)
-      super(key.to_s, value)
+    def override_method?(key : Types::Key)
```
robacarp: I don't think I've ever considered sending json with an `_method: "DELETE"` field, is that a common paradigm?

src/amber/router/request.cr
```diff
@@ -0,0 +1,55 @@
...
+class HTTP::Request
+  METHOD          = "_method"
+  OVERRIDE_HEADER = "X-HTTP-Method-Override"
+
+  @radix_route : Radix::Result(Amber::Route)?
+  @requested_method : String?
+  @params : Amber::Router::Params?
+
+  def method
+    case @method
```
robacarp: Can this be memoized? I could see this code running several times in a request.

jwaldrip:
Why not just reset the method on `HTTP::Request`?

jwaldrip:
Similar to here: https://github.com/obsidian/orion/blob/master/src/orion/handlers/method_override_param.cr

src/amber/router/params.cr
```diff
@@ -1,20 +1,104 @@
...
+    alias Params = Hash(String, String)
+  end
+
+  class Params
+    TYPE_EXT_REGEX   = Amber::Support::MimeTypes::TYPE_EXT_REGEX
+    URL_ENCODED_FORM = "application/x-www-form-urlencoded"
+    MULTIPART_FORM   = "multipart/form-data"
+    APPLICATION_JSON = "application/json"
+
+    getter empty = Types::Params.new
```
elorest: I don't like calling this empty. 

Also I think we should use HTTP::Params instead of Hash(String, String).
After all that was the reasoning behind adding it as a property it instead of inheriting from it in the first place.

eliasjpr:
@elorest 

I looked at other implementations and I can see why that kind of approach was taken. 

The decision of creating a Params class that does not inherit from a Hash or HTTP::Params is so to have a common API that enables access all kind of params regardless where the params live (Url, Form, Multipart, Body [XML, JSON]). Params class behaves like a Hash

I have failed to identify the breaking changes. Can you please point out the breaking changes you have found? I am sure those could be addressed as well.

src/amber/router/params.cr
```diff
@@ -1,20 +1,104 @@
...
+      @request.query_params
+    end
+
+    private def form
+      return HTTP::Params.parse("") unless content_type?(URL_ENCODED_FORM)
+      @form ||= Parsers::FormData.parse(@request)
+    end
+
+    private def multipart
+      return @multipart.not_nil! if @multipart
```
elorest: Rather than returning here it would be faster (fewer branches) to just not call multipart if @multipart exits. I posted this above where this is called as well.

src/amber/router/params.cr
```diff
@@ -1,20 +1,104 @@
...
+      self.[key]? || raise Amber::Exceptions::Validator::InvalidParam.new(key)
+    end
+
+    def []?(key : Types::Key)
+      _key = key.to_s
+      route[_key]? || override_method?(_key) || json[_key]?
+    end
+
+    def files
+      multipart
```
elorest: Check `if @multipart` before calling this is faster and more readable than doing the check in the method.

src/amber/router/params.cr
```diff
@@ -1,20 +1,104 @@
...
+    alias Params = Hash(String, String)
+  end
+
+  class Params
+    TYPE_EXT_REGEX   = Amber::Support::MimeTypes::TYPE_EXT_REGEX
+    URL_ENCODED_FORM = "application/x-www-form-urlencoded"
+    MULTIPART_FORM   = "multipart/form-data"
+    APPLICATION_JSON = "application/json"
+
+    getter empty_params = Types::Params.new
```
elorest: Seems a little strange to initialize this as an attribute when it might not even be used?

Would it be cleaner to just return `{} of Types::Params` or `{} of Hash(String, String)`?

To me those are much more readable.

spec/amber/router/params_spec.cr
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    describe "#key?" do
+      it "returns nil for non-existent key" do
+        params.key?("invalid").should be_nil
+      end
+
+      it "retuns value for existing key" do
+        params.key?("test").should eq "test"
```
robacarp: This seems to diverge from the crystal functionality for `Hash#key?`, which looks up a key for a given value. That feels like it might be confusing for people.

spec/amber/router/request_spec.cr
```diff
@@ -0,0 +1,127 @@
...
+        it "overrides form POST method to PUT, PATCH, DELETE" do
+          headers[HTTP::Request::OVERRIDE_HEADER] = method
+          request = HTTP::Request.new("POST", "/?test=test", headers)
+          request.method.should eq method
+        end
+
+        it "overrides form GET method to PUT, PATCH, DELETE" do
+          headers[HTTP::Request::OVERRIDE_HEADER] = method
+          request = HTTP::Request.new("GET", "/?test=test", headers)
+          request.method.should eq method
```
robacarp: really? I thought we decided not to do overrides with GET requests

spec/amber/router/request_spec.cr
```diff
@@ -0,0 +1,127 @@
...
+        end
+
+        it "takes form post over header override" do
```
robacarp: I don't understand how this `it` matches what's being tested below. Can you clarify a little?

spec/amber/router/params_spec.cr
eliasjpr: Interesting. Didn't knew this from crystal. I guess I should use `has_key?`

spec/amber/router/request_spec.cr
eliasjpr: Im fine we that. I can remove this, it will simplify the method

spec/amber/router/request_spec.cr
eliasjpr: Basically the request sends an override header and a override field in a form post, and the expectation is that it takes the form post over the header. 

This sends  the override method in the request of the body `request = HTTP::Request.new("GET", "/?test=test", headers, "_method=PATCH")` while `headers[HTTP::Request::OVERRIDE_HEADER]` sets the headers for the request

spec/amber/router/params_spec.cr
```diff
@@ -0,0 +1,61 @@
...
+      end
+
+      it "returns form data params" do
+        params["title"].should eq "title field"
+        params["_csrf"].should eq "PcCFp4oKJ1g-hZ-P7-phg0alC51pz7Pl12r0ZOncgxI"
+      end
+
+      it "raises error for non existent param" do
+        expect_raises Amber::Exceptions::Validator::InvalidParam do
+          params["invalid"]
```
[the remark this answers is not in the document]
eliasjpr: Yes there should be tests for it 😊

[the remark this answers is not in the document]
eliasjpr: See 👇🏻 I recently added the tests

PR #647, opened 2018-02-13
[Travis] Adds Dist Trusty and Sudo Required

eliasjpr:
### Description of the Change

As Per Travis CI docs 

> When specifying sudo: required, Travis CI runs each build in an isolated Google Compute Engine virtual machine that offer a vanilla build environment for every build.

This has the advantage that no state persists between builds, offering a clean slate and making sure that your tests run in an environment built from scratch.

Builds have access to a variety of services for data storage and messaging, and can install anything that’s required for them to run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ language: generic
 services:
   - docker
 
-before_install:
-  - docker build -t garnet-spec .
-
 env:
   - TEST_SUITE=./spec/build_spec.cr
   - TEST_SUITE=./spec/amber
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,13 +8,13 @@ services:
 before_install:
   - docker build -t garnet-spec .
```
drujensen: Do we need this?  I don't see `garnet-spec` image being referenced anywhere.

eliasjpr:
@robacarp yes, since the DB seemed to be shared in the build

PR #649, opened 2018-02-16
add toGranite() to Date prototype

damianham:
### Description of the Change

Adds a convenience method to the Date prototype to format date time strings according to the format expected by Granite ORM.

### Alternate Designs
None considered
### Benefits

Makes it easier to send a date time string from a javascript frontend to Amber with Granite ORM, e.g.

```javascript
  let toptions = {_csrf: this.csrf_token,
        eta: this.active_waypoint.eta.toGranite(),  // format the date in the format expected by Granite
        bearing: Math.round(this.active_waypoint.bearing),
        distance: Math.round(this.active_waypoint.range)
      };

      jqxhr = $.ajax(
        {
          url: "/waypoints/progress/"+this.active_waypoint.id,
          method: 'PUT',
          data: toptions
        })
        .done(function(data) {
          //console.log('updated waypoint OK', data)
        })
        .fail(function(data) {
          console.log('error updating waypoint', toptions, data)
        })
```

### Possible Drawbacks

little bit of javascript code bloat.


src/amber/cli/templates/app/src/assets/javascripts/main.js
```diff
@@ -1 +1,23 @@
...
+
+if (!Date.prototype.toGranite) {
+  (function() {
+
+    function pad(number) {
+      if (number < 10) {
+        return '0' + number;
+      }
+      return number;
+    }
```
robacarp: thank you for not including `left_pad` for this ;-)

src/amber/cli/templates/app/src/assets/javascripts/main.js
```diff
@@ -1 +1,23 @@
...
+
+    Date.prototype.toGranite = function() {
+      return this.getUTCFullYear() +
+        '-' + pad(this.getUTCMonth() + 1) +
+        '-' + pad(this.getUTCDate()) +
+        ' ' + pad(this.getUTCHours()) +
+        ':' + pad(this.getUTCMinutes()) +
+        ':' + pad(this.getUTCSeconds())  ;
+    };
```
robacarp: Does javascript not have a strfmt? I'm surprised javascript doesn't have an iso-8601 formatter that's compatible with granite.

src/amber/cli/templates/app/src/assets/javascripts/main.js
```diff
@@ -1 +1,23 @@
 import Amber from 'amber'
+
+if (!Date.prototype.toGranite) {
```
robacarp: I spend very little time javascripting so perhaps my instincts are off, but this method name feels weird to me. It's not emitting a Granite, it's emitting a formatted string representation of a datetime.

src/amber/cli/templates/app/src/assets/javascripts/main.js
drujensen: https://www.npmjs.com/browse/keyword/strftime

Which one will not be another `left_pad`?

robacarp: [how about this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString)

src/amber/cli/templates/app/src/assets/javascripts/main.js
drujensen: @robacarp I agree.  This should be a more generic name.  `toIso8601()` or `toDbFormat()`?

src/amber/cli/templates/app/src/assets/javascripts/main.js
damianham: Javascript Date has toISOString() but it generates a string with a T between the date and time.  I often use moment library to format dates in the presentation layer but of course I didn't want to introduce an unnecessary dependency when a simple function can do the job.

src/amber/cli/templates/app/src/assets/javascripts/main.js
damianham: copied it from the polyfill for toISOString() at MDN - always try to follow the experts when I can :-)

src/amber/cli/templates/app/src/assets/javascripts/main.js
damianham: I didn't check the code for the other ORMs for Amber, I am using granite ORM for my project, but if they all expect the same time-date format then I agree toDbFormat() or toAmberFormat() would be a better name. I think toISO8601() is slightly confusing because 8601 format is 2018-02-15T12:00:00 with the 'T' separator and also may be confused with toISOString().  I did originally call it toGraniteString() but thought the 'String' was redundant because if you use this patched in method you will be aware of what you are creating. I can rename it to whatever you think is the best name.

robacarp: Perhaps Granite should learn to parse datetimes from one of the iso8601 formats.

damianham: yes that would be a good option

PR #652, opened 2018-02-18
Add citrine i18n

docelic:
This commit adds:

1) citrine-i18n initializer config file
2) Citrine::Pipe::I18n pipe for locale detection
3) Minimal locale example in src/locales/en.yml
4) "t" and "l" shorthands for ::I18n.translate and ::I18n.localize
5) "Welcome to Amber Framework!" which now comes from a translation

What's missing is:

Maybe we should generate basic locale files (or just empty directories) in src/locales/{models,views,...}/ when these types are generated

This could be done in a separate commit/PR.

### DIFF
--- a/spec/amber/cli/commands/pipelines/pipelines_spec.cr
+++ b/spec/amber/cli/commands/pipelines/pipelines_spec.cr
@@ -22,6 +22,7 @@ module Amber::CLI
 
         web_default_plugs = %w(
           Amber::Pipe::PoweredByAmber
+          Citrine::Pipe::I18n
           Amber::Pipe::Error
           Amber::Pipe::Logger
           Amber::Pipe::Session
--- a/src/amber/cli/templates/app/config/initializers/i18n.cr
+++ b/src/amber/cli/templates/app/config/initializers/i18n.cr
@@ -1,0 +1,24 @@
+require "citrine-i18n"
+
+Citrine::I18n.configure do |settings|
+  # Backend storage (as supported by i18n.cr)
+  # settings.backend = I18n::Backend::Yaml.new
+
+  # Default locale (defaults to "en" and "./src/locales/en.yml").
+  # For a new default locale to be accepted, it must be found by the
+  # backend storage and reported in "settings.available_locales".
+  # settings.default_locale = "en"
+
+  # Separator between sublevels of data (defaults to '.')
+  # e.g. I18n.translate("some/thing") instead of "some.thing"
+  # settings.default_separator = '.'
+
+  # Returns the current exception handler. Defaults to an instance of
+  # I18n::ExceptionHandler.
+  # settings.exception_handler = ExceptionHandler.new
+
+  # The path from where the translations should be loaded
+  settings.load_path += ["./src/locales"]
+end
+
+I18n.init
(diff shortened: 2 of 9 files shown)


spec/amber/cli/commands/pipelines/pipelines_spec.cr
```diff
@@ -22,6 +22,7 @@ module Amber::CLI
 
         web_default_plugs = %w(
           Amber::Pipe::PoweredByAmber
+          Citrine::Pipe::I18n
```
eliasjpr: Lets have this as `Citrine::I18n` since this is it's own shard that can be used independently it should not inherit from Amber

src/amber/cli/templates/app/config/initializers/i18n.cr
```diff
@@ -0,0 +1,24 @@
...
+  # Separator between sublevels of data (defaults to '.')
+  # e.g. I18n.translate("some/thing") instead of "some.thing"
+  # settings.default_separator = '.'
+
+  # Returns the current exception handler. Defaults to an instance of
+  # I18n::ExceptionHandler.
+  # settings.exception_handler = ExceptionHandler.new
+
+  # The path from where the translations should be loaded
+  settings.load_path += ["./src/locales"]
```
eliasjpr: 🥇

src/amber/cli/templates/app/config/routes.cr
```diff
@@ -4,6 +4,7 @@ Amber::Server.configure do |app|
     # A plug accepts an instance of HTTP::Handler
     plug Amber::Pipe::PoweredByAmber.new
     # plug Amber::Pipe::ClientIp.new(["X-Forwarded-For"])
+    plug Citrine::Pipe::I18n.new
```
eliasjpr: Ditto

spec/amber/cli/commands/pipelines/pipelines_spec.cr
docelic: Hey,
I've modified citrine-i18n to not depend on Amber now (it simply includes HTTP::Handler instead of subclassing Amber::Pipe::Base).

In terms of pipe name, I kept it named Citrine::Pipe::I18n for now because it would be hard to rename pipe class to Citrine::I18n (because Citrine::I18n is already a module).

Are you suggesting that I rename Citrine::Pipe::I18n to Citrine::I18n::Pipe, or something else?

eliasjpr: > Are you suggesting that I rename Citrine::Pipe::I18n to Citrine::I18n::Pipe, or something else?

I think `Citrine::Pipe::I18n` is fine now that I understand

docelic: Yes, but your comment was on-point -- it makes no sense that citrine-i18n puts everything under Citrine::I18n, but then puts pipe in Citrine::Pipe::I18n.
Also, "Pipe" is Amber-specific term and you said that the shard should not be amber specific.

So I suggest that we rename the pipe to Citrine::I18n::Handler (or Citrine::I18n::Pipe if need be). Any thoughts from your side on that?

eliasjpr: I thought the exact same name :) `Citrine::I18n::Handler` go for it. I didnt want for you to make more changes.

eliasjpr:
@docelic go for it name it  `Citrine::I18n::Handler` I thought of the same

Issue #605, opened 2018-02-01
Add --license option to compiled project to be compliant with MIT


elorest:
### Description

We need a way to display licenses for compiled projects. 

Currently almost every crystal shard is under MIT license which requires the license to be distributed along with everything that uses it. This is easy in ruby since all gems need to be local on the machine that a project is running on. Unfortunately with crystal since we could distribute pre-compiled binaries, licenses are no longer of necessity present on all systems that uses them (this is a good thing but can easily lead to License violation.

My suggestion would be to compile in an option that displays all licenses.

Alternatively we could copy LICENSE files out of `/lib/[projects]` into `licenses/` or something.

faustinoaq:
```
> ./myamberapp --licenses
Amber Framework   - https://github.com/amberframework/amber/blob/master/LICENSE
Granite ORM       - https://github.com/amberframework/granite-orm/blob/master/LICENSE
Postgres Adapter  - https://github.com/will/crystal-pg/blob/master/LICENSE
# ...
Spinner           - https://github.com/askn/spinner/blob/master/LICENSE
```

WDYT?

PR #654, opened 2018-02-19
[WIP] Adds --license option.

elorest:
### Description of the Change

Adds `--license` option so that you can compile locally and then push the binary to a server, embedded system, etc and have it run without having to include all code and deps.

Resolves Issue: https://github.com/amberframework/amber/issues/605


eliasjpr:
I think this is great and probably would be better as a Shard so other people can use this outside of amber.  This also seems a thing that Crystal and Shards should be supporting. I would propose to talk to Crystal and Shards team to build  in this feature if not move it to its own shard

WDYT?

elorest:
Yeah i was actually thinking the same thing after I went to bed last night.

elorest:
Moved code to shard and refactored to use shard instead of built in code. Specs are in shard.

PR #656, opened 2018-02-19
[Travis] Removes Docker from Travis Ci Builds

eliasjpr:
### Description of the Change

After many attempts to solver the issue where docker does not connect to
Redis. I like to propose that we remove the Docker from the Travis
build, since travis supports redis and postgres databases without the
need of using docker compose.

### Benefits
A couple of benefits to this:

- Faster Build Times each built takes only ~3 minutes vs 5 mminutes
- Simple configuration for running tests on Travis and removes a  layer of
complexitiy
- For those who like do not want to have a Redis and or Postgress
installation to run Amber Specs they can still use Docker and Docker Compose.

With this change we should experience more consistency running Specs on
Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,21 @@
 dist: trusty
 sudo: required
-language: generic
+language: crystal
 
 services:
-  - docker
-
+  - postgres
+  - redis-server
+cache:
+  directories:
+    - bin/
+    - lib/
 env:
   - TEST_SUITE=./spec/build_spec.cr
   - TEST_SUITE=./spec/amber
 
 script:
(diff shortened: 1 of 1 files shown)


eliasjpr:
Downside? I am not sure honestly since we have a failing docker setup for Travis at the moment that is causing a lot of issues. I spent some time to try to fix it but I ran out of steam. 

My question is: Is docker something that we should be concern of testing, is that something a framework is responsible for? IMHO I don't think so.

drujensen:
@eliasjpr The biggest downside is when a new version crystal comes out.

Issue #240, opened 2017-09-08
[Router] Add `options` routes when CORS is used


fridgerator:
When building sites with SPA's, [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) are sent before all requests.  Currently the only way to handle this is to create separate`options` routes for each route used and a middleware to immediately return a 200 response:

```ruby
resources "/lists", ListController
options "/lists", ListController, :index
options "/lists/:id", ListController, :show
```

I haven't looked at how other frameworks handle this, but a user shouldn't have to also include options routes when using CORS.

drujensen:
@fridgerator Kemal and Kemalyst allows for handlers to happen before the router.  This is an example of where that comes in handy since you can look for `options` or `head` requests for all requests and properly handle them before requiring a route to be looked up.

@eliasjpr @elorest What do you think?  Should we look at allowing handlers to be added before the pipeline takes hold?

eliasjpr:
I consider that a HEAD request should run thru the entire pipeline since a header can be set at the controller action level or after. 

As per the OPTIONS this needs to be handled by the CORS handler and if the current request is a `preflight` request it should return as soon as the CORS pipe is hit. So it is likely to recommend that this pipe should be the first on the list for a given pipeline. Also I think we should add the Options requests by default unlike having the developer define those routes.

```
pipeline :api do
   plug CORS.new
   ...
end 
```

NeverHappened:
Tried to add options route as long as developer uses the CORS pipeline.
Not sure if this is the right way though.

https://github.com/amberframework/amber/pull/659

eliasjpr:
@NeverHappened thank you for contributing to this issue. I found your approach viable given the way things are structured. I was considering to just white-list any OPTIONS requests considering that these type of requests are issued in the background by the web browser in the background and are practically invisible to the developer. So if an options requests is received and the handler is not present the request does not get handled at all, if the CORS handler is present it then gets processed accordingly. 

I have made some improvements to the CORS handler here https://github.com/amberframework/amber/pull/665 and would like your input and review. :)

NeverHappened:
Cool :)

The only issue I think about is how does it work since request first goes through the
```crystal
@route = router.match_by_request(@request)
```
 and then if it matches it goes to the pipeline? There are no *options* route so it will fail. Or didn't I understand something?

eliasjpr:
[...]

PR #659, opened 2018-02-22
Add options routes for routes with CORS pipeline in it

NeverHappened:
### Description of the Change

Add the options routes for routes with CORS pipe

### Alternate Designs

Thought about other ways for doing this, but they were worse or impossible.
Maybe missed something.

### Possible Drawbacks

I don't like that I added dependency between router and pipeline handlers. Couldn't thought of any other way.
Also the function that I call is in the server.cr call, the main point of app, kinda becomes polluted with unnecessary details.


faustinoaq:
Fixes #240

eliasjpr:
Closing this PR in favor of https://github.com/amberframework/amber/pull/700

PR #662, opened 2018-02-24
Run migrations at the start of a test run

### LINKED ISSUE #646: Should generated apps run migrations against the test database at the start of a spec run?
### Description

I have this code in my spec_helper and it's saved me a ton of headache:

```crystal
require "micrate"
Micrate::DB.connection_url = Amber.settings.database_url
Micrate::Cli.run_up
```

Is that something worth adding to the generator?

robacarp:
### Description of the Change

Fixes #646, default generated apps to run migrations at the start of a test run.

### Benefits

When building a database with migrations, it's currently necessary to run migrations against the test environment before running tests.

Something like this: `AMBER_ENV=test amber db migrate`

### Possible Drawbacks

There is some risk to automatically running migrations. If someone manually overrides the ENV for some reason when running tests:

```
$ export AMBER_ENV=production
$ crystal spec # would run migrations
```

### DIFF
--- a/src/amber/cli/templates/app/spec/spec_helper.cr
+++ b/src/amber/cli/templates/app/spec/spec_helper.cr
@@ -4,6 +4,11 @@ require "spec"
 require "garnet_spec"
 require "../config/*"
 
+require "micrate"
+Micrate::DB.connection_url = Amber.settings.database_url
+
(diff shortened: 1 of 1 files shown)


robacarp:
Hm, I think this also fixes the test failures in the build spec

PR #663, opened 2018-02-25
Fix error generator

docelic:
Update src/handlers/error.cr.ecr and src/amber/pipes/error.cr.
  
The files /src/amber/cli/templates/error/src/handlers/error.cr.ecr
and src/amber/pipes/error.cr are pretty similar but have desynced
a little bit.
  
This commit fixes the textual descriptions and adds a missing
begin to make the files as similar as possible.
  
The next step will be to change location from /handlers/ in the
user application to /pipes/.


src/amber/pipes/error.cr
```diff
@@ -1,24 +1,26 @@
...
-        context.response.print(error.forbidden)
-      rescue ex : Amber::Exceptions::RouteNotFound
-        context.response.status_code = 404
-        error = Amber::Controller::Error.new(context, ex)
-        context.response.print(error.not_found)
-      rescue ex : Exception
-        context.response.status_code = 500
-        error = Amber::Controller::Error.new(context, ex)
-        context.response.print(error.internal_server_error)
+        begin
```
robacarp: This added begin is what's failing [ameba](https://github.com/amberframework/amber/blob/master/.travis.yml#L17)

docelic: @robacarp oh interesting, one of the error files had it and the other not. I assumed having it was preferrable. Will remove, thanks.

eliasjpr:
Closing in favor of  #685

PR #665, opened 2018-02-25
[CORS] Adds W3C Cors Specification Support

eliasjpr:
> The current CORS implementation is lacking some functionality and is failing to meet W3C specifications. The changes presented here provides facilities for dealing with CORS requests and responses while staying complaint to W3C specs.

ISSUE https://github.com/amberframework/amber/issues/240

## How It Works

This handler is compliant with the W3C CORS specification. As per this specification, It doesn’t put any CORS response headers in a connection that holds an invalid CORS request. To know what “invalid” CORS request means, have a look at the “Validity of CORS requests” section below.

When some options that are not mandatory and have no default value (such :max_age) at intialization, the relative header will often not be sent at all. This is compliant with the specification and at the same time it reduces the size of the response, even if just by a handful of bytes.

The following is a list of all the CORS response headers supported by the handler:

- Access-Control-Allow-Origin
- Access-Control-Allow-Methods
- Access-Control-Allow-Headers
- Access-Control-Allow-Credentials
- Access-Control-Expose-Headers
- Access-Control-Max-Age

## Validity of CORS requests

“Invalid CORS request” can mean that a request doesn’t have an Origin header (so it’s not a CORS request at all) or that it’s a CORS request but:

  - The Origin request header doesn’t match any of the allowed origins
  - The request is a preflight request but it requests to use a method
    or some headers that are not allowed (via the Access-Control-Request-Method
    and Access-Control-Request-Headers headers)

## Responding to preflight requests

When the request is a preflight request and is a valid one (valid origin, valid request method, and valid request headers), The CORS handler directly sends a response to that request instead of just adding headers to the connection. To do this, the handler halts the connection and sends a response.

## Requesting for inputs on the following

- [ ] Cleanup **CORS::OriginType** to support **String | Array(String) | Regex**
- [x] POssibly implement Forbidden Response for failed CORS validity, this is a little subjective since the W3C specs does not state to set a forbidden response for invalid requests, while other frameworks do implements a 403 Forbidden Response


src/amber/pipes/cors.cr
```diff
@@ -1,81 +1,133 @@
...
-        #   end
-        # end
+      private def put_expose_header(response)
+        response.headers[Headers::ALLOW_EXPOSE] = @expose_headers.as(Array).join(",") if @expose_headers
+      end
 
-        if allow_credentials
-          context.response.headers["Access-Control-Allow-credentials"] = "true"
-        end
+      private def put_response_headers(response)
```
robacarp: these `put_` method names are inconsistent with `set_preflight_headers` below, one pattern or the other should be consistent throughout the class

src/amber/pipes/cors.cr
```diff
@@ -1,81 +1,133 @@
...
+      def valid_method?(request, methods)
+        methods.includes? request.headers[Headers::REQUEST_METHOD]?
+      end
+
+      def valid_headers?(request, headers)
+        !(headers & request.headers[Headers::REQUEST_HEADERS].split(',')).empty?
+      end
+    end
+
+    struct Origin
```
robacarp: Above `OriginType` is defined, but it's not used at all in this struct. Should it be?

spec/amber/pipes/cors_spec.cr
```diff
@@ -1,56 +1,126 @@
...
+      context.response.headers[Amber::Pipe::Headers::ALLOW_CREDENTIALS].should eq "true"
+    end
+
+    context "when preflight request" do
+      it "process valid preflight request" do
+        origins = CORS::OriginType.new
+        origins << "example.com"
+        context = cors_context(
+          "OPTIONS",
+          "Origin": "example.com",
```
robacarp: should this and line 92 above point to a shared variable?

spec/amber/pipes/cors_spec.cr
```diff
@@ -1,56 +1,126 @@
...
+          "Access-Control-Request-Method": "PUT",
+          "Access-Control-Request-Headers": "Accept"
```
robacarp: this could use the header constants

spec/amber/pipes/cors_spec.cr
```diff
@@ -1,56 +1,126 @@
...
+  request = HTTP::Request.new(method, "/", headers)
+  create_context(request)
+end
+
+def assert_cors_success(context)
+  origin_header = context.response.headers["Access-Control-Allow-Origin"]?
+  origin_header.should_not be_nil
+end
+
+def assert_cors_not_success(context)
```
robacarp: this english of this method name is a little unexpected, how about `refute_cors_success`?

src/amber/pipes/cors.cr
```diff
@@ -1,81 +1,133 @@
...
+        else
+          return forbidden(context)
+        end
       end
 
-      def call(context : HTTP::Server::Context)
-        context.response.headers["Access-Control-Allow-Origin"] = allow_origin
+      def forbidden(context)
+        context.response.headers["Content-Type"] = "text/plain"
+        context.response.respond_with_error FORBIDDEN, 403
```
robacarp: How will this interact with the Error pipe?

eliasjpr: This writes to the response without raising an error. Should we raise an error here?

spec/amber/pipes/cors_spec.cr
eliasjpr: How would you do this in crystal? The arguments are passed as a NamedTuple. I am not sure how to accomplish what you are asking.

src/amber/pipes/cors.cr
eliasjpr: This was updated

src/amber/pipes/cors.cr
robacarp: I think so, that way it can be caught by the errors pipe and rendered nicely

eliasjpr: So we do not need to raise error here just respond with 403 according to the CORS specs since the Pref-light request is done behind the scenes by the browser

PR #668, opened 2018-03-03
[Params] Patches params #to_h

eliasjpr:
Fixes params #to_h method to return a flat hash of params.

### DIFF
--- a/spec/amber/router/params_spec.cr
+++ b/spec/amber/router/params_spec.cr
@@ -6,7 +6,7 @@ module Amber::Router
     headers["Content-Type"] = "multipart/form-data; boundary=fhhRFLCazlkA0dX; charset=UTF-8"
     multipart_content = ::File.read(::File.expand_path("spec/support/sample/multipart.txt"))
     multipart_body = multipart_content.gsub("\n", "\r\n")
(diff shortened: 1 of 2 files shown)


src/amber/router/params.cr
```diff
@@ -61,13 +61,11 @@ module Amber::Router
...
     def to_h
-      {
-        :multipart => multipart.to_h,
-        :json      => json.to_h,
-        :route     => route.to_h,
-        :form      => form.to_h,
-        :query     => query.to_h,
-      }
+      query.to_h
+        .merge(form.to_h)
```
elorest: This should probably be memoed?

eliasjpr: I think is better not to memoize in case a value is changed or added

spec/amber/router/params_spec.cr
```diff
@@ -24,6 +24,11 @@ module Amber::Router
           params["invalid"]
         end
       end
+
+      it "parses int value" do
+        params["status"].to_i32.should eq 1234
+        params["status"].to_i32?.should eq 1234
```
robacarp: 👍

PR #669, opened 2018-03-03
Simplify README

faustinoaq:
### Description of the Change

Current README looks very complex, this PR tries to add some enhancements and simplifies it a lot:

github com_amberframework_amber_blob_fa_screenshot_readme md

### Alternate Designs

Keep current README complex and confusing

### Benefits

1. Move some sections like Amber Philosophy to a new file
2. Use Techempower Framework Benchmark comparing related frameworks.
3. Use documentation links to reduce README content

### Possible Drawbacks

No, I like this new one :smile:


faustinoaq:
Also see `258 lines (184 sloc)  10.1 KB` (current) vs `102 lines (66 sloc)  4.9 KB` (new)

More than 100 lines less on README :sweat_smile:

faustinoaq:
Updated screenshot for this new README:

_(add new link)_

github com_amberframework_amber_blob_fa_readme-screenshot_readme md 1

README.md
```diff
@@ -229,29 +75,29 @@ Contributing to Amber can be a rewarding way to learn, teach, and build experien
...
+[See more Amber contributors](https://github.com/amberframework/amber/graphs/contributors)
 
 ## License
 
 This project is licensed under the MIT License - see the [LICENSE](LICENSE) file for details
 
 ## Acknowledgments
 
-* Inspired by Kemal, Rails, Phoenix, Hanami
+* Inspired by [Kemal](https://kemalcr.com/), [Rails](https://rubyonrails.org/) [Phoenix](https://phoenixframework.org/) and [Hanami](https://hanamirb.org/)
```
marksiemers: Keep all the commas:
* Inspired by [Kemal](https://kemalcr.com/), [Rails](https://rubyonrails.org/), [Phoenix](https://phoenixframework.org/), and [Hanami](https://hanamirb.org/)

marksiemers:
I think some sections could be combined. When there is a heading, then just one-line - it could be combined.

For example:
"Installation" and  "Usage" becomes one "Installation & Usage" heading

"Community", "Philosophy", and "Code of Conduct" could potentially be combined to a "Community" heading.

WDYT?

faustinoaq:
@marksiemers good idea, let me change it :+1:

README.md
```diff
@@ -25,199 +25,68 @@ The purpose of Amber is not to create yet another framework, but to take advanta
...
-  -t, --template  # Preconfigure for selected template engine. Options: slang | ecr
-                  (default: slang)
-  -h, --help      # Describe available commands and usages
-  -v, --version   # Prints Amber version
-
-Example:
-  amber new ~/Code/Projects/weblog
-  This generates a skeletal Amber installation in ~/Code/Projects/weblog.
+```
+git clone https://github.com/amberframework/amber.git
```
elorest: This is wrong as it will clone master instead of stable which isn't what what we should be recommending. I've changed this in the past. It should be.

```
curl -OL https://github.com/amberframework/amber/archive/stable.tar.gz
tar -xzf stable.tar.gz
cd amber-stable
make
```

faustinoaq: @elorest Ok, I will change it, but why not:

```
git clone https://github.com/amberframework/amber.git
cd amber
git checkout stable
sudo make install
```

It looks cleaner for me, WDYT?

faustinoaq:
> Overall I'm fine with it. I worried a bit that about posting the one Techempower benchmark that makes us look good though. I don't know how to replicate their results as usually in my tests we're substantially faster across the board.

I think we should try it, it will show to developers where we need to improve and how we're getting faster over time. Also comparing Ruby and Elixir Framework makes crystal looks better IMO

> It seems with these benchmarks though most of them show kemal beating us.

Kemal is a micro-framework, I think is ok to be faster in some tests.

> Images added should be https://github.com/amberframework/site-assets instead of part of amber which makes the repo larger.

I don't have access to that repo, Can you upload benchmark image on assets repo?

marksiemers:
@faustinoaq - I think @elorest 's point was about having a screenshot of the benchmark where amber is faster than Kemal - like cherry-picking the best benchmark.

faustinoaq:
> I think @elorest 's point was about having a screenshot of the benchmark where amber is faster than Kemal - like cherry-picking the best benchmark.

@marksiemers I'm not choosing the best benchmark for amber, I just using the [default benchmark view thet TFB provides](https://www.techempower.com/benchmarks/). In fact if some developer search "web framework benchmark" on google, he will see TFB Fortune tests first.

Why did I choose Kemal, Hanami, Phoenix and Rails to compare Amber?, because [we're "inspired" on these frameworks](https://github.com/amberframework/amber#acknowledgments)

I think TFB is a very nice benchmark test, I like how they have been comparing web framework performance over years.

README.md
elorest: Either way. The git way is a bit more overhead and forces people to have git. Most of them will though.

elorest:
@faustinoaq Ok good to know. I guess it's fine to post it though. I added you to the assets repo.

README.md
faustinoaq: I prefer the git way :sweat_smile: [Is similar how lucky does it](https://luckyframework.org/guides/installing/#install-lucky)

README.md
```diff
@@ -25,199 +25,69 @@ The purpose of Amber is not to create yet another framework, but to take advanta
...
-You should now be able to run `amber` in the command line.
-
-#### Mac OS X
-
-Best way to get `amber` on Mac OS X is via Homebrew:
-
-```shellsession
-$ brew install amberframework/amber/amber
+brew tap amberframework/amber
+brew install amber
```
marksiemers: @elorest - is the `tap` still necessary or will a simple `brew install amber` work now?

marksiemers:
The benchmark png that exists in this PR should be removed, correct? And moved to the assets repo, if it isn't already there.

README.md
robacarp: `brew install amber` works

faustinoaq: Done! :+1:

PR #670, opened 2018-03-04
Use shards command with production flag

faustinoaq:
### Description of the Change

Add `--production` flag and avoid using `@crystal deps` , use `@shards install` instead

### Alternate Designs

No

### Benefits

Avoid running ameba when running `make` to build amber executable

### Possible Drawbacks

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ all: build
 build: lib $(AMBER)
 
 lib:
-	@crystal deps
+	@shards install --production
 
 $(AMBER): $(AMBER_SOURCES) | $(OUT_DIR)
 	@echo "Building amber in $@"


faustinoaq:
@amberframework/core-team Some suggestions? :sweat_smile:

marksiemers:
This seems fine to me, but I don't know enough about the differences between the two commands to approve.

robacarp:
I'm in the same boat as @marksiemers.

faustinoaq:
> This seems fine to me, but I don't know enough about the differences between the two commands to approve.

@marksiemers @robacarp `crystal deps` has been removed here https://github.com/crystal-lang/crystal/pull/5544

So, we should use `shards install` instead

Oh, and `--production` flag is to avoid installing ameba, because isn't needed here :sweat_smile:

PR #673, opened 2018-03-05
Use Inflector to pluralize model names

jasonl99:
### Description of the Change
This uses an [inflector](https://github.com/phoffer/inflector.cr) to pluralize model variable and display names in scaffolding (views, controllers. models and migrations).  Currently, amber simply tacks on an "s" (for example, `Entry` becomes the table name `entrys` rather than `entries`).

### Alternate Designs

No other alternate designs were considered

### Benefits

Proper pluralization enhances code and site readability.

### Possible Drawbacks
This change relies on an external shard.   There is a risk that it will stop being maintained.


robacarp:
This is a step in a good direction! Thanks @jasonl99 

I think there is a blocker here though which is that granite doesn't currently support correctly inflected names for associations, and it also doesn't have override support for table names on associations.

faustinoaq:
I know amber doesn't have support for Portuguese or Spanish yet, but how it would behave with other languages in the future? I'm just curious :sweat_smile:  @amberframework/contributors

eliasjpr:
I like this, the only comment I have is the the shards is not being actively maintained. @amberframework/core-team should we consider forking?

faustinoaq:
> I know amber doesn't have support for Portuguese or Spanish yet, but how it would behave with other languages in the future?

Wow, https://github.com/watzon/cadmium by @watzon is a new shard that supports many awesome things, including inflectors (1)

1. https://github.com/watzon/cadmium#inflectors

and maybe multiple languages :tada: 

> The aggressive tokenizer currently has localization available for English (:en / nil), Spanish (:es), Persian (:fa), French (:fr), Indonesian (:id), Dutch (:nl), Norwegian (:no), Polish (:pl), Portuguese (:pt), Russian (:ru), and Swedish (:sv). If no language is included it will

src/amber/cli/templates/model.cr
```diff
@@ -21,7 +22,7 @@ module Amber::CLI
     end
 
     def table_name
-      @table_name ||= "#{@name}s"
+      @table_name ||= "#{Inflector.pluralize(@name)}"
```
marksiemers: I know this was already merged, but I think a refactor would be helpful. The code `Inflector.pluralize(@name)` shows up in lots of places. Why not have this in the template:
```
def plural_name
  @plural_name ||= Inflector.pluralize(@name)
end
```
Then `plural_name` can be used everywhere, and if amber ever swaps this inflector for another one (like what @faustinoaq mentioned) then it just has to change in this one place.

Also, the inflector being called that many times might incur some unnecessary overhead.

faustinoaq: I agree :+1:

elorest:
Yeah i agree with you.

PR #679, opened 2018-03-07
[Shard] Change citrine-i18n version to 0.1.0

faustinoaq:
### Description of the Change

Change citrine-i18n to version 0.1.0

### Alternate Designs

No

### Benefits

Use citrine-i18n version instead of master branch and avoid dependency issues

### Possible Drawbacks

No


faustinoaq:
> when test passes

@eliasjpr Done! :+1:

PR #682, opened 2018-03-07
[Shard] Removes Spinner Shard

eliasjpr:
This got back in somehow :(


elorest:
Isn't this shard being used in deploy?

PR #689, opened 2018-03-09
[CLI] Add a command to open db connection.

wontruefree:
This adds a CLI command `connection` for quick commandline access.

### DIFF
--- a/src/amber/cli/commands/database.cr
+++ b/src/amber/cli/commands/database.cr
@@ -43,8 +43,7 @@ module Amber::CLI
         CLI.toggle_colors(options.no_color?)
 
         if args.commands.empty?
-          Process.exec(command_line_tool, {database_url}) if database_url
-          exit! error: false
+          connect_to_database
         end
 
         args.commands.each do |command|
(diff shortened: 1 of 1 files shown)


eliasjpr:
I think this is a great addition. But I am curious to how this will work given that I don't see specs. 

Will this allow execute queries? How will this work with Sqlite and MySQL?

src/amber/cli/commands/database.cr
```diff
@@ -41,6 +41,11 @@ module Amber::CLI
 
       def run
         CLI.toggle_colors(options.no_color?)
+
+        if args.commands.empty?
```
eliasjpr: I don't know how I feel about defaulting to connect to the db when no arguments are present.

faustinoaq:
Fixes #708

src/amber/cli/commands/database.cr
robacarp: I'm assuming that this default would be invoked with simply `amber db`? If so, I see no problem with that.

Issue #690, opened 2018-03-10
[Controller] Set custom status_code in respond_with


rdelandesen:
### Description

By developing an API (for example), I am used to customizing the status code of the HTTP response.

**E.g.:**
- `201` for `create` action
- `400` for a call with missing or invalid parameters
- `403` if unauthorized etc.

It seems that it is not possible to customize the status code using `Amber::Controller::Helpers::Responders::Content # respond_with`

### Solution

Add a `status_code` attribute:

```
# /lib/amber/src/amber/controller/helpers/responders.cr

@status_code : Int32 = 200

def status_code(@status_code : Int32)
end

# respond_with
set_response(body: content.body, status_code: content.status_code, content_type: content.type)
```

Call:

```
respond_with do
  status_code 201
  json pet.to_json
end
```

What do you think about it?
(BTW, I take this opportunity to say congratulations for your work!)

elorest:
@rdelandesen 

What do you think of this syntax?

```cr
respond_with(201) do
  json pet.to_json
end
```

It might also be good to make this method public.

https://github.com/amberframework/amber/blob/master/src/amber/controller/helpers/responders.cr#L89

Should be a pretty easy fix. Do you want to submit a PR for that or should I?

PR #691, opened 2018-03-10
Set respond_with status_code

rdelandesen:
https://github.com/amberframework/amber/issues/690

Set custom respond_with(status_code)

```
respond_with(201) do
  json pet.to_json
end
```

### DIFF
--- a/src/amber/controller/helpers/responders.cr
+++ b/src/amber/controller/helpers/responders.cr
@@ -60,6 +60,12 @@ module Amber::Controller::Helpers
       end
     end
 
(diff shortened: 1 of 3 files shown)


src/amber/controller/helpers/responders.cr
```diff
@@ -60,6 +60,12 @@ module Amber::Controller::Helpers
       end
     end
 
+    def set_response(body, status_code = 200, content_type = Content::TYPE[:html])
```
eliasjpr: Did this method definition changed in any way? It seem like it did not but yet the method was moved

PR #700, opened 2018-03-17
[Router] Adds OPTIONS and HEAD routes by default

eliasjpr:
Issue https://github.com/amberframework/amber/issues/240

OPTIONS and HEAD routes should be automatically defined for routes
currently developers have to define the OPTIONS route via the option
route method.

- Removes the add_head method from router
- Adds OPTIONS and HEAD routes ass needed via DSL
- Removes the need to create OPTIONS and HEAD routes

With these changes developers should no longer need to specify the
OPTIONS and HEAD routes and will now be defined by default.

To learn more about HTTP Head and Options Specs Read https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

### DIFF
--- a/src/amber/dsl/router.cr
+++ b/src/amber/dsl/router.cr
@@ -28,6 +28,13 @@ module Amber::DSL
     {% for verb in RESOURCES %}
       macro {{verb.id}}(*args)
         route {{verb}}, \{{*args}}
+        {% if verb == :get %}
+          route :head, \{{*args}}
+        {% end %}
+        route {{verb}}, \{{*args}}
+        {% if ![:trace, :connect, :options, :head].includes? verb %}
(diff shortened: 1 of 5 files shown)


faustinoaq:
Fixes #240

src/amber/dsl/router.cr
```diff
@@ -28,6 +28,13 @@ module Amber::DSL
     {% for verb in RESOURCES %}
       macro {{verb.id}}(*args)
         route {{verb}}, \{{*args}}
+        {% if verb == :get %}
+          route :head, \{{*args}}
```
faustinoaq: I think we should align this line

BTW, macros aren't checked by crystal tool format, they keep untouched

PR #703, opened 2018-03-18
fix support for fetch_all in form params

jreinert:
### Description of the Change

Fixes a bug that prevented reading multi-values from form params

### Benefits

Multi valued fields for form params work


faustinoaq:
Fxes #704

Issue #671, opened 2018-03-04
[CLI] Generate pages using amber


faustinoaq:
### Description

Currently I'm trying to generate a page in `src/view` rendering it on a controller and adding a new route using `amber` command, but is not available yet :sweat: 

### Steps to Reproduce

1. Generate an amber project
2. Try `amber g page About`

**Expected behavior:** Add a view with a controller and a router entry `/about`

**Actual behavior:** I can't generate a page

**Reproduces how often:** 100%

### Versions

crystal 0.24.1

amber 0.6.7

### Additional Information

BTW, we should try to output a more friendly error:

```
 amber g view About
Template not found (Exception)
Failed to raise an exception: END_OF_STACK
[0x558cc5f880e6] *CallStack::print_backtrace:Int32 +118
[0x558cc5f4f6f6] __crystal_raise +86
[0x558cc5f4fe6e] ???
[0x558cc5fc3e09] *File#initialize<String, String, Int32, Nil, Nil>:(Int32 | Nil) +201
[0x558cc5fc3d21] *File::new<String, String, Int32, Nil, Nil>:File +209
[0x558cc5f7f87c] *CallStack::read_dwarf_sections:(Array(Tuple(UInt64, UInt64, String)) | Nil) +108
[0x558cc5f7f5bd] *CallStack::decode_line_number<UInt64>:Tuple(String | Nil, Int32, Int32) +45
[0x558cc5f7ed68] *CallStack#decode_backtrace:Array(String) +296
[0x558cc5f7ec22] *CallStack#printable_backtrace:Array(String) +50
[0x558cc5fc2e3d] *Exception+ +77
[0x558cc5fc24a5] *Exception+ +117
[0x558cc61a41fd] *Crystal::main<Int32, Pointer(Pointer(UInt8))>:Int32 +157
[0x558cc5f5a536] main +6
[0x7f73542edf4a] __libc_start_main +234
[0x558cc5f466ea] _start +42
[0x0] ???
```

:point_up_2: This is very verbose

This feature request is similar to `lucky gen.page About`

@amberframework/contributors WDYT?

elorest:
I don't think we should add a page generator as the functionality already exists.

`amber g controller About about:get` with existing generators will add an About controller with an action, view and route for about.

eliasjpr:
I completely agree.

I can see the benefit of having a view generator to generate a 1 file view but pages is redundant and ti seems to belong to lucky nomenclature

eliasjpr:
[...]

PR #710, opened 2018-03-23
[CLI] Add scaffold generator to recipe generator feature

damianham:
### Description of the Change
This change adds the scaffold generator for recipes and further extends
https://github.com/amberframework/amber/pull/705

Please see the conversation in PR 705 for details of this change.


faustinoaq:
Seems ameba detected some issues in this PR:

@damianham Check this message:

```
TrailingWhitespace: Trailing whitespace detected
```

src/amber/cli/recipes/recipe_fetcher.cr
```diff
@@ -0,0 +1,70 @@
...
+      # otherwise fetch from the github repository
+      return fetch_url template_path
+    end
+
+    def fetch_url(template_path)
+      # download the recipe zip file from the github repository
+      HTTP::Client.get("https://raw.githubusercontent.com/AmberRecipes/recipes/master/#{@name}.zip") do |response|
+        if response.status_code != 200
+          # raise an exception if the recipe zip was not found
+          raise "Could not find that recipe"
```
faustinoaq: Should we print a full exception here?

See #686 and #671 (Additional Information)

src/amber/cli/recipes/recipe_fetcher.cr
```diff
@@ -0,0 +1,70 @@
...
+      end
+
+      # return the path of the template directory
+      if Dir.exists?("#{template_path}/#{@kind}")
+        return "#{template_path}/#{@kind}"
+      end
+
+      # raise an exception if the recipe does not contain the generator for
+      # the element being generated
+      raise "Cannot generate a #{@kind} from that recipe"
```
faustinoaq: Same here :sweat_smile:

src/amber/cli/recipes/recipe.cr
```diff
@@ -0,0 +1,151 @@
...
+    rescue
+      p "Cannot generate a #{template_type} from #{recipe}"
+      false
+    end
+
+    def initialize(name : String, directory : String, recipe : String, fields = [] of String)
+      if name.match(/\A[a-zA-Z]/)
+        @name = name.underscore
+      else
+        raise "Name is not valid."
```
faustinoaq: ...and here :sweat_smile:

damianham:
All specs are now passing and exceptions removed from CLI

src/amber/cli/commands.cr
```diff
@@ -47,6 +48,7 @@ module Amber::CLI
       string ["-t", "--template"], desc: "# Preconfigure for selected template engine. Options: slang | ecr", default: "slang"
       string ["-d", "--database"], desc: "# Preconfigure for selected database. Options: pg | mysql | sqlite", default: "pg"
       string ["-m", "--model"], desc: "# Preconfigure for selected model. Options: granite | crecto", default: "granite"
+      string ["-r", "--recipe"], desc: "# Use a named recipe.  See documentation.", default: nil
```
robacarp: `see documentation` isn't a helpful message. Can this be something more like `see documentation at <amber web address>` or `run --recipe-help for more info` or something more prescriptive?

src/amber/cli/commands/generate.cr
```diff
@@ -23,11 +23,22 @@ module Amber::CLI
           template = Template.new("error", ".")
         else
           ensure_name_argument!
-          template = Template.new(args.name, ".", args.fields)
+          # if the app was generated from a recipe and the component can be
+          # generated for the recipe then use the recipe generator
+          if recipe && recipe != "none" && Amber::Recipes::Recipe.can_generate?(args.type, recipe)
```
robacarp: At first glance, having a recipe called "none" seems like it could be represented more clearly with just `nil`. Doing so would clean up the logic here a little as well.

src/amber/cli/commands/generate.cr
```diff
@@ -23,11 +23,22 @@ module Amber::CLI
...
+            template = Amber::Recipes::Recipe.new(args.name, ".", recipe, args.fields)
+          else
+            # otherwise use the standard template generator
```
robacarp: The comments in this file probably aren't necessary, the code is clear enough.

src/amber/cli/commands/new.cr
```diff
@@ -10,6 +10,7 @@ module Amber::CLI
         string "-d", desc: "database", any_of: %w(pg mysql sqlite), default: "pg"
         string "-t", desc: "template language", any_of: %w(slang ecr), default: "slang"
         string "-m", desc: "model type", any_of: %w(granite crecto), default: "granite"
+        string "-r", desc: "recipe", default: "none"
```
robacarp: I think if this is just `string "-r", desc: "recipe"` (with no default), the default will be nil and the logic in several places can be simplified.

src/amber/cli/recipes/file_entries.cr
```diff
@@ -0,0 +1,69 @@
+require "teeplate"
+
+module Amber::Recipes
+  module FileEntries
+    def each_file(abs, rel, &block : String, String ->)
```
robacarp: The names `abs` and `rel` throughout this file are opaque. I'm _guessing_ that they are `absolute_path` and `relative_path`? Similarly `tpl` `d`, etc.

Specifying the full name of a variable is a little more verbose, but it helps the long term maintainability of a codebase.

src/amber/cli/recipes/file_entries.cr
```diff
@@ -0,0 +1,69 @@
...
+        block.call abs, rel
+      end
+    end
+
+    def pack_liquid(files, abs, rel)
+      tpl = Liquid::Template.parse File.new(abs)
+      io = tpl.render @ctx.as(Liquid::Context)
+      files << ::Teeplate::StringData.new(rel, io.to_s, File.stat(abs).perm)
+    rescue ex
+      p "failed to process #{abs} - #{ex.message}"
```
robacarp: I'm guessing you want `puts` here instead of `p`? https://carc.in/#/r/3s3s

src/amber/cli/recipes/controller.cr
```diff
@@ -0,0 +1,60 @@
...
+      ROUTES
+
+      @action_names = @actions.map { |action, verb| action}
+
+      @template = RecipeFetcher.new("controller", @recipe).fetch
+
+      add_views
+    end
+
+    # setup the Liquid context
```
robacarp: This comment is redundant

src/amber/cli/recipes/file_entries.cr
```diff
@@ -0,0 +1,69 @@
...
+      File.open(abs) { |f| IO.copy(f, io) }
+
+      files << ::Teeplate::Base64Data.new(rel, io.size.to_u64, Base64.encode(io), File.stat(abs).perm)
+    end
+
+    def collect_files(files)
+      if ! @template.nil?
+
+        @ctx = Liquid::Context.new
+        set_context @ctx
```
robacarp: It seems like the `set_context` method is expected to be defined in the including class. If so, what sort of error message will be rendered when a future developer adds `include FileEntries` without that method? Can it be made more helpful?

src/amber/cli/recipes/file_entries.cr
```diff
@@ -0,0 +1,69 @@
...
+            pack_liquid files, abs, $1
+          else
+            # pack the file without processing
+            pack_blob files, abs, rel
+          end
+        end
+      end
+    end
+
+    # Returns collected file entries.
```
robacarp: Redundant comment.

src/amber/cli/recipes/filters.cr
```diff
@@ -0,0 +1,54 @@
+# add a pluralize filter to liquid
+# this should be a temporary measure until pluralize is available as a filter
+# in the liquid shard
```
robacarp: This comment isn't correct as this file is adding pluralize as well as underscore. It's probably redundant anyway.

src/amber/cli/recipes/filters.cr
```diff
@@ -0,0 +1,54 @@
...
+
+require "inflector"
+
+module Liquid::Filters
+
+  FilterRegister.register "pluralize", Pluralize
```
robacarp: Given that this code is written as a shim until the liquid project gets this filter, will this code break when the liquid project is updated?

src/amber/cli/recipes/recipe.cr
```diff
@@ -0,0 +1,148 @@
...
+      unless Dir.exists?(@directory)
+        Dir.mkdir_p(@directory)
+      end
+
+      @recipe = recipe
+
+      @fields = fields
+    end
+
+    def cleanup
```
robacarp: Your comment states that this method is specific to recipes fetched from github, but the method name doesn't imply that. Can it be `clean_github_cache` or something similarly transparent?

src/amber/cli/recipes/recipe.cr
```diff
@@ -0,0 +1,148 @@
...
+      # if we used a template from github then remove the temp folder
+      # this should probably be called as a command so templates are cached
+      FileUtils.rm_rf(AMBER_RECIPE_FOLDER)
+    end
+
+    def generate(template : String, options = nil)
+      case template
+      when "app"
+        if options
+          puts "Rendering App #{name} in #{directory} from #{options.r}"
```
robacarp: Should these puts statements be funneled into the logger instead? 

cc @eliasjpr

spec/amber/cli/recipes/recipe_fetcher_spec.cr
```diff
@@ -0,0 +1,34 @@
...
+      Dir.mkdir_p("./mydefault/app")
+      Dir.mkdir_p("./mydefault/controller")
+      Dir.mkdir_p("./mydefault/scaffold")
+    end
+
+    Spec.after_each do
+      FileUtils.rm_rf("./mydefault")
+    end
+
+    # a consequence of app initialization is setting the template source folder
```
robacarp: This comment doesn't make sense, is it obsolete?

src/amber/cli/recipes/recipe_fetcher.cr
```diff
@@ -0,0 +1,72 @@
+require "http/client"
+require "zip"
+
+module Amber::Recipes
+  class RecipeFetcher
+    getter kind : String
```
robacarp: `kind` is opaque, but it might be the right name for the job. Adding a comment with an example to the definition here could help future maintainers.

src/amber/cli/recipes/recipe_fetcher.cr
```diff
@@ -0,0 +1,72 @@
...
+    getter name : String
+    getter directory : String
+
+    def initialize(@kind : String, @name : String)
+      @directory = "#{Dir.current}/#{@name}/#{@kind}"
+    end
+
+    def fetch
+      # if the recipe exists in the local recipes folder then use that
```
robacarp: the comments in this module are redundant with the code

src/amber/cli/recipes/recipe_fetcher.cr
```diff
@@ -0,0 +1,72 @@
...
+        return "#{template_path}/#{@kind}"
+      end
+
+      # otherwise fetch from the github repository
+      return fetch_url template_path
+    end
+
+    def fetch_url(template_path)
+      # download the recipe zip file from the github repository
+      HTTP::Client.get("https://raw.githubusercontent.com/AmberRecipes/recipes/master/#{@name}.zip") do |response|
```
robacarp: This base url should be a constant somewhere

src/amber/cli/recipes/scaffold/controller.cr
```diff
@@ -0,0 +1,83 @@
...
+      when "int32", "int64", "integer"
+        "1"
+      when "float32", "float64", "float"
+        "1.00"
+      when "bool", "boolean"
+        "true"
+      when "time", "timestamp"
+        Time.now.to_s
+      when "ref", "reference", "references"
+        rand(100).to_s
```
robacarp: I don't understand this

src/amber/cli/recipes/recipe.cr
eliasjpr: It seems that it is being funneled thru the logger see line https://github.com/amberframework/amber/pull/710/files#diff-7129a7c5dcf68d7430b572158d3977c3R88

src/amber/cli/recipes/recipe_fetcher.cr
eliasjpr: (Non Blocking) How many `kind` of recipes there are? If the list is short and known maybe converting it to an Enum would improve the intention.

src/amber/cli/recipes/recipe_fetcher.cr
eliasjpr: (Non Blocking) I will opt to make this URL configurable via the `.amber.yml` something like recipe sources. In case people would like to have their own private repo of recipes.

src/amber/cli/recipes/recipe.cr
```diff
@@ -0,0 +1,148 @@
...
+      else
+        CLI.logger.error "Template not found", "Generate", :light_red
+      end
+    end
+
+    def model
+      CLI.config.model
+    end
+
+    def puts(msg)
```
robacarp: I missed this on the first time through, but shadowing puts makes me a little uncomfortable. I've done it in the past on some of my projects and regretted it. How about just `log`?

src/amber/cli/recipes/file_entries.cr
damianham: it doesn't compile

in src/amber/cli/recipes/file_entries.cr:43: undefined method 'set_context'

        set_context @ctx

src/amber/cli/recipes/filters.cr
damianham: no it will simply overwrite the existing filter

src/amber/cli/recipes/scaffold/controller.cr
damianham: src/amber/cli/templates/field.cr recognizes "ref", "reference" and "references" as a reference field type e.g. user_id:ref and default values are used when creating model instances in controller test specs

damianham:
@faustinoaq @robacarp @elorest cleanups and requested changes committed, please take another look and point out any more required changes thanks.

damianham:
@eliasjpr I made the URL for recipe sources configurable via the .amber.yml.  As per @faustinoaq previous question I think recipes should be a repository within the amberframework (so you guys can manage the PRs :-) and also there should probably be a default.zip in the root of that repository which contains the default templates.

src/amber/cli/recipes/scaffold/controller.cr
robacarp: Sorry, I guess what I don’t understand is the Rand value for it

damianham: Yes of course - if there is a database constraint to ensure the referenced record exists then creating the record in the spec will crap out - but it is the amber test suite - so there won't be a constraint - so all hunky dory

robacarp:
I'm not 100% sure what the relationship between #716 and this pull is. There seems to be a lot of shared code?

Here are my thoughts:
- I'd still like to see a more complete test coverage of these awesome features.
- What happens when I generate an api-only application, but then later want to scaffold out a traditional full-web controller? (not blocking, this can be solved later)
- While reviewing your doc pull, I found some concerns about local caching:
  - Dropping an arbitrary `.amber_recipe_cache` folder in whatever pwd a new/generate command is executed in feels dangerous to me. I expect that I'll be finding these folders everywhere on my system, and it also seems like it has the potential to produce phantom errors. Could this live in some ...more predictable ...place? `/usr/local/share` or something?
  - Caching should be scoped by repository as well, to avoid naming conflicts.
  - When I generate a app with a recipe, wait 6 months, and then generate another app with that recipe, will the new app use the 6 month old recipe cache? Update it? Download another? What if I go back to the first and generate a new scaffold? Does it update the recipe, use an updated recipe, or something else?

Most importantly, if you're going to gift the repo to the amber org, now is the time. Doing it later will only cause a lot of confusing updates to documentation and code.

damianham:
@robocarp

716 is an extension of this code and so the 716 branch is branched off this code, hence what appears to be duplication.  The merge will of course sort it out.

I couldn't think of anything worthwhile to test more than there is since in essence this PR simply changes the location of the template for the generator - that bit is tested.  All the other code is a copy of existing code.

If you generate an api-only application, but then later want to scaffold out a traditional full-web controller you would have to edit .amber.yml and remove the recipe name

Using the recipe generator with amber new myapp -r recipe will create a recipe cache folder in the current folder (the parent of myapp).  Subsequent use in the app folder will use the cache folder in the parent.  I create all my work in one workspace folder so it is likely I will have only 1 recipe cache folder.
I don't think placing the cache folder in /usr/local/share is a good idea because of visibility,   $HOME/.amber/recipe_cache might be a better idea.

"Caching should be scoped by repository as well, to avoid naming conflicts".  Good point - let's solve this later as an enhancement.

Yes the generator will use the old recipe - as yet there is no cache expiry mechanism - again something for a future enhancement.

If you create a recipes repository in amberframework it is a simple change to the URL (and the documentation).  I didn't see that repo appear yet so went ahead with the current URL.

All good points but I don't see these as blockers.

faustinoaq:
> Most importantly, if you're going to gift the repo to the amber org, now is the time. Doing it later will only cause a lot of confusing updates to documentation and code.

Yeah, please, I agree, we should move that repo :)

> If you create a recipes repository in amberframework it is a simple change to the URL (and the documentation). I didn't see that repo appear yet so went ahead with the current URL.

@damianham Done! :tada: , please see: https://github.com/amberframework/recipes

^ I already gave you access, check your email inbox :wink:

damianham:
@faustinoaq recipes repo is now amberframework/recipes and has been approved by all so I think this can be merged.

faustinoaq:
> I think this can be merged.

Oh yeah, ping @amberframework/core-team ! :smile:

src/amber/cli/recipes/recipe_fetcher.cr
```diff
@@ -0,0 +1,72 @@
...
+
+      if Dir.exists?("../#{template_path}/#{@kind}")
+        return "../#{template_path}/#{@kind}"
+      end
+
+      return fetch_url template_path
+    end
+
+    def recipe_source
+      CLI.config.recipe_source || "https://raw.githubusercontent.com/amberframework/recipes/master/"
```
drujensen: Recipes should be shards not zip files.  This is adding another packaging system to the mix with the need to handle versioning, compatibility, dependencies, etc.

robacarp: Distributing recipes as shards feels like it'll be quite a wedge. They're not versioned, have no dependencies, and really aren't even crystal code. They're just collections of templates and folders. A zip is a great way to package a recipe for download, I just don't think that this way of organizing and distributing "official" amber recipes is going to scale well.

If Amber is really going to provide a handful of blessed recipes, I see no reason why they can't all live in the same repository. But I'd like them to live _as code_, not as zips, so pull requests are functional and file history works. A CI system can create zips and publish them to s3 or something for distribution. 

I think it also makes sense for others to be able to publish a recipe as a single git repository, but keeping up with N repositories for blessed recipes feels exausting. Amber is already 21 separate repositories. Imagine trying to update for a breaking change across N repositories, that requires coordinating pull requests and branches across every repository to match the amber core.

faustinoaq: @robacarp Nice! so you mean, we should maintain just source code in the `amberframework/recipes` repo and then generate zips using CI :+1: 

@damianham WDYT?

damianham: I don't see a need for multiple repositories, the recipes can be organised in sub folders from the root 
amberframework/recipes/react
amberframework/recipes/vue
amberframework/recipes/angular
amberframework/recipes/plugins

a recipe name can include path separators e.g.  -r react/modular/streaming to use http://amberframework.org/recipes/react/modular/streaming.zip
if the recipes are published to zips on the website by CI

Adam-Stomski: hey, sorry I'm just curious why not just use git and pull the repo? It automaticly handles verification you can pull from public or private repo, and don't need the step with creating/updating/managing zip file

faustinoaq: @Adam-Stomski That will need to clone the entire repo before using it, so if the repo grows enough, we will be unable to use recipes because is so big.

I think keeping recipes source code is fine, then create zip files from source and storing them inside recipe directory is a nice option.

faustinoaq: This thread is getting big :sweat_smile: 

I opened a tracking issue for this, please leave your comments here: https://github.com/amberframework/amber/issues/750

PR #715, opened 2018-03-29
[Bug Fix] Fixed Delete link to work similar to Rails UJS

elorest:
### Description of the Change

Added a few lines of javascript which submit the delete link as a form while still allowing it to be styled like the other links.

### Alternate Designs

Just add a form straight to the page with a submit button. After looking over all the modern frameworks I decided not to do it this way.

### Benefits

Delete button uses POST instead of GET and overrides with DELETE method.

### DIFF
--- a/assets/js/amber.js
+++ b/assets/js/amber.js
@@ -226,3 +226,19 @@ export class Socket {
 module.exports = {
   Socket: Socket
 }
+
+
+/**
+ * Allows delete links to post for security and ease of use similar to Rails jquery_ujs
+ */
+$(document).ready(() => {
+  $( "a[data-method='delete']" ).click(function(){
+    var confirm_message = $(this).data("confirm") || "Are you sure?"
+    if(confirm(confirm_message)){
+      const form = $(`<form action='${this.href}' method='POST'><input type='hidden' name='_method' value='DELETE' /></form>`);
+      $(document.body).append(form);
+      form.submit();
+    }
+    return false;
+  }); 
+}); 
--- a/assets/js/amber.js
+++ b/assets/js/amber.js
@@ -231,14 +231,23 @@ module.exports = {
 /**
  * Allows delete links to post for security and ease of use similar to Rails jquery_ujs
  */
-$(document).ready(() => {
-  $( "a[data-method='delete']" ).click(function(){
-    var confirm_message = $(this).data("confirm") || "Are you sure?"
-    if(confirm(confirm_message)){
-      const form = $(`<form action='${this.href}' method='POST'><input type='hidden' name='_method' value='DELETE' /></form>`);
-      $(document.body).append(form);
-      form.submit();
-    }
-    return false;
-  }); 
-}); 
+document.addEventListener("DOMContentLoaded", function () {
+    document.querySelectorAll("a[data-method='delete']").forEach(function (element) {
+        element.addEventListener("click", function (e) {
+            e.preventDefault();
+            var message = element.getAttribute("data-confirm") || "Are you sure?";
+            if (confirm(message)) {
+                var form = document.createElement("form");
+                var input = document.createElement("input");
+                form.setAttribute("action", element.getAttribute("href"));
+                form.setAttribute("method", "POST");
+                input.setAttribute("type", "hidden");
+                input.setAttribute("value", "DELETE");
+                form.appendChild(input);
+                document.body.appendChild(form);
+                form.submit();
+            }
+            return false;
+        })
+    })
+});
--- a/src/amber/cli/templates/scaffold/view/src/views/{{name}}/index.ecr.ecr
+++ b/src/amber/cli/templates/scaffold/view/src/views/{{name}}/index.ecr.ecr
@@ -26,7 +26,7 @@
           <span>
             <%="<"%>%= link_to("read", "/<%= Inflector.pluralize(@name) %>/#{<%= @name %>.id}", class: "btn btn-success btn-xs") -%>
             <%="<"%>%= link_to("edit", "/<%= Inflector.pluralize(@name) %>/#{<%= @name %>.id}/edit", class: "btn btn-success btn-xs") -%>
-            <%="<"%>%= link_to("delete", "/<%= Inflector.pluralize(@name) %>/#{<%= @name %>.id}?_method=delete&_csrf=#{csrf_token}", class: "btn btn-danger btn-xs", onclick: "return confirm('Are you sure?');") -%>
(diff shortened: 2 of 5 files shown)


assets/js/amber.js
```diff
@@ -226,3 +226,18 @@ export class Socket {
...
   Socket: Socket
 }
+
+
+/**
+ * Allows delete links to post for security and ease of use similar to Rails jquery_ujs
+ */
+$(document).ready(() => {
+  $( "a[data-method='delete']" ).click(function(){
+    if(confirm("Are you sure you want to delete this?")){
```
eliasjpr: What if someone would like to display a different message? I thought using `button_to` would have been lesser of a change even if styling was required since the bootstrap theme we should is just that a bootstrap.

assets/js/amber.js
```diff
@@ -226,3 +226,18 @@ export class Socket {
 module.exports = {
   Socket: Socket
 }
+
+
+/**
+ * Allows delete links to post for security and ease of use similar to Rails jquery_ujs
+ */
+$(document).ready(() => {
+  $( "a[data-method='delete']" ).click(function(){
```
eliasjpr: This would mean that any element in the page will behave like a delete button. What if the intention is not use it as such. This is making a lot of assumptions and is very particular to someone specific use case.

assets/js/amber.js
```diff
@@ -226,3 +226,18 @@ export class Socket {
...
+    if(confirm("Are you sure you want to delete this?")){
+      const form = $(`<form action='${this.href}' method='POST'><input type='hidden' name='_method' value='DELETE' /></form>`);
+      $(document.body).append(form);
+      form.submit();
+    }
```
eliasjpr: Also there seems to be missing Ajax error handling to this method.

elorest: This isn't an Ajax submission. It actually posts like with rails ujs

assets/js/amber.js
elorest: It would use data-message or something like rails. I ran out of time for that last night and my son is in surgery right now so I'm afk.

elorest:
I choose not to use the bottom concept after looking at a lot of frameworks. It would be frustrating to end users to have to style one of the menu items completely differently and not have it look like a link. As much as I hate JavaScript hacks this seems to be the convention by far. Alternatively I think it would be best to go back to overriding the get method with delete.

I chose not to use jQuery UJS because it has a whole lot of code that we don't need and it was really simple to do this with just like five or six lines

assets/js/amber.js
elorest: Only elements with `data-method=delete`. Every other element will NOT behave like a delete button.

assets/js/amber.js
robacarp: Displaying a different message is a good idea, but doesn't seem like it should be required to get this feature in.

assets/js/amber.js
faustinoaq: Hey @elorest What about adding a `data.message` attribute ?

so We would be able to do this:

```html
<a href="/post/delete/1"
    data-method="delete"
    data-message="¿Estás seguro que deseas borrar esto?">Borrar</a>
```

WDYT?

faustinoaq:
> I would recommend to use the button_to since it leaves out the implementation details, and it will keep it simpler.

@eliasjpr I agree and I like it, although we would take some time to get `button_to` working :sweat_smile:

elorest:
@faustinoaq that was my intention as I commented above. I ran out of time last night and my son is currently in surgery so I'm afk.

JS isn't my favorite language so any pointers or suggestions on the code would be appreciated. 

I am convinced though that doing it this way instead of embedding an entire form and submit button into show and index scaffolding templates will be easier to for end users and more consistent with the rest of the community.

elorest:
@robacarp I agree but I also think that this is the best approach. Not necessarily this exact code but as I don't really consider myself a JS Artisan but something like this.

elorest:
I think that the message should be set from data-confirm instead of data-message since that's how rails does it.

assets/js/amber.js
```diff
@@ -226,3 +226,28 @@ export class Socket {
...
+                var input = document.createElement("input");
+                form.setAttribute("action", element.getAttribute("href"));
+                form.setAttribute("method", "POST");
+                input.setAttribute("type", "hidden");
+                input.setAttribute("value", "DELETE");
+                form.appendChild(input);
+                document.body.appendChild(form);
+                form.submit();
+            }
+            return false;
```
Sija: The `return false` works on jQuery bind function only.

assets/js/amber.js
```diff
@@ -226,3 +226,28 @@ export class Socket {
 module.exports = {
   Socket: Socket
 }
+
+
+/**
+ * Allows delete links to post for security and ease of use similar to Rails jquery_ujs
+ */
+document.addEventListener("DOMContentLoaded", function () {
+    document.querySelectorAll("a[data-method='delete']").forEach(function (element) {
```
Sija: Why not make it more generic, not only for _delete_?

faustinoaq: Hi @Sija , what do you propose? :sweat_smile:

elorest: Are there other cases where we would want a link to trigger a post?

faustinoaq: @Sija Ok, what about using this:

```js
document.querySelectorAll("a[data-method]").forEach(function (element) {
    var method = element.getAttribute("data-method");
    element.addEventListener("click", function (e) {
        e.preventDefault();
        var message = element.getAttribute("data-confirm") || "Are you sure?";
        if (confirm(message)) {
            var form = document.createElement("form");
            var input = document.createElement("input");
            form.setAttribute("action", element.getAttribute("href"));
            form.setAttribute("method", "POST");
            input.setAttribute("type", "hidden");
            input.setAttribute("value", method);
            form.appendChild(input);
            document.body.appendChild(form);
            form.submit();
        }
    });
});
```

^ I don't if amber support put or patch method overwrite, though :sweat_smile:

faustinoaq: Demo: https://jsfiddle.net/rgvsmhvy/1/ :tada:

Sija: uhm, exactly that's what I had in mind, yet it could be expanded still, up to the point of porting [`rails-ujs`](https://github.com/rails/rails/tree/master/actionview/app/assets/javascripts/rails-ujs) - would you like that? I could prepare a PR with initial implementation if needed.

Sija: > ^ I don't if amber support put or patch method overwrite, though 😅

Looks like it is: https://github.com/amberframework/amber/blob/f84fe8c4b43f9e0b307f0cd7ef1a5085e284e45a/spec/amber/router/request_spec.cr#L86-L91

faustinoaq: @Sija Oh, excellent, would be a nice PR :smile:

faustinoaq: @Sija I just opened an issue here: https://github.com/amberframework/amber/issues/738

Thank for the idea :+1:

PR #722, opened 2018-04-01
Try to fix Travis CI

faustinoaq:
### Description of the Change

Travis still fails randomly, This PR tries to investigate the root cause of this issue, feel free to collaborate :wink: 

### Alternate Designs

Not applicable

### Benefits

More reliable CI

### Possible Drawbacks

Not applicable

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -1,8 +1,3 @@
-if ENV["TEST_SUITE"] == "./spec/amber"
-  puts "WAITING: sleep 1.minute before running the amber specs to ensure all is ok..."
-  sleep 1.minute
-end
-
 # NOTE: Constants should be set before require begins.
 ENV["AMBER_ENV"] = "test"
 ENV[Amber::Support::ENCRYPT_ENV] = "mnDiAY4OyVjqg5u0wvpr0MoBkOGXBeYo7_ysjwsNzmw"
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -1,5 +1,9 @@
-# NOTE: Constants should be set before require begins.
+if ENV["TEST_SUITE"] == "./spec/amber"
+  puts "WAITING: sleep 1.minute before running the amber specs to ensure all is ok..."
+  sleep 1.minute
+end
 
+# NOTE: Constants should be set before require begins.
 ENV["AMBER_ENV"] = "test"
 ENV[Amber::Support::ENCRYPT_ENV] = "mnDiAY4OyVjqg5u0wvpr0MoBkOGXBeYo7_ysjwsNzmw"
 TEST_PATH         = "spec/support/sample"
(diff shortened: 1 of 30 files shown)


spec/spec_helper.cr
```diff
@@ -1,5 +1,9 @@
-# NOTE: Constants should be set before require begins.
+if ENV["TEST_SUITE"] == "./spec/amber"
+  puts "WAITING: sleep 1.minute before running the amber specs to ensure all is ok..."
+  sleep 1.minute
```
robacarp: sleeping a whole minute feels like overkill. Could we do something more intelligent, like try to connect with the database and sleep 100ms when it fails to connect?

faustinoaq: > sleeping a whole minute feels like overkill.

I was inspired by TechEmpower Benchmark, TFB used to wait a whole minute before running test, I guess they removed it after replacing all by docker :sweat_smile: 

>  like try to connect with the database and sleep 100ms when it fails to connect?

Oh, ok, what command do you suggest? `psql`, `amber exec`?

faustinoaq: Travis is still failing I guess this can do the work:

^ https://travis-ci.org/amberframework/amber/jobs/363480820#L1454

Maybe I can wait less time, `30.seconds`, WDYT? :sweat_smile:

faustinoaq: https://travis-ci.org/amberframework/amber/jobs/363482182#L1454

^ Failing here as well :sweat_smile:

robacarp: What if we add this to he build and shell out to it before starting the tests? https://github.com/vishnubob/wait-for-it

faustinoaq: Maybe I can create a loop to check if connection and tables are created, like you suggested, WDYT?

I think adding a new dependence to the spec suit is an non required overload :sweat_smile:

epergo:
Hi! According to travis documentation, the name for postgres service is `postgresql` https://docs.travis-ci.com/user/database-setup#PostgreSQL, not plain `postgres`

All builds fail when trying to boot postgres service, even those that success (yeah, weird)

image

I don't know if this change is enough to fix it (probably not) but I think postgres service name should be spelled as documentation says, unless there is a reason to have the service specified like that.

Hope it helps!

epergo:
Are you sure @faustinoaq ? [this](https://travis-ci.org/amberframework/amber/jobs/364004678) build has failed after fixing postgres service name, I'm afraid it will fail again at any time.

Maybe is possible to log the outcome of:

```crystal
MainCommand.run ["db", "drop"]
MainCommand.run ["db", "create", "migrate"]
```
(Don't know if it is possible) To see if is failing when creating/migrating the database

faustinoaq:
@epergo Oh, I spoke too soon :sweat_smile: I guess there is some other thing we're unaware of

I guess we can log the `MainCommand` @eliasjpr :eyes: 

Maybe we can ensure those commands are executing sequentially

faustinoaq:
@epergo Good point! :+1:

I guess we did that because we need to test something. Actually I don't have a full knowledge about the spec suite we're using, I'm sure @eliasjpr has some comments about this :+1:

faustinoaq:
Moved to https://github.com/amberframework/amber/pull/740

PR #727, opened 2018-04-03
target generated shard.yml based on compile-time flags

robacarp:
### Description of the Change

Current release process requires editing the template for the generated shard.yml to point to a different place for the release and stable branches. This would remove that burden from the releaser / release scripts and place it instead in the code.

@docelic I know you have contributed to the makefile in the past, so I'm tagging you here. My make is weak, and I'm sure that there's a way to clean this logic up and maybe even de-duplicate it. 

Comments and suggestions appreciated.


src/amber/cli.cr
```diff
@@ -4,4 +4,12 @@ require "./exceptions/*"
 require "./environment"
 require "./cli/commands"
 
+{% if flag? :amber_release %}
+  AMBER_SHARD_TARGET = :current_version
+{% elsif flag? :amber_stable %}
+  AMBER_SHARD_TARGET = :stable_branch
+{% else %}
+  AMBER_SHARD_TARGET = :master_branch
```
robacarp: Certainly there's a better place for these definitions, don't you think, @eliasjpr @elorest?

eliasjpr: We do have the `.amber.yml` file that stores CLI config settings.

docelic:
@robacarp great, I see some little things that could be improved in the Makefile recipe, but I'll think about the simplest way to add/make them. Will follow up as soon as I have the idea worked out.

src/amber/cli/templates/app/shard.yml.ecr
```diff
@@ -18,8 +18,13 @@ targets:
 dependencies:
   amber:
     github: amberframework/amber
+<% if AMBER_SHARD_TARGET == :current_version -%>
+    version: <%= Amber::VERSION %>
+<% elsif AMBER_SHARD_TARGET == :stable_branch -%>
+    branch: stable
+<% elsif AMBER_SHARD_TARGET == :master_branch -%>
```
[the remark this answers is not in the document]
faustinoaq: Wow, this is nice! :+1:

faustinoaq:
@elorest Interesting idea :+1:

robacarp:
@elorest yeah, I was thinking about that after our discussion the other day. Does it make sense to have both? So the default is to target the current branch but it can be overridden with the CLI?

elorest:
@robacarp So there are 2 issues one is the version of the CLI which might not be compatible with the version/branch of project. This PR doesn't deal with that as that does have to be compiled in from the correct code.

However the current issue of what version of amber the generated shard file creates is probably best handled by a cli option. If we use compile time flags in the makefile we'll run into an additional complication where we could checkout master and compile it to only create projects pointed to stable even though a cli generated from master could easily create a project that won't work with the required version. 

If someone has to choose a cli option `amber new myproject --version=0.7.2` or  `amber new myproject --branch=master` at least the expected result is more obvious.

robacarp:
@elorest I think we might be talking about different things. Being able to target random versions of amber when generating an app feels like it's going to open the door for a bunch of weird bugs where the generated app has code from a newer Amber, but it's targeted at an older release (or vice versa).

Can you lay out a scenario to match what you're talking about to help me understand you?

faustinoaq:
Actually, the current release process is a bit complex because we need to be careful to change the `branch: master` line by the `version: <%= Amber::VERSION %>` one

So, another advantage is that we don't need to comment/uncomment things before releasing a new version, because this :point_down: will already cover that.

```ecr
    <% if ENV["AMBER_ENV"]? == "test" %>
    branch: master
    <% else %>
    version: <%= Amber::VERSION %>
    <% end %>
```

So, the `shard.yml.ecr` template will point always to latest version unless we use:

```
AMBER_ENV=test amber new myapp  # already used by crystal spec suite our Travis CI
```

src/amber/cli/templates/app/shard.yml.ecr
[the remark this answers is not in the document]
faustinoaq: @robacarp Try using this:

```ecr
    <% if ENV["AMBER_ENV"]? == "test" %>
    branch: master
    <% else %>
    version: <%= Amber::VERSION %>
    <% end %>
```

Actually, `branch: master` is only used for tests

src/amber/cli.cr
faustinoaq: Please, see my comments here: https://github.com/amberframework/amber/pull/727#issuecomment-380900041

faustinoaq:
I opened a issue here: #759

elorest:
@robacarp 

> @elorest I think we might be talking about different things. Being able to target random versions of amber when generating an app feels like it's going to open the door for a bunch of weird bugs where the generated app has code from a newer Amber, but it's targeted at an older release (or vice versa).

You are correct. That is the very scenario that the current, somewhat complex, approach solves.
Your suggested solution has the same issues, although in my mind it's even more unpredictable because it leads to the binary being created in multiple states... 

For instance if you were to checkout master and run `make release` it will compile a binary hard coded to create projects with master generators but pointing the last release; a condition which can cause errors and is hard to debug.

Alternatively if you checkout stable and run `make` it will just create a binary which points to master by default but the generated code will be from stable.

> Can you lay out a scenario to match what you're talking about to help me understand you?

My suggestion was just that if someone implicitly stated the version when generating a project it would be less of a surprise if they ran into issues. 
It would also allow someone who knew if it would cause issues or an amber developer to easily generate the desired shard.yml in cases where it won't break anything, or to test if it will break anything....

docelic:
Alternatively, we could still use the idea but have it create differently named binaries in bin/, like amber-<version> or similar.

faustinoaq:
After reading your comments (and my comments :sweat_smile: ) again, I think actually amber generated `shard.yml` is fine. I mean, adding a overhead to compile the right branch/version is not good IMO. Currently behaivior:

- if an user compile amber from `master` branch then `shard.yml` will point to `branch: master`
- if an user compile amber from `tagged` version then `shard.yml` will point to that version `version: X.X.X`
- if an user compile amber from `stable` version then `shard.yml` will point to latest version `version: X.X.X`

^^ This works for most users I think

So, WDYT? Do we need this PR? :sweat_smile:

docelic:
I agree with your comments @faustinoaq in that even the current behavior should be OK.

PR #740, opened 2018-04-10
[CI] Attempts to fix Build Spec For Consistency

eliasjpr:
- Removes the Prepare DB YAML, this was use for when CI used to run on a
docker image which is no longer the case

- Renames the Postgres Travis CI service to Postgresql

### Alternative Setup

The following  setup can be use to only run `build_spec` for master branches and not PRs

```yaml
dist: trusty
sudo: required
language: crystal

services:
  - postgresql
  - redis-server

cache:
  directories:
    - bin
    - lib
    - .shards

jobs:
  include:
    - stage: test
      script: bin/ameba
      script: crystal spec ./spec/amber
    - stage: regression
      script: bin/ameba
      script: crystal spec ./spec/build_spec.cr -D run_build_tests
      if: branch = master

notifications:
  webhooks:
    urls:
      - https://webhooks.gitter.im/e/aaf02221d4649d70b384
    on_success: change  # options: [always|never|change] default: always
    on_failure: always  # options: [always|never|change] default: always
    on_start: never     # options: [always|never|change] default: always


```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,9 @@ cache:
     - .shards
 
 script:
-  - bin/ameba ./src && crystal spec ./spec/amber && crystal spec ./spec/build_spec.cr -D run_build_tests
+  - bin/ameba ./src
+  - crystal spec ./spec/amber
+  - crystal spec ./spec/build_spec.cr -D run_build_tests
 
 notifications:
   webhooks:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,19 +3,17 @@ sudo: required
 language: crystal
 
 services:
-  - postgres
+  - postgresql
   - redis-server
+
 cache:
   directories:
-    - bin/
-    - lib/
-env:
-  - TEST_SUITE=./spec/build_spec.cr
-  - TEST_SUITE=./spec/amber
+    - bin
+    - lib
+    - .shards
 
 script:
-  - bin/ameba
-  - crystal spec $TEST_SUITE -D run_build_tests
+  - bin/ameba ./src && crystal spec ./spec/amber && crystal spec ./spec/build_spec.cr -D run_build_tests
 
 notifications:
   webhooks:
--- a/spec/build_spec.cr
+++ b/spec/build_spec.cr
@@ -26,7 +26,6 @@ module Amber::CLI
       MainCommand.run ["generate", "channel", "Falcon"]
 
       prepare_yaml(Dir.current)
-      prepare_db_yml(CLIHelper::BASE_ENV_PATH) if ENV["CI"]? == "true"
       Amber::CLI.env = "test"
       Amber::CLI.settings.logger = Amber::Environment::Logger.new(nil)
 


eliasjpr:
Not sure how everyone feels about running all the tests in a single instance, but this seems to get rid of the race condition that we sometimes experience running parallel builds. This also takes 2.5 minutes longer. 

An alt

eliasjpr:
@faustinoaq interestingly enough having the script in one liner cut the time by 1.5 minutes approx. It seems that every script entry the state gets cleared out taking some time.

I am running it again to see. Locally the it only takes about 37 seconds to run all the tests :( everything else is setup time on travis.

eliasjpr:
I am noticing that specs are passing consistently for new PRs not sure if its worth it to merge this. WDYGT?

drujensen:
@eliasjpr are we using the postgres service?  If so, we should fix this.  https://docs.travis-ci.com/user/database-setup/#PostgreSQL

drujensen:
just noticed the other PR that addresses the postgresql service.  You can close this in favor of that one.

faustinoaq:
@drujensen This PR is better :sweat_smile: I gonna clone mine if this solve the Travis issue

drujensen:
@faustinoaq 👍

eliasjpr:
@faustinoaq made it multilines

eliasjpr:
> @eliasjpr are we using the postgres service? If so, we should fix this. https://docs.travis-ci.com/user/database-setup/#PostgreSQL

We corrected the `postgresql` here. Am I missing something? @drujensen

faustinoaq:
@eliasjpr check this: https://travis-ci.org/amberframework/amber/builds/364682782#L615

screenshot_20180410_110305

eliasjpr:
@faustinoaq it seems that has to do with the fact that the test was interrupted by the push to the branch

faustinoaq:
@eliasjpr I don't know, seems that this was failing [here](https://travis-ci.org/amberframework/amber/jobs/364680508) and [here](https://travis-ci.org/amberframework/amber/jobs/364661903), even with this change.

Also, the PR #741 is passing now [here](https://travis-ci.org/amberframework/amber/builds/364675930?utm_source=github_status&utm_medium=notification) even without this changes.

Looks like is a random fail after all :sweat_smile:

.travis.yml
```diff
@@ -3,19 +3,19 @@ sudo: required
 language: crystal
 
 services:
-  - postgres
+  - postgresql
   - redis-server
+
 cache:
```
faustinoaq: We should cache `.cache/crystal` as well, WDYT?

eliasjpr: Why we should cache that directory?

faustinoaq: To reduce compilation times even more

I just realize we're checking `bin/#{APPNAME}` before compiling, so NVM, I'm ok without this :+1: 

Sorry for the blocking review :sweat_drops:

Issue #730, opened 2018-04-06
Params not always responding with the correct value for checkboxes


robacarp:
### Description

I'm working with an app that needs raw unvalidated parameters in the controllers. Eventually I'd like to update it to work with parameter validation, but for now it needs to stay that way. Updating to the latest amber had me in a snag around a particular usage of checkboxes.

### Steps to Reproduce

1. Generate amber app from latest release
1. `shards build amber`
1. `bin/amber generate scaffold animal name:string cat:boolean`
1. `bin/amber db migrate`
1. `bin/amber watch`
1. Add code to make animal_controller.cr#animal_params look like the sample below
1. Visit http://localhost:3000/animals/new
1. Check the box and submit, look in the logs for the output.

```crystal
  def animal_params
    puts "params.raw_params['cat']"
    puts params.raw_params["cat"]
    puts "params.raw_params.to_h['cat']"
    puts params.raw_params.to_h["cat"]
    puts "params['cat']"
    puts params["cat"]

    params.validation do
      required(:name) { |f| !f.nil? }
      required(:cat) { |f| !f.nil? }
    end
  end
```

**Expected behavior:** [What you expect to happen]
```text
params.raw_params['cat']
1
params.raw_params.to_h['cat']
`
params['cat']
1
```

**Actual behavior:** [What actually happens]
```text
params.raw_params['cat']
1
params.raw_params.to_h['cat']
0
params['cat']
1
```

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Versions

```text
> bin/amber -v
Amber CLI (amberframework.org) - v0.7.2

> crystal -v
Crystal 0.24.2 (2018-03-10)

LLVM: 5.0.1
Default target: x86_64-apple-macosx
```

### Additional Information

For the time being, it's not a blocker. The workaround is simply to not use `params.raw_params.to_h`.

eliasjpr:
@robacarp you can user `context.params` which is the raw params for the request.

robacarp:
@eliasjpr that may be possible, but `params.raw_params.to_h` should either get removed or fixed. As is, this is a bug

PR #745, opened 2018-04-11
[Params] To hash method not returning correct value

eliasjpr:
Issue: https://github.com/amberframework/amber/issues/730

### Description of the Change

When form submission with repeated form input names it returns the last
value of the list.

Example:

```html
<input name="cat" value="1" type="checkbox" checked>
<input name="cat" value="0" type="checkbox">
```

Actual

```crystal
params.raw_params['cat'] == 1
params.raw_params.to_h['cat'] == 0
params['cat'] == 1
```

Expected

```crystal
params.raw_params['cat'] == 1
params.raw_params.to_h['cat'] == 1
params['cat'] == 1
```

### Resolution

Removes the used of `.merge()` method in favor of manually merging the
hash.

### Possible Drawbacks

This slows down the hash generation of the params

### DIFF
--- a/spec/amber/router/params_spec.cr
+++ b/spec/amber/router/params_spec.cr
@@ -103,6 +103,7 @@ module Amber::Router
 
       it "returns a hash with all params" do
         params.to_h.keys.should eq %w(test _method status _csrf title content)
+        params.to_h["test"].should eq "test"
       end
     end
   end
--- a/src/amber/router/params.cr
+++ b/src/amber/router/params.cr
@@ -65,11 +65,12 @@ module Amber::Router
     end
 
     def to_h
-      query.to_h
-           .merge(form.to_h)
-           .merge(route)
-           .merge(json)
-           .merge(multipart)
+      params_hash = Hash(String, String | Array(String) | Nil).new
+      query.each { |key, _| params_hash[key] = query[key] }
+      route.each { |key, _| params_hash[key] = route[key] }
+      multipart.each { |key, _| params_hash[key] = multipart[key] }
+      form.each { |key, _| params_hash[key] = form[key] }
+      params_hash
     end
 
     private def query


src/amber/router/params.cr
```diff
@@ -65,11 +65,12 @@ module Amber::Router
     end
 
     def to_h
-      query.to_h
-           .merge(form.to_h)
-           .merge(route)
-           .merge(json)
-           .merge(multipart)
+      params_hash = Hash(String, String | Array(String) | Nil).new
```
eliasjpr: Thought of memoizing this but it felt too risky, since the params can change withing the request

robacarp:
Can you add a test which covers the bug?

eliasjpr:
@robacarp I added a spec here https://github.com/amberframework/amber/pull/745/files#diff-027a0c9ac83ea70410609c5a51cdff66R106 but I will add more clear test around form posts.

src/amber/router/params.cr
eliasjpr: The order here is important since we give form a higher priority it will override query and route params

src/amber/router/params.cr
```diff
@@ -65,11 +65,12 @@ module Amber::Router
...
-           .merge(form.to_h)
-           .merge(route)
-           .merge(json)
-           .merge(multipart)
+      params_hash = Hash(String, String?).new
+      query.each { |key, _| params_hash[key] = query[key] }
+      route.each { |key, _| params_hash[key] = route[key] }
+      multipart.each { |key, _| params_hash[key] = multipart[key] }
+      form.each { |key, _| params_hash[key] = form[key] }
+      params_hash
```
veelenga: This should be equivalent but a bit faster/cleaner:

```crystal
query.to_h
     .dup
     .merge(route)
     .merge(multipart)
     .merge(form)
```

spec/amber/router/params_spec.cr
```diff
@@ -94,16 +80,35 @@ module Amber::Router
...
+      end
+
+      it "returns the first value for key" do
+        url_encoded_form_post.to_h["cat"].should eq "1"
       end
     end
   end
 end
+
+def url_encoded_form_post
```
veelenga: These methods need to be a part of a module above and marked as `protected`/`private`. Otherwise, they can be accessed in other files.

faustinoaq:
> how can I get the route and form param if they have the same name? ex: `/post/:username` on `/post/jondoe?username=foobar`

>> One of those will override the other, you cannot choose which
>> Hopefully the route would override the query param
>> (I think the opposite might be a security problem)

Ref: https://gitter.im/amberframework/amber?at=5acd886e080a3850531c40bd

> The order here is important since we give form a higher priority it will overwrite query and route params

@eliasjpr @robacarp Is that the right way?

spec/amber/router/params_spec.cr
```diff
@@ -94,16 +80,35 @@ module Amber::Router
...
+  headers = HTTP::Headers.new
+  headers["Content-Type"] = "application/x-www-form-urlencoded"
+  body = "cat=1&cat=0&cat=-1&test_form=test1&test_form=test2&test_both=form1&test_both=form2"
```
robacarp: sorry, I missed this subtlety of the spec.

src/amber/router/params.cr
robacarp: @veelenga that's essentially the exact code that is causing the bug

veelenga: Nope. It has an additional `dup`.

eliasjpr: This partially works. The problem we are experiencing besides the mutation is that the value return for a given key is the last value present. If you were to run the following. 

```
query.to_h
     .dup
     .merge(route)
     .merge(multipart)
     .merge(form.to_h)
```

You get the following

```bash
Amber::Router::Params #to_h returns the first value for key
     Failure/Error: url_encoded_form_post.to_h["cat"].should eq "1"

       Expected: "1"
            got: "-1"
```

faustinoaq: @eliasjpr Why not invert the order then? :sweat_smile: 

```
form.to_h
    .dup
    .merge(multipart)
    .merge(route)
    .merge(query.to_h)
```

robacarp: The bug here has nothing to do with the way params from different fields are merged into the params hash. It has to do with the way multiple values from the same hash are chosen.

In other words, given this series of form parameters:

```html
<input name="animal" value="cat">
<input name="animal" value="dog">
```

What is the value of `params["animal"]`?
What is the value of `params.raw_params["animal"]`? 
What is the value of `params.raw_params.to_h["animal"]`?

All three should be the same. With the code in master, you get:

`params["animal"] => "cat"`
`params.raw_params["animal"] => "cat"`
`params.raw_params.to_h["animal"] => "dog"`

This is the problem that running a straight `.merge` creates, and why it does not work.

robacarp:
@eliasjpr there may be discussion about what order to put the merge elements in, but that discussion is unrelated to the bug you're addressing in this pull request. Can we leave the merge order the same and address that later, if needed?

eliasjpr:
@robacarp updated accordingly

eliasjpr:
@amberframework/contributors this is ready for another set of eyes.

eliasjpr:
Ping! 🔔

elorest:
I'll read as soon as I get through this meeting. Thanks.

src/amber/router/params.cr
```diff
@@ -64,12 +64,14 @@ module Amber::Router
...
 
-    def to_h
-      query.to_h
-           .merge(form.to_h)
-           .merge(route)
-           .merge(json)
-           .merge(multipart)
+    def to_h : Types::Params
+      params_hash = Types::Params.new
+      query.each { |key, _| params_hash[key] = query[key] }
```
elorest: Looks good.

Sija:
> IMO the order of precedence is
> 
> ```
> query
> route_params
> multipart
> form
> ```
>
> The later overriding the previous.

`route_params` should always be last, otherwise taking `/user/:username` route as an example, sending `username` as multipart/form param would overwrite the former, which is a serious security consideration!

robacarp:
@Sija I agree with you, but that can be addressed in another pull request.

PR #764, opened 2018-04-19
[CLI] change the default recipe source URL

damianham:
Issue: https://github.com/amberframework/amber/issues/750
RE: https://github.com/amberframework/recipes/pull/3

Change the default recipe source URL to add the dist folder and support
URLs as a recipe name.

- changes the default recipe source URL
- adds support for a URL as a recipe name

Issue #750, opened 2018-04-12
Recipes feature review


faustinoaq:
### Description

> Distributing recipes as shards feels like it'll be quite a wedge. They're not versioned, have no dependencies, and really aren't even crystal code. They're just collections of templates and folders. A zip is a great way to package a recipe for download, I just don't think that this way of organizing and distributing "official" amber recipes is going to scale well.

> If Amber is really going to provide a handful of blessed recipes, I see no reason why they can't all live in the same repository. But I'd like them to live as code, not as zips, so pull requests are functional and file history works. A CI system can create zips and publish them to s3 or something for distribution.

> I think it also makes sense for others to be able to publish a recipe as a single git repository, but keeping up with N repositories for blessed recipes feels exausting. Amber is already 21 separate repositories. Imagine trying to update for a breaking change across N repositories, that requires coordinating pull requests and branches across every repository to match the amber core.

by @robacarp 

### Steps to Reproduce

Not applicable

### Versions

Master https://github.com/amberframework/amber/commit/bea1063f640e0c7466989fd50cd0453b812d7a39

### Additional Information

/cc @damianham @drujensen @Adam-Stomski

faustinoaq:
I think keeping recipes source code is fine, then create zip files from source and storing them inside recipe directory is a nice option.

drujensen:
>> Distributing recipes as shards feels like it'll be quite a wedge. They're not versioned, have no dependencies, and really aren't even crystal code.

I'm not following this logic.  A recipe will be dependent on the version of Amber and possibly other shards.  For example, it may be a MongoDB recipe that would inject a shard dependency.  The version of this injected dependency will be specified by the recipe. IMO, It is code, even though its in a templating language.  My take is that we will need a packaging management to handle recipes and that a shard is the best paradigm in crystal.

damianham:
The recipe contains an app component which has a shard.yml with the dependencies listed so once you create an app from the recipe you are using the versions of the dependencies give in the shard.yml.

drujensen:
@damianham what happens when Amber has a breaking change?  Do we need to update all the recipes to avoid broken templates?  How is versioning going to be handled with zip files?

faustinoaq:
@drujensen You have a good point, so I think we need to use one (1) recipe per repo anyway, and keeping the `.zip` packaging in the release page of each repo.

[...]

Issue #749, opened 2018-04-11
Add support for before and after hooks in amber generators


faustinoaq:
### Description

This will allow us to run commands before and/or after generation of an artifact and they could be specified in a yaml file in the recipe

### Steps to Reproduce

Not applicable

### Versions

v0.7.2

### Additional Information

Also, this will allow us to run external tools to perform task like code formatting, installing new dependencies and so on.

drujensen:
@damianham What about allowing for a list of recipes in priority order?
```
basic/granite
djensen/react
```
This would require searching for the template in `djensen/react` first and if not found, searching in `basic/granite`

This would allow you to override templates and/or add new templates without requiring you to copy everything from scratch.

drujensen:
It would be nice to be able to fork a recipe and include it as a shard.  The  priority search order could be the shard priority order.

drujensen:
@damianham Is there anyway we can avoid creating our own packaging manager?  `recipe.yml` is just a pre-cursor to another dependency management and version control.

damianham:
I see it as just a way to specify some configuration items for the recipe, principally the parent recipe that the recipe extends.  I don't think we will have the ability to specify a version of the parent recipe that a recipe extends as the release version will apply to all recipes so therefore I don't foresee the ability to manage dependencies of a recipe.  There may be a recipes that add dependencies to the generated application's shards.  I think that would be performed by the after_app hook.

Issue #733, opened 2018-04-07
[CLI] Recipe cache per repository


damianham:
### Description

In the initial implementation of the recipes feature the recipe cache does not differentiate between recipe sources.  Under certain conditions the wrong recipe will be used.  The user may setup the recipe source URL in .amber.yml but if the user does not also clear the recipe cache or change the recipe name then the existing recipe in the recipe cache will be used rather than the same named recipe from the new recipe repository.

### Steps to Reproduce

1. user creates an application with a recipe from amberframework/recipes
2. user sets the recipe source URL in .amber.yml to acquire recipes from a different recipe repo
3. user does not clear the recipe cache or does not change the recipe name in .amber.yml
4. user creates scaffolding - the recipe from the amberframeworks/recipes in the recipe cache will be used 

**Expected behavior:** 

The recipe from the configured recipe source should be henceforth.

**Actual behavior:** 

The recipe from amberframework/recipes is used henceforth.

**Reproduces how often:** 

100%

### Versions

>0.7.2

PR #766, opened 2018-04-21
Add Jennifer support

imdrasil:
**WIP**: this PR is not ready to be merged

ATM this PR is created to describe my current vision how to integrate Jennifer to Amber. After clarifying all aspects and details - tests suite will be added and related Jennifer release created.

### Description of the Change

Add Jennifer and Sam support 

### Alternate Designs

The reason why Sam has been added as well is its usage by some Jennifer processes.

### Benefits

- Jennifer may be set as default ORM
- Sam tasks may be invoked using `amber` utility
- micrate and default jennifer migration dsl may be used depending on a project settings


shard.yml
```diff
@@ -70,6 +70,10 @@ dependencies:
     github: amberframework/teeplate
     version: ~> 0.5.0
 
+  inflector:
```
faustinoaq: We already have this shard some lines below :sweat_smile:

damianham:
Supporting many different ORMs is going to get complicated in recipes.  Maybe the model folder in a recipe should have sub folders for each ORM.  It would simplify the model template considerably.

I didn't look too deeply into Jennifer but support for scopes is nice.  How does it present errors ?

imdrasil:
hi @damianham  
about subfolders for models - totally agree with this approach. If this will be approved - will rework it with pleasure
about errors - if you are talking about validation errors, ATM they are stored as a string-based sentence, but they are generated using i18n translations (very similar how Rails ActiveModel does)

shard.yml
imdrasil: missed this ). Will fix later

damianham:
Hi @imdrasil 

Actually I wouldn't bother reworking it because I don't think this is the right solution.  I think the right solution to provide support for Jennifer is a recipe.  It is a new feature that has just been merged into the main branch and over time I think it will solve many issues.  There have been discussions about supporting plugins, and different build/packaging systems and even a suggestion to remove the built-in generators and replace them with a default recipe.  Support for Jennifer could be provided by a recipe also and the core amber code will not need to be modified.  However you have added jennifer to the model options and it gave me a decent idea which builds on the idea of extracting the built-in app generator and using a default recipe instead.  What we could do is have a collection of default recipes for the various options and build the recipe name from the given options if no recipe option is given, e.g.

amber new myapp

create a new amber app from a recipe at default/pg/granite/slang
(equivalent to amber new myapp -r default/pg/granite/slang)

amber new myapp -d mysql -m jennifer -t ecr 

create a new amber app from a recipe at default/mysql/jennifer/ecr
(equivalent to amber new myapp -r default/mysql/jennifer/ecr)

eliasjpr:
@imdrasil this is awesome! 

@damianham at the time there was no `recipe` feature, but you are right having this as a recipe is the ideal approach the good thing here is that I believe this entire PR can be the  recipe with small to none of the changes, correct? 

I will review this PR, it will take some time since is a big PR

damianham:
@eliasjpr yes that is correct this entire PR could be provided by a recipe.  I have submitted a new PR to the amberframework/recipes repo which creates basic/granite and basic/crecto so to support Jennifer it is a case of copy the granite or crecto recipe to basic/jennifer and then amend accordingly.

imdrasil:
@damianham found it. So I will wait until it will be merged and rework this PR as a recipe. What do you think?

damianham:
@imdrasil yes I think that is the best way forward.

damianham:
@imdrasil the basic recipes have been merged into the recipes repo now so you could start to implement a jennifer ORM recipe

imdrasil:
Hm, as I can see recipes includes only markup. So I should move all markup there and leave all specific classes in this PR, am I right?

damianham:
I think that this PR will not be needed, everything to use jennifer in an app would be built in to the recipe.  What you would probably do is start with a copy of basic/granite  in basic/jennifer and modify the model and scaffold accordingly and if the Jennifer ORM model provides the same kind of interface to views as granite then you will probably not need to modify the scaffold views.  I think the files to modify in the basic/jennifer recipe would probably be:
basic/jennifer/app/shard.yml.lqd
basic/jennifer/app/config/initializers/database.cr.lqd
basic/jennifer/model/src/models/{{name}}.cr.lqd
basic/jennifer/model/spec/models/{{name}}.cr.lqd
basic/jennifer/scaffold/controllers/src/controllers/{{name}}_controller.cr.lqd
basic/jennifer/scaffold/controllers/spec/controllers/{{name}}_controller_spec.cr.lqd

Then you can add a test in spec/recipes_spec.cr for your recipe.

If you want to go the extra mile you could also create a misc/modular_jennifer recipe and follow the directory layout of the misc/modular recipe which simply places the controller, models and views in the same folder.  Adding a misc/modular_jennifer recipe after creating the basic/jennifer recipe should be as simple as copying the controller and model templates in the right place.

damianham:
That is true unless there is a react/jennifer recipe.  An app is created with a recipe and subsequent amber generate scaffold uses the scaffold template in the recipe that was used to generate the app.  However in https://github.com/amberframework/amber/issues/749  I discuss an option whereby a recipe could depend on another recipe.  The problem we will soon have a situation where we update the amber release and we than have to update the shard.yml file in all recipes, so recipe maintenance will become a real chore.  By moving to a recipe dependency chain a recipe could be diffs from a base recipe.  We are planning to implement before and after hooks and plugins which will give us more flexibility.  The hooks will be for each kind of artifact being generated and stored within the recipe/hooks folder.  We could also move the ORMs to plugins and install an ORM plugin during app generation.  We may end up with everything in recipes and plugins and the existing built-in generators being completely removed.  I just need to find some time to get around to implementing those features.  For the moment if you can implement a basic/jennifer recipe that would be awesome.

drujensen:
@damianham @faustinoaq Are we going to remove the flags `-d`, `-m` and `-t` and replace them with different recipes and plugins?

drujensen:
@damianham How do you upgrade the recipes on a project when a new version of Amber comes out?

damianham:
@drujensen I don't foresee any need to change the behaviour for the -t flag at the moment as it just selects whether to generate .ecr or .slang views and both are provided in the recipe template.  We can retain the -m flag and internally convert that to an appropriate recipe by appending the given ORM name to 'basic/'.  E.g -m granite (or no -m flag) would use basic/granite, -m crecto would use basic/crecto, -m jennifer would use basic/jennifer -m mongo basic/mongo etc.  However I also like the idea of moving the ORM to a plugin so the -m would resolve to an ORM plugin and if we can move the ORM to a plugin then I am sure we would be able to move the views to a plugin so -t could support all of the template languages supported by crystal using the same principle.

One of the problems we will face with plugins is that with the default layout and the modular layout we have 2 directory layouts so plugins will have to support these 2 directory layouts.  TBH I would be in favour of moving completely to the modular layout to avoid this issue.  If you look at the react/preact_redux recipe I have everything related to a feature located in one folder i.e. model, controller, views, javascript and stylesheet.  It is a good way to organise code. For example MEAN.JS organises an application this way.

As you pointed out in another PR or issue, maintenance of recipes is going to be a lot of work as new amber releases come out and so separating the generation of an amber app into a base app with the shards and config etc and then adding plugins according to the command line arguments may alleviate the burden somewhat.

I think @robocarp already mentioned that we would need to have a mechanism to auto update recipes as new releases come out.  There is also an issue https://github.com/amberframework/amber/issues/733 to have a recipe cache per repository so maybe we can implement auto updating recipes along with that enhancement.  Once your app is generated then you would have to update shards and config manually as you do now to use a newer version of amber or shards etc.

At the moment I am making decisions regarding recipes but I would really appreciate any suggestions and guidance from everyone about how you would like it to work.  Also are we happy with the recipes organisation at the moment?  Is it a good idea to have default at the root and everything else in a sub folder?  Should modular recipe be at the root alongside default?  Should modular be the default app template?  Should we work towards a base app and then additional plugins as discussed above?  There are many questions and everyone's feedback is appreciated.  What would be even more appreciated is people chipping in and implementing some of these new features :-)

faustinoaq:
@drujensen I understand your concerns, although, I think recipes is a great idea

> Here are my concerns with recipes so far:

- **Code duplication**: I think we can reuse code by using symlinks. Also We can manage this by just using `amberframework/recipes` for _blessed_ recipes, so you still can use your own recipe url via `recipes_source` on your `.amber.yml` file. This can already allow you to have your own recipes stores one-recipe-one-repo or whatever your want.
- **Zip files in git repositories**: I solved that, please see: https://github.com/amberframework/recipes/pull/14 Now `.zip` files are not stored on repo, and recipes are automatically tested and deployed on master, always pointing to `dist` tag.
- **Recipes all living in the same git repo**: we already discussed that on https://github.com/amberframework/amber/issues/750 so I think, we can reuse code here and avoid to have many repositories. So the `amberframework/recipes` repo is easier to maintain and review IMO.
- **Custom package management for recipes**: This one is a bit confusing, recipes are not packages, they are just pre-created amber projects with custom config.
- **Maintenance of Amber updates**: I think I already managed to do that as well, each recipe is a **full** amber project, so, you always can use the amber version that comes within the recipe, by example: `amber new ambegram -r misc_react` can be generated with amber `v0.8.5` but the recipe itself can use `v0.8.1` via `bin/amber`. Also I think we can offer important snapshots to save recipes across versions, by example if amber `v0.8.0` is not compatible with `v0.9.0`, and I want to have 2 `modular` recipes for each version, then I can create a folder with `misc/modular@0.8.0` with custom changes (even reusing code via symlinks) then using this `amber new blog -r misc_modular@0.8.0` (already possible with https://github.com/amberframework/recipes/pull/14)

> Unfortunately, I don't have any time right now to help. I will have some time this summer. I have some idea's on how to address these concerns (mainly changing recipes into crystal libraries as shards) and will work on this when I get a chance.

I think recipes as shard would be an interesting idea (just like [amber-saludo](https://github.com/faustinoaq/amber-saludo) example), although, if you use shards then you can't generate a full project from scratch like recipes do, you will need to have a project first, well at least a `shard.yml` file and compile the generator again and again for each recipe you want to try.

I think recipes is a more elegant and easier to maintain approach.

Recipes are easier to test, maintain and deploy. One repo for blessed recipes is the way to go IMHO. :sweat_smile: 

WDYT?

drujensen:
@faustinoaq I think we need to discuss our goals with Amber and I don't see hosting and maintaining recipes as one of them.

IMO the Amber team does not have the band-width or time to maintain recipes for possibly hundreds of different flavors of front-end and back-end combinations.

If we do decide to support the ability to `mod` amber, this should be an exercise outside of the core project. This is why I think `shards` is a better approach.  Your example project `amber-saludo` is the right way to do this IMO.

I understand the issue with bootstrapping a new project using shards and here is my proposed solution.

I think we should be able to start a project using `crystal init app blog` and then add the amber shard to it.  Then running a `bin/amber init` command will allow us to bootstrap a project and convert it to an amber project.

With this in mind, I think any modifications to the templates (recipes) could simply be adding a shard to the project before running the `amber init` command.  

There are many advantages to this approach:

1. We no longer maintain multiple flavors of amber recipes.  We can remove the `-m` and possibly `-t` flags and keep the code base at a maintainable level.
2. Anyone can create a new flavor of amber template by forking our base recipe repo and modifying it.  
3. We remove hosting recipes in the cloud and zip files as a dependency.
4. We remove maintaining different flavors of templates and duplicate code.

I would like to make sure we stay focused on the goals we set for Amber and I don't see maintaining multiple flavors of different recipes as one of those goals.

damianham:
@drujensen I agree that recipes are beyond the original goals of Amber and maintaining many blessed recipes could become a time consuming task.  However, the fact that we curate blessed recipes also gives us the ability to ensure they stay up to date and perform well with no nasty surprises.  I based the idea and implementation a little bit on Yo http://yeoman.io/ but made it a bit easier to create an app from a recipe by naming the recipe and downloading on the fly rather than the Yo method of installing a generator as an npm module.  Once you have an amber binary on the path creating a new app from a recipe is very simple and does not involve pre-installing generators.  You seem to be in favour of the Yo method of installing the generator before use.

There is one way in which we could have the simplicity of the recipe name as an arg to the new app generator while removing blessed curated recipes from the core project and that is to have a shard which converts the recipe name to a URL to the recipe zip file which could be anywhere on the web.  https://github.com/amberframework/amber/pull/764 adds support for a recipe URL but as a user of the recipe feature I would rather type
```
amber new myapp -r basic/granite
```
than
```
amber new myapp -r https://raw.githubusercontent.com/amberframework/recipes/master/dist/basic/granite
or
amber new myapp -r https://raw.githubusercontent.com/damianham/recipes/master/dist/basic/granite
```
hence from my point of view, curated recipes, either in their current form or via a shard that converts the name to a URL, are a much easier option for end users.

imdrasil:
@faustinoaq agree with a lot of mentioned stuff, but:
- AFAIK I can't use both react recipe and crecto (e.g.) or any other recipe
- a lot of binaries in bin folder IMO may be friendly enough if their launch is automated

faustinoaq:
> AFAIK I can't use both react recipe and crecto (e.g.) or any other recipe

@imdrasil Yeah, we're still work-in-progress, IMO would be possible to create a `crecto_react`  recipe (or any recipe combination), although, I think is better to create your own recipe, and host it in your own repo like `imdrasil/crecto-react` (Just like if you try to publish an amber example using amber+crecto+react), then you upload a `.zip` file in your `imdrasil/crecto-react/releases` page and finally this would be ready to use on `amber new CrectoReactApp -r <recipe.zip>`
 
> a lot of binaries in bin folder IMO may be friendly enough if their launch is automated

Yeah, you're right. I proposed that before, but we concluded this is going to be too complex and the extra layer is useless and hard to maintain, please see my issue here: https://github.com/amberframework/amber/issues/573

Then, @damianham implemented recipes and we considered this is a better way to solve the plugin system. 

I was the first who had the idea of amber plugin system, that's the reason why I created [amber-saludo](https://github.com/faustinoaq/amber-saludo) and after testing recipes for a while I think @damianham approach is way better than mine.

| comparison | binary plugin | recipes |
| --- | --- | --- |
| fast creation | no (requires downloads or clone repo plus previous compilation) via `shard build` | yes (just download zip file) |
| fast build | no (requires `shard install` and build just for the plugin) | yes (just compile your new amber project) |
| easy to config  | no (requires to setup `shard.yml` target and dependencies) | yes (just use `amber new <name> -r <recipe>`) |
| easy to maintain | no (requires to track and review multiples repos, issues and PRs) | yes (just one repo with automated tests and deploy) |
| easy to create | no (requires to figure out how amber generators work) | yes  (required folders and guides are already specified by @damianham :tada: ) |

So, in conclusion recipes is way better than binary plugins or shards.

faustinoaq:
BTW, also recipes can test a lot of things like this https://github.com/amberframework/recipes/pull/13

We're not testing Crecto enough on this repo. That's the reason why crecto scaffolding is failing see: #621 

Recipes can test almost everything in the right way and very automated using one job per recipe :tada: 

Also recipes can reduce the build spec time for amber itself (delegating some specs to recipes)

drujensen:
@faustinoaq I am not proposing to create binaries.  I am proposing to deliver the recipes using shards.  Its quite different.  Maybe we can spend some time discussing.

faustinoaq:
> I am not proposing to create binaries. I am proposing to deliver the recipes using shards. Its quite different. Maybe we can spend some time discussing.

@drujensen Oh, Yeah, no problem :sweat_smile: 

I mentioned the binaries idea again because that's the way I proposed amber-saludo :sweat_smile: 

After all, I think using shards to store recipes may be a good idea :+1: 

We can store the required directories inside a shard repo and create the app from `lib/myrecipe` using:

```
mkdir myapp  # 1. a dir is required
shard init  # 2. To create shard.yml
vim shard.yml # 3. To add shard (myrecipe) depencencies and amber
shards install # 4. To install shard (recipe) and amber
shards build # 5. to build bin/amber target
bin/amber new myapp . -r lib/myrecipe  # 6. to create an amber project based on shard (myrecipe)
```

^^ This would be a nice option, because we don't need to install amber first (just crystal && shards) but IMO is easier to just execute:

```
amber new myapp -r /path/to/myrecipe
# or...
amber new myapp -r https://mydomain.com/path/to/myrecipe.zip
# or (blessed recipes)
amber new myapp -r default
amber new myapp -r basic_granite
amber new myapp -r basic_crecto
amber new myapp -r misc_modular
amber new myapp -r misc_react
```

Just imagine the amount of work and time you can save thanks to recipes :sparkles: 

Thank you @damianham :wink:

drujensen:
@faustinoaq see my edited response ^^^

faustinoaq:
I already working on update recipes with latest amber master changes, we only need an `API_TOKEN` to merge https://github.com/amberframework/recipes/pull/14 and start testing recipes with automated spec builds and deployment :tada:

drujensen:
@faustinoaq ok.  I guess I missed the boat on this one.   How do I provide you an API key?

faustinoaq:
> I hope you are following what I am proposing. I really wish I had some time to demonstrate this.

Oh, very interesting idea, I really like it, to me this can be a nice recipes enhancement/refactoring :+1: 

I think we should try the current recipes implementation by @damianham and then we can do some changes like the ideas you proposed :+1: 

> How do I provide you an API key?

Oh, nice, I sent you a personal message on gitter, you can do this on https://github.com/settings/tokens

drujensen:
@faustinoaq we can still support this with my proposal:
```
amber new myapp -r /path/to/myrecipe
# or...
amber new myapp -r https://mydomain.com/path/to/myrecipe.zip
# or (blessed recipes)
amber new myapp -r default
amber new myapp -r basic_granite
amber new myapp -r basic_crecto
amber new myapp -r misc_modular
amber new myapp -r misc_react
```

but I recommend we remove all the blessed recipes and go with:
```
amber new myapp -r drujensen/default
amber new myapp -r drujensen/basic_granite
amber new myapp -r drujensen/basic_crecto
amber new myapp -r drujensen/misc_modular
amber new myapp -r drujensen/misc_react
```
which would point the shard in the github repo.

faustinoaq:
@drujensen Yeah, I understand you. I think using shard (repos) as recipes would require some refactoring because is still not supported by current recipes implementation, and AFAIK the shards (recipes) should be cloned or downloaded from releases page. I think is doable but will require some work.

For now, I think we can test current recipe implementation, :smile: Did you see my message on gitter?

robacarp:
@faustinoaq @drujensen I agree with targeting recipes with shards, and especially with the overlay idea that @drujensen has mentioned a few times. I see that as the "eventual destination" of this feature. Recipes is a big task that needs a lot of iteration before it's going to be perfect.

faustinoaq:
@imdrasil Is this still work in progress?

@robacarp Yeah, I agree recipes can be shards, although, we need to test current implementation I'm already doing a lot of work on recipes to remove code duplication and add more tests about code formatting, ameba and `crystal spec` foreach recipe :sparkles: 

The final goal would be recipes as shards, but we still need some features like `shards install drujensen/granite-recipe` and so on, see: https://github.com/crystal-lang/shards/issues/144. Currently recipes are being deployed and tested automatically, so I think the current recipes implementation by @damianham would be enough.

imdrasil:
@faustinoaq yeah and going to finish refactoring till tomorrow evening (my time). Sorry for such a long running PR - I got a lot of work to do on my work.

faustinoaq:
Hi @imdrasil I just created, https://github.com/amberframework/recipes/pull/20, please take a look! :wink:

PR #771, opened 2018-04-25
Changed visibility scope of some Amber::Server methods

Sija:
### Description of the Change

This PR:

* Makes public some of the `Amber::Server` methods to make possible accessing them from outer scope when needed, like for instance in [raven.cr](https://github.com/Sija/raven.cr) shard when building full request url - see [Kemal integration](https://github.com/Sija/raven.cr/blob/33cf933064ff8d287439b8d36361b1ea00a2ead7/src/raven/integrations/kemal.cr#L19-L22) as an example.
* Moves `.colorize` calls from the method's return value directly to `Amber::Server.start`, where it should've been in the first place.

### Alternate Designs

Monkey-patching.

### Benefits

```cr
# being able to call

Amber::Server.instance.version      # => "Amber 0.7.2"
Amber::Server.instance.scheme       # => "http"
Amber::Server.instance.ssl_enabled? # => false
```

### Possible Drawbacks

None (I would know of).

### DIFF
--- a/src/amber/server/server.cr
+++ b/src/amber/server/server.cr
@@ -53,7 +53,7 @@ module Amber
 
     def start
       time = Time.now
-      logger.info "#{version} serving application \"#{settings.name.capitalize}\" at #{host_url}".to_s
+      logger.info "#{version.colorize(:light_cyan)} serving application \"#{settings.name.capitalize}\" at #{host_url.colorize(:light_cyan).mode(:underline)}"
       handler.prepare_pipelines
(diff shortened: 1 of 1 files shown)


Sija:
ping?

robacarp:
For whatever reason, Travis CI is failing to report on this pull, but [it is passing](https://travis-ci.org/amberframework/amber/builds/371043791) just fine, so I'm going to merge it without the status passing.

PR #774, opened 2018-04-27
Support Procs in respond_with

damianham:
### Description of the Change

When using controller#respond_with all content for each given format
is pre-rendered for the request.  There may be a performance penalty when
rendering large lists which are then discarded or there are more than
2 response formats.

- change the type of @available_responses hash value to allow Procs
- if the response for a format is a Proc use Proc.call as body content
- only set the response status code if it is the default 200
- set the Location header rather than add the url to the header

The status code is only overwritten if it is the default 200.  This allows a redirect in the Proc otherwise when the Proc is evaluated and the response status code is set to 302 the set_response method will overwrite the redirect status code with the respond_with status code.

### Alternate Designs
None considered

### Benefits

Procs are supported for formats in #respond_with so that content for a request format may be
lazy rendered and/or a Proc for a format may issue a redirect.

### Possible Drawbacks
None known


elorest:
Worth pointing out that HTML will be prerendered in the macro regardless.

Did you solve the issue you had posted earlier in gitter?

elorest:
> @elorest OK well that has confused me :-) I thought the final output HTML from a Proc would be rendered in Proc.call with the record set from the database query and I thought the render macro builds a template string builder which is then given the record set from the query at run time.

That's sort of correct. So the macro takes the template and creates a string builder from it which then exists inside of the proc. When the proc is called the string interpolations are resolved and the html string is returned. So most of the work there happens at compile time with a bit more at runtime which is handled by your proc.

src/amber/controller/helpers/responders.cr
```diff
@@ -61,9 +66,12 @@ module Amber::Controller::Helpers
     end
 
     def set_response(body, status_code = 200, content_type = Content::TYPE[:html])
-      context.response.status_code = status_code
+      if body.is_a?(Proc)
+        body = body.call
+      end
+      context.response.status_code = status_code if context.response.status_code == 200
```
elorest: `set_response` shouldn't have logic to set the status code or not. IMO any logic on this should be determined on in the `respond_with` method.

`set_response` should just set whatever status_code is passed in or default 200.

```cr
     protected def respond_with(status_code = 200, &block)
       content = with Content.new(requested_responses) yield
       if content.body
         set_response(body: content.body, status_code: status_code, content_type: content.type)
       else
         set_response(body: "Response Not Acceptable.", status_code: 406, content_type: Content::TYPE[:text])
       end
     end
```

src/amber/controller/helpers/responders.cr
```diff
@@ -61,9 +66,12 @@ module Amber::Controller::Helpers
     end
 
     def set_response(body, status_code = 200, content_type = Content::TYPE[:html])
-      context.response.status_code = status_code
+      if body.is_a?(Proc)
```
elorest: body should be resolved within the instance of content instead of inside of set response. `set_response` should have the Single Responsibility of setting the values passed into it.

```cr
#body inside of Content
       def body
         _body = @available_responses[type]?
         if _body.is_a?(Proc)
           _body = _body.call
         end
        return _body
       end
```

src/amber/controller/helpers/responders.cr
```diff
@@ -61,9 +66,12 @@ module Amber::Controller::Helpers
...
+        body = body.call
+      end
+      context.response.status_code = status_code if context.response.status_code == 200
       context.response.content_type = content_type
-      context.content = body
+      context.content = body.is_a?(String) ? body : body.to_s
```
elorest: By the time it gets into the `set_response` method body should already be a string.

src/amber/controller/helpers/responders.cr
damianham: The reason that check was added to the line that sets the status_code is that a redirect in a Proc may have already set a status_code to 302 - but the respond_with block has a default status_code of 200 which would overwrite the redirect status_code - hence the additional check so that we only use the respond_with status code if the existing status_code is the default 200.  Otherwise the method is as before apart from calling the Proc to get the Proc body content.

src/amber/controller/helpers/responders.cr
damianham: The reason the Proc.call is in set_response is that if it is Content.body then the Proc will be called twice in respond_with.  Once to test if it exists and again to pass it to set_response.  Calling the Proc in set_response delays the execution of the Proc until the last method call in the request handle process.

src/amber/controller/helpers/responders.cr
damianham: see above

damianham:
any changes still required for this PR ?

src/amber/controller/helpers/responders.cr
```diff
@@ -13,30 +13,35 @@ module Amber::Controller::Helpers
...
-      @available_responses = Hash(String, String).new
+      @available_responses = Hash(String, String | Proc(String) | Proc(Int32)).new
       @type : String? = nil
 
       def initialize(@requested_responses)
       end
 
       # TODO: add JS type similar to rails.
-      def html(html : String)
+      def html(html : String | Proc(String) | Proc(Int32))
```
robacarp: what's the relevance of a proc returning an Int32?

damianham: If your proc redirects the last statement in the redirect code path is setting the response status code (an Int32) so that is the Proc return type

faustinoaq:
Requires #783 to fix Travis specs

spec/amber/controller/respond_with_spec.cr
```diff
@@ -7,6 +7,10 @@ module Amber::Controller
       request.headers["Accept"] = ""
       context = create_context(request)
 
+      Spec.before_each do
+        context.response.status_code = 200
```
robacarp: I don't think you want to do this here... `Spec.before_each` is global and applies to all `it` blocks in the suite, it's not scoped to a describe / context block.

faustinoaq: @damianham Please review this @robacarp comment, we need this because https://github.com/amberframework/recipes/pull/4 depends on it :sweat_smile:

src/amber/controller/helpers/responders.cr
```diff
@@ -13,30 +13,35 @@ module Amber::Controller::Helpers
       ACCEPT_SEPARATOR_REGEX = /,|,\s/
 
       @requested_responses : Array(String)
-      @available_responses = Hash(String, String).new
+      @available_responses = Hash(String, String | Proc(String) | Proc(Int32)).new
```
eliasjpr: Can we create an alias for the Proc type here to simplify. Have it scope to responders

```crystal
alias ProcType = Proc(String) | Proc(Int32)
```

spec/amber/controller/respond_with_spec.cr
damianham: Agree it should not be global but it seems there is no option to run before/after hooks that apply only to the current spec context.  The test suite passes with this in place and fails if I remove it. I did spend some time trying to remove this hook but I couldn't get the test suite to pass so I think it is a case of accept this or reject the PR.  I don't see any real downsides to explicitly setting the status_code to the default status.

src/amber/controller/helpers/responders.cr
damianham: done

src/amber/controller/helpers/responders.cr
eliasjpr: I believe a case statement will clean this up a bit

```crystal 
case _body = @available_responses[type]?
when Proc then _body.call
else _body
end
```

What do you guys think?

src/amber/controller/helpers/responders.cr
eliasjpr: Replace `Proc(String) | Proc(Int32)` with `ProcType` in this file

src/amber/controller/helpers/responders.cr
elorest: @damianham Sorry I'm responding to this so late.

Body could easily be cached as `@body` which would stop it from being resolved multiple times.

src/amber/controller/helpers/responders.cr
elorest: Same as my response above.

src/amber/controller/helpers/responders.cr
elorest: This could be easily solved by just saying that html views, text, json blocks etc shouldn't be setting response types. We're allowing a block to be passed in so that all response strings don't have to be rendered on load, not so that they can do whatever logic they want.

damianham: Apart from not rendering something that will be unused my use case is to redirect for one of the formats - e.g. update redirects for html and renders the object for json.  A redirect sets a specific status code.  An alternative would be to set the status as an argument to the format e.g
```
respond_with do
  html 302, ->{redirect("/users"); flash.success("Saved changes to your profile")}
  json user.to_json
end
```
WDYT?

src/amber/controller/helpers/responders.cr
damianham: yes good point

spec/amber/controller/respond_with_spec.cr
robacarp: The downside is that you don't know what tests will be written in the future that this will cause a weird breakage and possibly sink a day of debugging. I'd much rather have this one line duplicated across the tests that need it than have the global leakage.

src/amber/controller/helpers/responders.cr
```diff
@@ -45,7 +54,10 @@ module Amber::Controller::Helpers
       end
 
       def body
-        @available_responses[type]?
+        @body ||=(  case _body = @available_responses[type]?
```
elorest:
```cr
       @body ||= begin
           case _body = @available_responses[type]?
           when Proc
              _body.call
           else 
             _body
           end
         end
```

src/amber/controller/helpers/responders.cr
```diff
@@ -61,7 +73,7 @@ module Amber::Controller::Helpers
     end
 
     def set_response(body, status_code = 200, content_type = Content::TYPE[:html])
-      context.response.status_code = status_code
+      context.response.status_code = status_code if context.response.status_code == 200
```
elorest: Add warning when not being overwritten.

src/amber/controller/helpers/responders.cr
```diff
@@ -13,30 +16,36 @@ module Amber::Controller::Helpers
...
+      @available_responses = Hash(String, String | ProcType).new
       @type : String? = nil
+      @body : String | Int32 | Nil = nil
 
       def initialize(@requested_responses)
       end
 
       # TODO: add JS type similar to rails.
-      def html(html : String)
+      def html(html : String |ProcType)
```
faustinoaq: Indentation issue here `def html(html : String | ProcType)`

src/amber/controller/helpers/responders.cr
```diff
@@ -13,30 +16,36 @@ module Amber::Controller::Helpers
...
         @available_responses[TYPE[:html]] = html; self
       end
 
-      def xml(xml : String)
+      def xml(xml : String |ProcType)
```
faustinoaq: `:dito:`

drujensen:
@elorest @faustinoaq looks like this is ready.  Please re-review.

faustinoaq:
> looks like this is ready. Please re-review.

Yeah, just some indentation issues here: https://github.com/amberframework/amber/pull/774#discussion_r191063634

@damianham ^^

PR #775, opened 2018-04-29
Investigate Travis CI failures

faustinoaq:
### Description of the Change

Travis is still failing. This PR tries to fix that.

screenshot_20180429_171801

 ^^ https://travis-ci.org/amberframework/amber/jobs/372591645#L1455

### Alternate Designs

Investigating...

### Benefits

Reliable specs

### Possible Drawbacks

Investigating...


faustinoaq:
I discovered this "ghost" failure on Travis CI only happens "sometimes" during original (first) build,

image

^^  if you restart the build using restart build button, the issue dissapear.

faustinoaq:
Also this issue is not related to database connection but database creation and migration

I was able to reproduce this issue on my local machine by using:

```
amber generate model Bat user:reference name:string body:text age:integer published:bool
crystal spec # Without executing amber db migrate
```

^^ If I run migrations before executing specs the issue dissapear

So, Looks like the migrations are runing too late or some issue is happening with them on Travis CI specifically.

faustinoaq:
I stopped triggering builds because seems the `sleep`'s doesn't fix the issue

I think we should use some mechanism to ensure `db migrate` has been executed successfully

faustinoaq:
Travis CI is passing now, look like the issue is a bug inside `amber db migrate` (a.k.a micrate)

This issue consist in migration filenames generated by micrate using timestamps

We are using milliseconds, although these files are generated too fast, so,  name get overlapped "sometimes".

Is like a _race condition_ when filenames are generated.

I suggest to fix it inside micrate by using incremental numbers + plus timestamps, see https://github.com/amberframework/micrate/issues/20

faustinoaq:
Investigation has concluded successfully :sunglasses: 

Now this PR has been replaced by #783 , #784 and #785 :tada:

faustinoaq:
Checking filename is a bit complex and very hacky, so, this is fixed now by micrate v0.3.1, please see https://github.com/amberframework/micrate/releases/tag/v0.3.1

PR #777, opened 2018-05-03
Use break instead of exit in Amber::Server#start

c910335:
### Description of the Change

I found that Amber isn't compatible with [Mass Spec](https://github.com/c910335/mass-spec) because these `exit` exit the whole program including tests.
Mass Spec is a Web API testing library that wants to support all the Crystal Web Frameworks, but these `exit` close the possibility.
Thus I want to replace them with `break`.

### Alternate Designs

`return`

### Benefits

- If there are some codes after `Amber::Server.start`, they'll be executed when closing.
- Support [Mass Spec](https://github.com/c910335/mass-spec)

### Possible Drawbacks

No

### DIFF
--- a/src/amber/server/server.cr
+++ b/src/amber/server/server.cr
@@ -59,9 +59,9 @@ module Amber
       server.tls = Amber::SSL.new(settings.ssl_key_file.not_nil!, settings.ssl_cert_file.not_nil!).generate_tls if ssl_enabled?
 
       Signal::INT.trap do
+        Signal::INT.reset
         logger.info "Shutting down Amber"
         server.close
-        exit
       end
 
       loop do
@@ -69,7 +69,7 @@ module Amber
           logger.info "Server started in #{Amber.env.colorize(:yellow)}."
           logger.info "Startup Time #{Time.now - time}".colorize(:white)
           server.listen(settings.port_reuse)
-          exit
+          break
         rescue e : Errno
           if e.errno == Errno::EMFILE
             logger.error e.message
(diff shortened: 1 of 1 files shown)


src/amber/server/server.cr
```diff
@@ -61,23 +61,22 @@ module Amber
       Signal::INT.trap do
         logger.info "Shutting down Amber"
         server.close
-        exit
       end
```
faustinoaq: @c910335 I think the `exit` is needed here otherwise we won't be able to exit amber process inside the signal trap.

Perhaps, we can use `at_exit` instead:

```cr
at_exit do
  logger.info "Shutting down Amber"
  server.close
end
```

WDYT?

c910335: `Signal::INT.trap` is for ^c, and `at_exit` is for `exit`.
Their purposes are different.

c910335: I think `server.close` is enough to exit the process.

c910335: I found that adding `Signal::INT.reset` would be better.

```crystal
Signal::INT.trap do
  Signal::INT.reset
  logger.info "Shutting down Amber"
  server.close
end
```

Then the first ^c will try to close the server, and the further ^c will kill the process.

WDYT?

faustinoaq: @c910335 Interesting idea, let me try that :+1:

faustinoaq: @c910335 Can you add `Signal::INT.reset`? so I can approve this :sweat_smile:

c910335: @faustinoaq done

PR #783, opened 2018-05-12
Fix race condition on migration files

faustinoaq:
### Description of the Change

This PR fixes Travis failures discovered on #775 

### Alternate Designs

This approach check migration files before create them.

### Benefits

Fixes random Travis CI failures

### Possible Drawbacks

No

PR #785, opened 2018-05-12
Enhancements for Travis and Specs

faustinoaq:
### Description of the Change

We changed Travis jobs by a single Travis build because we thought this was causing Travis failures.

I discovered and fixed this failure on #783 

So, this change will speed up our CI

This PR also removes the macro flag to avoid misleading results on specs

### Alternate Designs

No

### Benefits

Speed up CI builds

### Possible Drawbacks

No

### DIFF
--- a/spec/build_spec.cr
+++ b/spec/build_spec.cr
@@ -1,4 +1,3 @@
-{% if flag?(:run_build_tests) %}
 require "./spec_helper"
 require "./support/helpers/cli_helper"
 
@@ -70,4 +69,3 @@ module Amber::CLI
     cleanup
   end
 end
-{% end %}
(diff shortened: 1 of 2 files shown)


faustinoaq:
Requires #783 to fix Travis specs

faustinoaq:
This PR also prints `shards update` inside `build_spec.cr` to see what dependencies are being installed

spec/build_spec.cr
```diff
@@ -35,7 +34,7 @@ module Amber::CLI
       puts "====== DONE Database created #{TEST_APP_NAME} ======"
 
       puts "RUNNING: shard update started..."
-      `shards update`
+      puts `shards update`
```
eliasjpr: Is puts here intentional? I thought we wanted to keep the specs logging to minimal

Issue #790, opened 2018-05-14
partial not expected behavior


jaysneg:
### Description
if create subfolder at layouts and copy there files: _nav.slang and application.slang
and change at layouts/portal/application.slang path of render:
== render(partial: "layouts/portal/_nav.slang")

appears error

### Steps to Reproduce

1. copy application.slang and _nav.slang to folder at layouts
2. try to render partial _nav.slang:
== render(partial: "layouts/portal/_nav.slang")

**Expected behavior:** it will render partial 

**Actual behavior:** Shows error

**Reproduces how often:** always

### Versions

Amber 0.7.2

### Additional Information

it works in this way 

== render(partial: "portal/_nav.slang", path: "src/views/layouts/")

faustinoaq:
Hi @jaysneg Thank for report this issue :+1: 

Can you share us the error output using a [gist](https://gist.github.com/)? :sweat_smile:

elorest:
This isn't technically a bug but it might be a good place to discuss expected behavior. 

Currently if there are more than 2 parts in the path it will assume the root is your project root instead of `src/views`. 

For instance this would work: `render(partial: "src/views/layouts/tmp/_nav.slang")` as would `render(partial: "specs/samples/views/layouts/tmp/_nav.slang")`

`render(partial: "layouts/_nav.slang")` works because there are only 2 parts.

It might be better to only use the `path:` option to set other paths instead of defaulting to that for longer path names.

This would be as simple as getting rid of this conditional in render.cr and making sure the tests still pass.

```cr
    private macro render_template(filename, path = "src/views")
      {% if filename.id.split("/").size > 2 %}
        Kilt.render("{{filename.id}}")
      {% else %}
        Kilt.render("#{{{path}}}/{{filename.id}}")
      {% end %}
    end
```

elorest:
I could probably find time to work on that tonight if desired.

jaysneg:
jaysneg/gist:c5bd12aef52531b84d4647e8074b17ba

shobhitic:
Looks like this has been fixed. Should we close it?

PR #791, opened 2018-05-15
Simplifies render by removing edge case.

elorest:
### Description of the Change

Resolves issue: https://github.com/amberframework/amber/issues/790

Simplifies render by removing edge case, which is mostly only useful in specs. 

Still easily testable by passing in path param.

### Alternate Designs

Leave it how it is.

### Benefits

More straight forward with fewer magics to remember.

### Possible Drawbacks

Less magic, but more readable.

### DIFF
--- a/spec/support/fixtures/controller_fixtures.cr
+++ b/spec/support/fixtures/controller_fixtures.cr
@@ -84,11 +84,11 @@ end
 
 class RenderController < Amber::Controller::Base
   def render_template_page
-    render("spec/support/sample/views/test/test.slang", layout: false)
+    render(path: "spec/support/sample/views", template: "test/test.slang", layout: false)
   end
 
   def render_partial
-    render(partial: "spec/support/sample/views/test/_test.slang")
+    render(path: "spec/support/sample/views", partial: "test/_test.slang")
   end
 
   def render_with_layout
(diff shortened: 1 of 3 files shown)


faustinoaq:
@amberframework/contributors  @elorest Should we update [Layour Documentation](https://amberframework.gitbook.io/amber/guides/views#layouts) after this is merged?

elorest:
@faustinoaq I just looked through them and it doesn't seem to change the existing docs. There are somethings in there that are wrong though that I need to fix. How do I access the current docs. Can I still edit the github docs project?

faustinoaq:
> I just looked through them and it doesn't seem to change the existing docs. There are somethings in there that are wrong though that I need to fix. How do I access the current docs. Can I still edit the github docs project?

You can do both, edit directly on GitBook or use [amber docs repository](https://github.com/amberframework/docs)

faustinoaq:
@damianham I think this change just broke modular recipe :sweat_smile: 

See: https://travis-ci.org/amberframework/recipes/jobs/383198970

damianham:
Yes it did.  @elorest this change needs backing out so that people can continue to use the modular recipe.  An alternative is to modify amber_render_module and add render_template macro with the original macro implementation.

faustinoaq:
> An alternative is to modify amber_render_module and add render_template macro with the original macro implementation.

I think we can do that as workaround, WDYT? :sweat_smile: 

This PR also conflicts with #808 /cc @elorest

damianham:
merged the PR @faustinoaq thanks for your great work on everything

PR #803, opened 2018-05-19
fixes blocking cors issues

### LINKED ISSUE #799: Couple of issues with CORS in master
@eliasjpr 

### Description

* ~~I dont think that `preflight?` should be memoized to a variable.  It seems to be using the previous request method as the first request on subsequent requests.  Does each request create a new instance of the `CORS` class? - [code](https://github.com/amberframework/amber/blob/master/src/amber/pipes/cors.cr#L75)~~ this is fine

* `Headers::ALLOW_METHOD` should be plural - [link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods) - [code](https://github.com/amberframework/amber/blob/master/src/amber/pipes/cors.cr#L13)

fridgerator:
### Description of the Change

Fixes #799 

* `Headers::ALLOW_METHOD` should be plural [link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods)

* memoized `@_preflight` variable was being used across multiple requests

@faustinoaq @eliasjpr just made the 2nd change related to the `@_preflight` variable

### Benefits

CORS pipe can be used with api / spa applications.

### Possible Drawbacks

None

### DIFF
--- a/spec/amber/pipes/cors_spec.cr
+++ b/spec/amber/pipes/cors_spec.cr
@@ -94,6 +94,21 @@ module Amber::Pipe
           context.response.status_code = 200
           context.response.headers["Content-Length"].should eq "0"
         end
+
+        it "should set the correct headers" do
+          context = cors_context(
+            "OPTIONS",
+            "Origin": "example.com",
+            "Access-Control-Request-Method": "PUT",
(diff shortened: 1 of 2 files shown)


fridgerator:
Added specs to check CORS response headers

PR #804, opened 2018-05-19
fixes not_nil runtime exception in json parser

### LINKED ISSUE #800: JSON parser request.body is sometimes null
### Description

Returning JSON from a controller action `{ok: "ok"}.to_json` throws an error [link](https://pastebin.com/ipdaMQcM).

Amber logs a `200` response, however the browser shows a `500` response.

In [`amber/router/parsers/json.cr`](https://github.com/amberframework/amber/blob/master/src/amber/router/parsers/json.cr#L5) my request body is nil, and this `.not_nil!` is causing the exception

### Steps to Reproduce

I tried to reproduce this with a fresh amber install using jquery `$.ajax` but couldnt.  However my project is using [vue-resource](https://github.com/pagekit/vue-resource) and it happens consistently.

@damianham Also mentioned in gitter this would occasionally happen to him while working on the react recipe

### Versions

* Amber - master branch [link](f6ea26f8d620c42dec33c81b77c90fc74cef5a36)
* Crystal - 0.24.2

fridgerator:
### Description of the Change

Fixes #800 

It is possible for the request body to be nil with a JSON request (DELETE request for example), causing Amber to log a 200 response whereas the actual response in the browser is 500.

This change checks for existence rather than using `not_nil!`.

### DIFF
--- a/spec/amber/router/parsers/json_spec.cr
+++ b/spec/amber/router/parsers/json_spec.cr
@@ -1,0 +1,21 @@
+require "../../../../spec_helper"
+
+module Amber::Router::Parsers
+  describe JSON do
+    describe "self.parse" do
+      it "should allow a nil body" do
+        request = HTTP::Request.new("PUT", "/")
+        request.body.should be_nil
+        params = JSON.parse(request)
+        params.keys.empty?.should be_true
+      end
+
+      it "should parse a json body" do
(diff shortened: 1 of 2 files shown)


fridgerator:
Parser specs did not exist, so I created a JSON parser spec and tested for both nil and not nil request body types

PR #814, opened 2018-05-23
Use context.locale instead of ::I18n.locale

### LINKED ISSUE #776: Current locale mechanism isn't "thread safe"
### Description

Technically, this applies to other shards.
But since the locale is now included in the generator of the new application, I'm writing here.

`Citrine::I18n::Handler` use `I18n.locale` to set current locale for request
But this locale sets globally for all application.
if I'm not mistaken it is called race condition.

Example - we have two requests long and short
* `locale_fast` - really fast
* `locale_long` - lasts three seconds

If we call `locale_long` with one `Accept-Language`, then `locale_fast` with another `Accept-Language` - locale_long respond with wrong locale

### Steps to Reproduce

```crystal
class LocaleTestController < ApplicationController
  def locale_fast
    I18n.locale
  end

  def locale_long
    sleep(3)
    respond_with do
      text I18n.locale.to_s
    end
  end
end
```
**Expected behavior:** 

After paralel call `locale_long` and `locale_fast` with differend `Accept-Language` all respond with own locale key

**Actual behavior:**

`locale_long` respods with last called locale

100% times

### Versions

`Amber CLI (amberframework.org) - v0.7.2`

```
Crystal 0.24.2 [4f9ed8d03] (2018-03-08)

LLVM: 4.0.0
```

### Additional Information

I have some suggestions for resolve this issue.

First - we need new instance of `I18n` store in `context` for each request.
And in helpers use this instance for translations
But now it's impossible - shard `I18n` use class (`@@`) variables for store config and locale value

Second (i use it for now) - store in  `context` i18n_key and use it in trasnslations with `force_locale` argument.

docelic:
This commit makes the locale choice thread-safe.

The corresponding changes to the citrine-i18n shard have already been made.

Fixes #776.

### DIFF
--- a/src/amber/cli/templates/app/shard.yml.ecr
+++ b/src/amber/cli/templates/app/shard.yml.ecr
@@ -23,7 +23,7 @@ dependencies:
 
   citrine-i18n:
     github: amberframework/citrine-i18n
-    version: 0.1.0
+    version: 0.2.0
 
 <% case @model when "crecto" -%>
   crecto:
--- a/src/amber/controller/helpers/i18n.cr
+++ b/src/amber/controller/helpers/i18n.cr
@@ -1,11 +1,11 @@
 module Amber::Controller::Helpers
   module I18n
     def t(*arg)
-      ::I18n.translate(*arg)
+      ::I18n.translate(*arg, force_locale: context.locale)
     end
 
     def l(*arg)
-      ::I18n.localize(*arg)
+      ::I18n.localize(*arg, force_locale: context.locale)
     end
   end
 end
--- a/src/amber/controller/helpers/i18n.cr
+++ b/src/amber/controller/helpers/i18n.cr
@@ -1,5 +1,3 @@
-require "citrine-i18n"
-
 module Amber::Controller::Helpers
   module I18n
     def t(*arg)
--- a/src/amber/controller/helpers/i18n.cr
+++ b/src/amber/controller/helpers/i18n.cr
@@ -1,3 +1,5 @@
+require "citrine-i18n"
+
 module Amber::Controller::Helpers
   module I18n
     def t(*arg)


docelic:
@faustinoaq HTTP::Server::Context is expanded with property "locale" inside citrine-i18n shard. But it appears that inside Amber, "context" isn't seen. Got any suggestions how to solve it?

faustinoaq:
@docelic I think you need to delegate `:locale` to context here:

https://github.com/amberframework/amber/blob/f84fe8c4b43f9e0b307f0cd7ef1a5085e284e45a/src/amber/controller/base.cr#L40


BTW, I just realize test error are being printed twice, Can we fix that? 

https://github.com/amberframework/amber/blob/f84fe8c4b43f9e0b307f0cd7ef1a5085e284e45a/spec/build_spec.cr#L54-L67

We should puts `spec_results` once IMO :sweat_smile: /cc @eliasjpr 

Something like this:

```cr
puts spec_result # => We can use just one puts here because spec_results output without errors are minimal

it "can be executed" do
    spec_result.should contain "Finished in"
end

it "has no errors" do
    spec_result.should_not contain "Error in line"
end

it "has no failures" do
    spec_result.should_not contain "Failures"
end
```

docelic:
@faustinoaq Hm, if I am understanding the error correctly, my problem is not that "locale" doesn't exist, but rather than "context.locale" doesn't exist, which means that somehow Amber::Controller::Helpers::I18n's methods t() and l() were checked before citrine-i18n was required and before it added the property "locale" to Context.

docelic:
Looks like the update to citrine-i18n v0.2.0 solved the issue, and the current error reported by travis is not directly related to the code.

Could someone check/confirm that's the case? Thanks!

faustinoaq:
> Outdated shard.lock (wrong resolver). Please run shards update instead.

Why? Aren't CI tests created from scratch? :sweat_smile: 

Ref: https://travis-ci.org/amberframework/amber/jobs/382976615

docelic:
Do we have an update here? Thanks!

faustinoaq:
> Installing citrine-i18n (0.1.0 at 0.2.0)

@docelic Travis is still failing https://travis-ci.org/amberframework/amber/jobs/382976615#L731

You forgot to update `0.1.0` to `0.2.0` on https://github.com/amberframework/citrine-i18n/blob/master/shard.yml#L3 :sweat_smile:

faustinoaq:
Fixed! https://github.com/amberframework/citrine-i18n/blob/master/shard.yml#L3 :tada:

docelic:
@faustinoaq Oh great, thanks. I indeed forgot, I updated Amber's shard.yml to use the new version :)

docelic:
It was originally @eliasjpr's suggestion that we put it in src/locale. Let's see his current thoughts on that.

PR #816, opened 2018-05-26
Remove Granite.settings.logger from spec

faustinoaq:
### Description of the Change

This PR removes `Granite.settings.logger` from spec because causes compilation error on projects generated with `-m crecto`

### Alternate Designs

Use `spec_helper.cr.ecr` template

### Benefits

Print query logs (very useful for debugging specs)
Fix `crystal spec` on projects generated with `-m crecto`

### Possible Drawbacks

No


src/amber/cli/templates/app/spec/spec_helper.cr
```diff
@@ -11,9 +11,6 @@ Micrate::DB.connection_url = Amber.settings.database_url
 # Automatically run migrations on the test database
 Micrate::Cli.run_up
 
-# Disable query logger for tests
-Granite.settings.logger = Logger.new nil
-
```
veelenga: Maybe we can set it dynamically? Take a look [here](https://stackoverflow.com/a/45908692/1586762)

faustinoaq: @veelenga yeah, but I think Granite logs are very useful, you can see query time and raw sql information, these logs helped me to fix some issues on Amber itself :sweat_smile:

veelenga: Okay, even better 👍

robacarp: I'd prefer to leave it disabled because it generates so much noise. A verbose query log makes it hard to tell what happens when a test is failing. When tests all pass, the output of the suite should be minimal, indicating that nothing unexpected happened.

faustinoaq: @robacarp Crecto spec fails because granite dependency is missing on crecto projects. Maybe we should use a `spec_helper.cr.ecr` template as well, like  this:

```ecr
<%- if database == "granite" %>
Granite.settings.logger = Logger.new nil
<% end %>
```

WDYT?

robacarp: @faustinoaq yes, thats a good idea.

robacarp: @faustinoaq what happened here? Did the idea for the conditional get scrapped?

faustinoaq: @robacarp Yeah, let me add this conditional code :+1:

eliasjpr:
I think instead of removing the logger for Granite we should render the right logger for the right ORM WDYT?

faustinoaq:
@eliasjpr Yeah, I already fixing this :+1:

faustinoaq:
Required by https://github.com/amberframework/amber/pull/850

faustinoaq:
@robacarp for some reason this is still generating Granite logs, any idea?

I checked twice and `spec_helper.cr` is generated with the code to disable logs, Are we enabling logs in some other file? maybe in some initializer?

Ref: https://travis-ci.org/amberframework/amber/jobs/394384754#L736

faustinoaq:
Opened tracking issue here: https://github.com/amberframework/amber/issues/885

PR #818, opened 2018-05-26
Add missing migrations generator

faustinoaq:
### Description of the Change

I'm working on recipes specs and I just realize `amber g model` doesn't generate migration for the model. This PR fixes that :+1: 

### Alternate Designs

No

### Benefits

Migrations are created wit the Models when recipes are used just like current amber templates.

### Possible Drawbacks

No

/cc @damianham


src/amber/cli/recipes/recipe.cr
```diff
@@ -66,6 +66,11 @@ module Amber::Recipes
         Controller.new(name, @recipe, @fields).render(directory, list: true, color: true)
       when "model"
         info "Rendering Model #{name} from #{@recipe}"
+        if model == "crecto"
+          Amber::CLI::CrectoMigration.new(name, @fields).render(directory, list: true, color: true)
```
eliasjpr: This is fine but I would have created a Factory Class to handle this if statement

```crystal

MigrationStrategy.generate_migration_for(model, name, @fields)

class MigrationStrategy
  STRATEGIES = {
    "crecto": Amber::CLI::CrectoMigration,
    "granite": Amber::CLI::GraniteMigration
  }

   def self.generate_migration_for(model, name, fields)
       STRATEGIES[model].new(name, @fields).render(directory, list: true, color: true)
   end
end
```

faustinoaq: @eliasjpr Oh, ok, should we change `template.cr` as well?

https://github.com/amberframework/amber/blob/ce6dec47a0e3c8fe6863a10819d0591700dd4813/src/amber/cli/templates/template.cr#L72-L80

eliasjpr: @faustinoaq after looking more closely at the Recipe class it can use some cleanup. There is a lot of duplication and branching which makes it hard to read and maintain.

I quickly took a shot at this, but I have not test it. Let me know if you would like to pair on it so I can explain the idea behind it.

```crystal
class ModelGenerator
  MIGRATIONS = {
    "crecto": Amber::CLI::CrectoMigration,
    "granite": Amber::CLI::GraniteMigration
  }
  def self.build(recipe, options)
    [
      MIGRATIONS[recipe.model].new(recipe.name, recipe.fields),
      Model.new(recipe.name, recipe.recipe, recipe.fields)
    ]
  end
end

class AppGenerator
  def self.build(recipe, options)
    [
      App.new(recipe.name, *options)
    ]
  end
end

class ControllerGenerator
  def self.build(recipe, options)
    [
      Controller.new(recipe.name, recipe.recipe, recipe.fields)
    ]
  end
end

class ScaffoldGenerator
  def self.build(recipe, options)
    [
      ModelGenerator.new(recipe),
      Scaffold::Controller.new(recipe.name, recipe.recipe, recipe.fields),
      Scaffold::View.new(recipe.name, recipe.recipe, recipe.fields)
    ]
  end
end

class GeneratorBuilder
  TEMPLATES = {
    "scaffold": ScaffoldGenerator,
    "controller": ControllerGenerator,
    "app": AppGenerator,
    "model": ModelGenerator,
  }

  def self.build_and_generate(template, recipe, options)
    TEMPLATES[template].build(recipe, options).each do |klass|
      klass.render(recipe.directory, list: true, color: true)
    end
  end
end

# src/amber/cli/recipes/recipe.cr
class Recipe
  ...
  def generate(template : String, options = nil)
    puts_info(template, options)
    GeneratorBuilder.build_and_generate(template, self, options)
    install_dependencies(options)
  end

  def puts_info(template, options)
    info "Rendering #{template.capitalize} #{name} in #{directory} from #{options.r}"
  end

  def install_dependencies(options)
    if options.deps?
      info "Installing Dependencies"
      Amber::CLI::Helpers.run("cd #{name} && shards update")
    end
  end
  ....
end
```

eliasjpr: It feels we can do some good cleanup to the CLI generators

faustinoaq: Yeah, I think this can be done in a new PR, WDYT?

faustinoaq: I think this PR is fine as is, this is a requirement to fix this HACK: https://github.com/amberframework/recipes/blob/master/spec/spec_helper.cr#L19

A new refactoring can be done in a new PR :sweat_smile:

faustinoaq: @eliasjpr Can you open an tracking issue about this?

PR #824, opened 2018-05-28
Fix Crecto auth spec

faustinoaq:
### Description of the Change

This PR fixes wrong generated spec by `amber g auth User` when Crecto is used.

This was implemented on #815 by @wontruefree :+1: 

### Alternate Designs

No

### Benefits

Fixes Crecto spec

### Possible Drawbacks

No


wontruefree:
That makes a lot more sense just looking at it.
👍

PR #825, opened 2018-05-28
Update assets

faustinoaq:
### Description of the Change

This PR does:

1. Update main `package.json` depedencies
2. Recompile `amber.min.js` with latest dependencies
3. Update dependencies on `package.json.ecr` template
4. Recompile default CSS && JS bundles provided with `amber new` and latest dependencies

### Alternate Designs

No, this change is required by upcoming `amber watch` configuration

### Benefits

Fixes most NPM package vulnerabilities and prepare asset stuff for next PRs

### Possible Drawbacks

No


robacarp:
Since .min.js and .bundle.js files are essentially compiled, why are we tracking those in the amber repo?

faustinoaq:
> Since .min.js and .bundle.js files are essentially compiled, why are we tracking those in the amber repo?

You're right this is just an update of NPM/JS assets (to avoid vulnerability alerts on GH). I will remove them in my next PR about new `amber watch` configuration :wink: 

For now I think this is ok :+1:

drujensen:
@robacarp @faustinoaq we leave the compiled assets so that the site will display immediately without requiring npm to finish compiling.  Any reason to remove them?

faustinoaq:
> we leave the compiled assets so that the site will display immediately without requiring npm to finish compiling.


@drujensen Yeah, that's a good reason. In my next PR (new `amber watch` config). I will use a isolated vanilla.js script for amber and hot reloading. so this won't be a problem anymore.

For now, I think this is ok :+1:

robacarp:
@drujensen I’m sorry, I forgot we’ve discussed this before at least once. I’m fine with that. My compulsion is to never commit compiled or transpiled code...but I remember the reason behind this now.

PR #828, opened 2018-05-28
api generator

drujensen:
### Description of the Change

Provide the ability to generate an API endpoint with model and migration.  

### Alternate Designs

none

### Benefits

Easily scaffold an API for quick POC's

### Possible Drawbacks

This is a working version but there are several improvements needed IMO:
1. We need to be able to support `amber g api Api::Post title:string` to provide namespaces.
2. We need to provide a simpler `responds_with` option for `json`
3. The specs need some improvement to check the DB for actual changes.

#### Setup
```
amber new blog --deps
cd blog
amber g api post title:string
amber db drop create migrate
amber w
```

#### Index
```
curl \
  -i 'http://localhost:3000/posts' \
  -X GET \
  -H "Content-Type:application/json" \
  -H 'Accept: application/json' \
  -H "Origin:localhost"
```

#### Show
```
curl \
  -i 'http://localhost:3000/posts/1' \
  -X GET \
  -H "Content-Type:application/json" \
  -H 'Accept: application/json' \
  -H "Origin:localhost"
```

#### Create
```
curl \
  -i 'http://localhost:3000/posts' \
  -X POST \
  -H "Content-Type:application/json" \
  -H 'Accept: application/json' \
  -H "Origin:localhost" \
  -d '{"title": "New Blog Post"}'
```

#### Update
```
curl \
  -i 'http://localhost:3000/posts/1' \
  -X PATCH \
  -H "Content-Type:application/json" \
  -H 'Accept: application/json' \
  -H "Origin:localhost" \
  -d '{"title": "Old Blog Post"}'
```

#### Destroy
```
curl \
  -i 'http://localhost:3000/posts/1' \
  -X DELETE \
  -H "Content-Type:application/json" \
  -H 'Accept: application/json' \
  -H "Origin:localhost"
```

### DIFF
--- a/src/amber/cli/templates/api/controller.cr
+++ b/src/amber/cli/templates/api/controller.cr
@@ -23,7 +23,7 @@ module Amber::CLI::Api
       field_hash
 
       add_routes :api, <<-ROUTE
-        resources "/#{Inflector.pluralize(@name)}", #{class_name}Controller
+        resources "/#{Inflector.pluralize(@name)}", #{class_name}Controller, except: [:new, :edit]
       ROUTE
     end
 
--- a/src/amber/cli/templates/api/controller.cr
+++ b/src/amber/cli/templates/api/controller.cr
@@ -1,0 +1,62 @@
+require "../field.cr"
+require "inflector"
+
+module Amber::CLI::Api
+  class Controller < Teeplate::FileTree
+    include Amber::CLI::Helpers
+
+    @name : String
+    @fields : Array(Field)
+    @visible_fields : Array(String)
+    @database : String
+    @language : String
+    @fields_hash = {} of String => String
+
+    def initialize(@name, fields)
+      @language = CLI.config.language
+      @database = CLI.config.database
+      @fields = fields.map { |field| Field.new(field, database: @database) }
+      @fields += %w(created_at:time updated_at:time).map do |f|
+        Field.new(f, hidden: true, database: @database)
+      end
+      @visible_fields = visible_fields
+      field_hash
+
+      add_routes :api, <<-ROUTE
+        resources "/#{Inflector.pluralize(@name)}", #{class_name}Controller
+      ROUTE
+    end
+
+    def field_hash
+      @fields.each do |f|
+        if !%w(created_at updated_at).includes?(f.name)
+          field_name = f.reference? ? "#{f.name}_id" : f.name
+          @fields_hash[field_name] = default_value(f.cr_type) unless f.nil?
+        end
+      end
+    end
+
+    def visible_fields
+      @fields.reject { |f| f.hidden }.map do |f|
+        f.reference? ? "#{f.name}_id" : f.name
+      end
+    end
+
+    private def default_value(field_type)
+      case field_type.downcase
+      when "int32", "int64", "integer"
+        "1"
+      when "float32", "float64", "float"
+        "1.00"
+      when "bool", "boolean"
+        "true"
+      when "time", "timestamp"
+        Time.now.to_s
+      when "ref", "reference", "references"
+        rand(100).to_s
+      else
+        "Fake"
+      end
+    end
+  end
+end
--- a/src/amber/cli/templates/api/controller/crecto/spec/controllers/{{name}}_controller_spec.cr.ecr
+++ b/src/amber/cli/templates/api/controller/crecto/spec/controllers/{{name}}_controller_spec.cr.ecr
@@ -26,10 +26,9 @@ class <%= class_name %>ControllerTest < GarnetSpec::Controller::Test
 
   def initialize
     @handler = Amber::Pipe::Pipeline.new
-    @handler.build :web do
+    @handler.build :api do
       plug Amber::Pipe::Error.new
       plug Amber::Pipe::Session.new
-      plug Amber::Pipe::Flash.new
     end
     @handler.prepare_pipelines
   end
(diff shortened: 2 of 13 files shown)


damianham:
I would have thought recipes were the best way to provide an API application template.

faustinoaq:
> I would have thought recipes were the best way to provide an API application template.

Yeah, I have been contributing and testing recipes. An API recipe would full fill this use-case in a very nice way, although, granite and crecto templates are already here, so maybe we can have a `basic/api` recipe or even just `api` recipe and keep this. @damianham WDYT?

PS: Let me try to create an API recipe based on this PR :wink: Just like I did with Jennifer :+1: 

BTW, recipes are easier to test (all use-cases combinations are being tested, see: [recipes/specs](https://github.com/amberframework/recipes/tree/master/spec) I found and fixed a lot of bugs with this), and auto-deployed, so maybe in the future we can remove all amber embedded templates and generators and just use recipes :sweat_smile: 

Also, I just realize amber templates already have a lot of duplicated code, recipes can solve that by using  hierarchical recipes, see: https://github.com/amberframework/recipes/issues/16

drujensen:
@damianham @faustinoaq I think you guys are thinking of a different use case than I am.

I would like to generate an API and a WEB interface in the same application.  Amber is well suited for this using pipelines.   I am not looking for a new application that is API only.  Does that make sense?

drujensen:
@faustinoaq If you want to work on an API only application, then I think building a recipe is fine but that will not help my use case.  I would still want this API generator.

damianham:
@drujensen a recipe is still the best way to do that.  The react/preact_redux recipe does exactly that (if you remove the React JS stuff) although it doesn't use different pipelines it uses respond_with to respond with either html or json.  If you are creating a new api generator in the CLI because you want to be able to add routes to 2 pipelines during scaffolding - modules and routes defiend in modules are the way to solve that problem :-)
@faustinoaq I think an api recipe at the root rather than basic/api.

drujensen:
>> by creating an internal api generator you are just duplicating the scaffold generator but with JSON controller responses and without the views.

It's a little more than that since we will be introducing an `api` pipeline as well.  Also the routes `resources` will need to be changed to remove the `new` and `edit` routes.

Eventually, we will be moving the recipes into their own repos, so your example:
`amber new my_api_app -r api` will be `amber new my_api_app -r damianahm/api` but close enough.

drujensen:
>> Also I see this is limited to granite or crecto, thus limiting support for other ORMs

This is true and something the core team has been discussing.  We might remove the `-m` flag and use recipes to handle different ORM's.  It's still something we have been debating.

faustinoaq:
> sorry @faustinoaq after all your excellent work with recipes

No problem, Thank you for your new proposal :smile: :+1:

src/amber/cli/templates/api/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -0,0 +1,73 @@
...
+    end
+  end
+
+  def create
+    <%= @name %> = <%= class_name %>.new(<%= @name %>_params.validate!)
+
+    if <%= @name %>.valid? && <%= @name %>.save
+      responds_with 201 do
+        json <%= @name %>.to_json
+      end
```
robacarp: These `responds_with` blocks seem verbose given that it's always going to respond with json, is that the only way to render json with Amber?

robacarp: Actually, shouldn't these be `respond_with` (singular)?

drujensen: still a WIP but yeah.

src/amber/cli/templates/api/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
```diff
@@ -0,0 +1,84 @@
...
+    response = subject.get "/<%= Inflector.pluralize(@name) %>"
+
+    response.status_code.should eq(200)
+    response.body.should contain("Fake")
+  end
+
+  it "renders <%= @name %> show json" do
+    <%= class_name %>.clear
+    model = create_<%= @name.downcase %>
+    location = "/<%= Inflector.pluralize(@name) %>/#{model.id}"
```
robacarp: I see this snippet everywhere. Does it make sense to put the plural in the teeplate controller above?

drujensen: @robacarp Yes, yes it does.  👍

robacarp:
@elorest that seems like it'd be a handy utility, but I think it's outside the scope of this generator pull

src/amber/cli/templates/api/controller.cr
```diff
@@ -0,0 +1,62 @@
...
+      when "float32", "float64", "float"
+        "1.00"
+      when "bool", "boolean"
+        "true"
+      when "time", "timestamp"
+        Time.now.to_s
+      when "ref", "reference", "references"
+        rand(100).to_s
+      else
+        "Fake"
```
eliasjpr: Why not just an empty string here?

src/amber/cli/templates/api/controller/crecto/spec/controllers/{{name}}_controller_spec.cr.ecr
```diff
@@ -0,0 +1,84 @@
...
+  end
+end
+
+describe <%= class_name %>ControllerTest do
+  subject = <%= class_name %>ControllerTest.new
+
+  it "renders <%= @name %> index json" do
+    <%= class_name %>.clear
+    model = create_<%= @name.downcase %>
+    response = subject.get "/<%= Inflector.pluralize(@name) %>"
```
eliasjpr: The inflector is used a lot here. Maybe we can find a way to put this `Inflector.pluralize(@name)` in a higher order and just pass a variable to the template? 

```crystal
resource_path = "/#{Inflector.pluralize(@name)}"
```

elorest:
> @elorest that seems like it'd be a handy utility, but I think it's outside the scope of this generator pull

@robacarp I'm fine with merging without that option. Just pointing it out as rails scaffolding allows you to skip models and also asks if we should overwrite existing files.

PR #832, opened 2018-05-31
fix CORS issue when multiple origins

arambert:
### Description of the Change

Memoized @request_origin variable was being used across multiple requests, this PR forces a refresh at each request.

### Benefits

This allows using Amber with Pipe::CORS for an API receiving requests from multiple origins.
For instance:

* if a first request comes with a 'sample.com' origin, it will receive an answer with 'Access-Control-Allow-Origin' set to sample.com
* if a second request comes with a 'example.com' origin, it will now receive an answer with 'Access-Control-Allow-Origin' set to example.com (and not sample.com like it was the case before this PR)

### Possible Drawbacks

Maybe a very very small loss in performance because `@request_origin = request.headers[Headers::ORIGIN]? || request.headers[Headers::X_ORIGIN]?` will be computed at each request instead for simply returning @request_origin when set.


arambert:
thanks to you all 👍

Issue #817, opened 2018-05-26
Generate specs for Crecto


faustinoaq:
### Description

Currently `amber new app -m crecto` doesn't generate specs

You can see there isn't spec folder on https://github.com/amberframework/amber/tree/master/src/amber/cli/templates/model/crecto/src/models 

### Steps to Reproduce

1. `amber new app -m crecto`
2. `crystal spec`

**Expected behavior:** Execute controllers and models specs based on crecto

**Actual behavior:** Error fixed by #816 but still no specs

**Reproduces how often:** 100%

### Versions

amber master (https://github.com/amberframework/amber/commit/bf55c441f862ba0603b4a4a9023d3e8421e35358)

### Additional Information

@damianham already added this missing specs to basic crecto recipe, see: https://github.com/amberframework/recipes/tree/master/basic/crecto/model/spec/models (it has some bugs, I already working on some fixes :wink: )

eliasjpr:
Fixed #837

PR #837, opened 2018-06-04
Fix crecto controller spec

mofumofu3n:
### Description of the Change

This PR fixes wrong generated spec by `amber g scaffold` when Crecto is used.

### Alternate Designs

No

### Benefits

Fixes Crecto controller spec.

### Possible Drawbacks

No

### DIFF
--- a/src/amber/cli/templates/scaffold/controller/crecto/spec/controllers/{{name}}_controller_spec.cr.ecr
+++ b/src/amber/cli/templates/scaffold/controller/crecto/spec/controllers/{{name}}_controller_spec.cr.ecr
@@ -36,10 +36,14 @@ class <%= class_name %>ControllerTest < GarnetSpec::Controller::Test
 end
 
 describe <%= class_name %>ControllerTest do
+
+  Spec.before_each do
+    Repo.delete_all(<%= class_name %>)
+  end
+
   subject = <%= class_name %>ControllerTest.new
 
   it "renders <%= @name %> index template" do
-    <%= class_name %>.clear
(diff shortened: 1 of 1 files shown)


faustinoaq:
@mofumofu3n Nice, Thank you!

I was working on this. Well, I guess you were faster :sweat_smile: :+1:

mofumofu3n:
You were creating the related issue! (#817)
I think it will fix after this issue merged.

src/amber/cli/templates/scaffold/controller/crecto/spec/controllers/{{name}}_controller_spec.cr.ecr
```diff
@@ -36,18 +36,21 @@ class <%= class_name %>ControllerTest < GarnetSpec::Controller::Test
 end
 
 describe <%= class_name %>ControllerTest do
+
+  Spec.before_each do
+    Repo.delete_all(<%= class_name %>)
+  end
+
```
robacarp: This is probably not going to do what you expect. 

> Instructs the spec runner to execute the given block before each spec, regardless of where this method is invoked.
https://crystal-lang.org/api/0.24.2/Spec.html#before_each%28%26block%29-class-method

Due to this weakness with how Spec#before_each is implemented, it should be avoided in generated code at almost all costs.

mofumofu3n: I understand this weakness!

robacarp:
@mofumofu3n thanks!

PR #838, opened 2018-06-07
fix controller generator

drujensen:
### Description of the Change

The controller generator was improperly naming the views.

### Alternate Designs

none

### Benefits

fixes the generator

### Possible Drawbacks

none

### DIFF
--- a/src/amber/cli/templates/controller.cr
+++ b/src/amber/cli/templates/controller.cr
@@ -30,7 +30,7 @@ module Amber::CLI
     def add_views
       @actions.each do |action|
         FileUtils.mkdir_p("src/views/#{@name}")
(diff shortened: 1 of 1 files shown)


faustinoaq:
Hi @drujensen I already did a similar fix for recipes controllers here:https://github.com/amberframework/amber/pull/823

I think we can use a similar approach here :wink:

src/amber/cli/templates/controller.cr
```diff
@@ -30,7 +30,7 @@ module Amber::CLI
     def add_views
       @actions.each do |action|
```
faustinoaq: This can be fixed by using `@actions.each do |action, _|` (notice the `_`)

drujensen:
@faustinoaq pr change done.  please review. thx.

PR #839, opened 2018-06-08
Update Benchmarks

faustinoaq:
### Description of the Change

This PR updates Amber Benchmark section with latest TFB round 16 using same settings:

- Default front-page test screenshot  (Fortunes)
- Comparing "inspired by" frameworks (Hanami, Kemal, Rails, Phoenix)

Amber is still the fastest in fortunes tests, also, Amber has increased performance (almost 5x) in plaintext and json tests :rocket: 

However, crystal in general is still very slow on db query tests :sweat_smile: 

### Alternate Designs

- Add more test screenshots (json, plaintext, etc) or
- Remove benchmarks (well, I think benchmarks are nice to compare and improve) or
- Replace TFB by other benchmarks (maybe using `wrk` or similar, although, IMO TFB are very nice)

### Benefits

Use latest TFB benchmark to show updated results (Still the fastest in our category [ruby-like syntax])

### Possible Drawbacks

In this round amber got some errors, maybe because high concurrency issues, although, seems performance wasn't affected at all :sweat_smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,11 @@ Read Amber documentation on https://docs.amberframework.org
 
 ## Benchmarks
 
-[Techempower Framework Benchmarks - Round 15 (2018-02-14)](https://www.techempower.com/benchmarks/#section=data-r15&hw=ph&test=fortune&f=zik073-zik0zj-zik0zj-zik0zj-zhxjwf-zik0zj-gnbmym-cn3)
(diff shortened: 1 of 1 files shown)


faustinoaq:
Latest rounf 15 here: https://www.techempower.com/benchmarks

PR #850, opened 2018-06-16
Add more build specs

faustinoaq:
### Description of the Change

This PR add more build specs like [recipes repo already does](https://github.com/amberframework/recipes/tree/master/spec):

| model/template | ecr | slang |
| --- | --- | --- |
| granite | :+1:  | :+1:  |
| crecto | :+1:  | :+1:  |

### Alternate Designs

Add more built specs (mysql, sqlite)?

### Benefits

This will help us to prevent issues with projects ecr and crecto because this combination is currently untested. We had [several issues in the past](https://github.com/orgs/amberframework/projects/1?fullscreen=true&card_filter_query=crecto+is%3Aclosed) with crecto/ecr mainly because no build specs on them

### Possible Drawbacks

This will check all amber features are working fine (well, except other db :sweat_smile: )

### DIFF
--- a/spec/build_spec_helper.cr
+++ b/spec/build_spec_helper.cr
@@ -1,0 +1,71 @@
+require "./spec_helper"
+require "./support/helpers/cli_helper"
+
+include CLIHelper
+
+macro generate_app(*options)
+  ENV["AMBER_ENV"] = "test"
+
+  cleanup
+  scaffold_app(TESTING_APP, *{{options}})
+
+  options = ["user:reference", "name:string", "body:text", "age:integer", "published:bool"]
+  temp_options = options - ["user:reference", "age:integer"]
+  MainCommand.run ["generate", "auth", "User"] | (options - ["user:reference"])
+  MainCommand.run ["generate", "error"]
+  MainCommand.run ["generate", "scaffold", "Animal"] | temp_options
+  MainCommand.run ["generate", "scaffold", "Post"] | options
+  MainCommand.run ["generate", "scaffold", "PostComment"] | (options + ["post:reference"])
+  MainCommand.run ["generate", "model", "Bat"] | options
+  MainCommand.run ["generate", "migration", "Crocodile"] | options
+  MainCommand.run ["generate", "mailer", "Dinosaur"] | options
+  MainCommand.run ["generate", "socket", "Eagle"] | ["soar", "nest"]
+  MainCommand.run ["generate", "channel", "Falcon"]
+
+  prepare_yaml(Dir.current)
+  Amber::CLI.env = "test"
+  Amber::CLI.settings.logger = Amber::Environment::Logger.new(nil)
+
+  puts "RUNNING: amber db drop create migrate - started..."
+  MainCommand.run ["db", "drop", "create", "migrate"]
+
+  puts "RUNNING: shards update - started..."
+  system("shards update")
+
+  puts "RUNNING: shards build #{TEST_APP_NAME} - started..."
+  system("shards build #{TEST_APP_NAME}")
+end
+
+macro check_formatting
+  it "check formatting on generated files" do
+    system("crystal tool format --check src").should be_true
+  end
+end
+
+macro check_binary
+  it "generates a binary" do
+    File.exists?("bin/#{TEST_APP_NAME}").should be_true
+  end
+end
+
+macro check_app_specs
+  context "crystal spec" do
+    puts "RUNNING: crystal spec #{TESTING_APP} - started..."
+
+    spec_result = `crystal spec`
+
+    puts spec_result
+
+    it "can be executed" do
+      spec_result.should contain "Finished in"
+    end
+
+    it "has no errors" do
+      spec_result.should_not contain "Error in line"
+    end
+
+    it "has no failures" do
+      spec_result.should_not contain "Failures"
+    end
+  end
+end
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,8 @@ cache:
     - .shards
 
 env:
-  - TEST_SUITE=./spec/build_spec.cr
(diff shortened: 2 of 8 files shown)


eliasjpr:
It would be great if we can move all the templates to recipes, not sure if thats possible, but this will removed the overhead of testing the generated apps and have those tests in one central place. 

Not sure how feasible is this since I know there  are concerns about pulling generators from the recipe repos.

eliasjpr:
@faustinoaq can we move these tests to the Recipe repo? Even if we duplicate the templates for now. 

@damianham do you think you can work on making a default recipe for amber. @amberframework/contributors lets also brainstorm how we can move the API generator (if possible)

Ideally the CLI can be a shell that only runs command without the template dependencies. ❤️

faustinoaq:
> can we move these tests to the Recipe repo? Even if we duplicate the templates for now.

All these test (an more) already exist on recipe repo, see: https://github.com/amberframework/recipes/tree/master/spec

@eliasjpr We already discussed recipes on #750 and we concluded we gonna try current recipe implementation (optional), and migrate templates to recipes or shards progressively.

Currently, This PR is very useful to cover untested features like crecto and ECR templates for amber projects generated without recipes.

damianham:
@eliasjpr my current thinking based on the work I did on my Maze test framework is to have a base (default) recipe as a shard in the organisation so it would be at https://github.com/amberframework/base

I haven't answered all the questions so it is still a work in progress but I think it is moving in the right direction.  Going forward we might need to remove the ORM option from the CLI so that the base recipe is Granite only with other ORMs as different recipes.

faustinoaq:
@eliasjpr @elorest This PR is passing now! :tada:

faustinoaq:
Thanks to this PR I found several issues on models, formatting, etc :+1:

spec/build_spec_helper.cr
```diff
@@ -0,0 +1,77 @@
+require "./spec_helper"
+require "./support/helpers/cli_helper"
+
+include CLIHelper
+
+macro generate_app(*options)
```
eliasjpr: why this has to be a macro?

spec/build_spec_helper.cr
```diff
@@ -0,0 +1,77 @@
...
+  end
+end
+
+macro check_app_specs
+  context "crystal spec" do
+    puts "RUNNING: crystal spec #{TESTING_APP} - started..."
+
+    spec_result = `crystal spec`
+
+    puts spec_result
```
eliasjpr: Do we want to output here? why not output on failures only?

spec/build_spec_helper.cr
```diff
@@ -0,0 +1,77 @@
...
+
+    it "can be executed" do
```
eliasjpr: it compiles project?

spec/build_spec_helper.cr
faustinoaq: @eliasjpr This was a fix introduced on #820 

`puts spec_result` on success will print insignificant output

spec/build_spec_helper.cr
faustinoaq: @eliasjpr I did this to preserve every spec scope (actually this just avoid boilerplate code :sweat_smile: )

spec/build_spec_helper.cr
faustinoaq: @eliasjpr Actually this spec was introduced by you [here](https://github.com/amberframework/amber/blame/f8be9e157774cd9475ca26953da6c14beed1623c/spec/build_spec.cr) :sweat_smile:  

Maybe we should simply use `system("crystal spec").should be_true` WDYT?

spec/build_spec_helper.cr
eliasjpr: I think this can be done without a macro, I will give it a try. The specs should run independently

spec/build_spec_helper.cr
eliasjpr: what I mean to say is to rename it to `it "compiles amber project"`

eliasjpr: I think thats fair, and since we already doing that we can delete this spec completely

eliasjpr: This is a duplicate expectation we can delete this spec

eliasjpr: I like this a lot `system("crystal spec").should be_true`

spec/build_spec_helper.cr
```diff
@@ -0,0 +1,77 @@
...
+      spec_result.should contain "Finished in"
+    end
+
+    it "has no errors" do
+      spec_result.should_not contain "Error in line"
+    end
+
+    it "has no failures" do
```
eliasjpr: We can delete this spec

spec/build_spec_helper.cr
```diff
@@ -0,0 +1,77 @@
...
+
+    spec_result = `crystal spec`
+
+    puts spec_result
+
+    it "can be executed" do
+      spec_result.should contain "Finished in"
+    end
+
+    it "has no errors" do
```
eliasjpr: we can delete this spec if you add `system("crystal spec").should be_true`

spec/build_spec_helper.cr
faustinoaq: I removed macros :+1:

Also the `describe` are running their specs `it` independently :tada:

spec/build_spec_helper.cr
faustinoaq: Done! :+1:

spec/build_spec_helper.cr
faustinoaq: Done! :+1:

spec/build_spec_helper.cr
faustinoaq: Done! :+1:

Issue #851, opened 2018-06-16
[CLI] Cleanup Scaffold Generator Granite/Crecto


eliasjpr:
### Description

Clean up scaffold generator using Granite new features

### Example 

```crystal 
class PostController < ApplicationController
  getter post : Post = Post.new

  before_action do
    only [:show, :edit, :update, :destroy] { set_post }
  end

  def index
    posts = Post.all
    render("index.slang")
  end

  def show
    render("show.slang")
  end

  def new
    render("new.slang")
  end

  def edit
    render("edit.slang")
  end

  def create
    post = Post.new(post_params.validate!)

    if post.save
      redirect_to action: :index, flash: { "success" => "Created Post successfully." }
    else
      redirect_to action: :index, flash: { "errors" => @post.errors }
    end
  end

  def update
    @post.set_attributes(post_params.validate!)

    if @post.save
      redirect_to action: :index, flash: { "success" => "Updated Post successfully." }
    else
      redirect_to action: :index, flash: { "errors" => @post.errors }
    end
  end

  def destroy
    @post.destroy
    redirect_to action: :index, flash: { "success" => "Deleted Post successfully." }
  end

  private def post_params
    params.validation do
      required(:title) { |f| !f.nil? }
      required(:body) { |f| !f.nil? }
    end
  end

  private def set_post
    @post = Post.find!(params[:id])
  end
end

```

### Alternatives 

- Render can accept a `flash` param, this will. allow to have even thinner controller

PR #852, opened 2018-06-16
[Cleanup] Cleans Controller Scaffold

eliasjpr:
Issue: https://github.com/amberframework/amber/issues/851

### Description of the Change

Granite adds some nice features that allows controllers to be thinner
and cleaner. With these changes controllers becomes almost on par with
Rails controller approach.

### Benefits

* Removes conditional branches
* Uses before_action filters
* Users Granite `find!`
* Laverages on Granite save to run validations instead of calling
`valid?` and then `save`

### Expected Behavior

- We expect behavior to be the same.

### DIFF
--- a/src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
+++ b/src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
@@ -1,74 +1,61 @@
 class <%= class_name %>Controller < ApplicationController
+  getter <%= @name %> = <%= class_name %>.new
+
+  before_action do
+    only [:show, :edit, :update, :destroy] { set_<%= @name %> }
+  end
+
   def index
     <%= name_plural %> = <%= class_name %>.all
     render("index.<%= @language %>")
   end
 
   def show
-    if <%= @name %> = <%= class_name %>.find params["id"]
-      render("show.<%= @language %>")
-    else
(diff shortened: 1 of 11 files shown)


src/amber/cli/templates/scaffold/controller/granite/src/controllers/{{name}}_controller.cr.ecr
```diff
@@ -1,74 +1,61 @@
...
-
-    if <%= @name %>.valid? && <%= @name %>.save
-      flash["success"] = "Created <%= class_name %> successfully."
-      redirect_to "/<%= Inflector.pluralize(@name) %>"
-    else
-      flash["danger"] = "Could not create <%= class_name %>!"
-      render("new.<%= @language %>")
-    end
+  def edit
+    render("edit.slang")
```
drujensen: might not be `slang`

eliasjpr:
Would be nice to have render take a flash object then we can clean these lines:

```crystal
flash["danger"] = "Could not update <%= class_name %>!"
render("edit.<%= @language %>")
```

faustinoaq:
@eliasjpr Looks like we need to update generated specs here as well :sweat_smile:

faustinoaq:
I just realize this PR depends on #849 

@eliasjpr I changed this code without checking that PR, my bad :sweat_smile:

eliasjpr:
Weird because I push those fixes about the set method today

faustinoaq:
@eliasjpr can you sync and push them again :sweat_smile:

Issue #855, opened 2018-06-16
[TravisCI] Run Travis in Supported Crystal Version


eliasjpr:
### Description

Travis should run in the current support version of Crystal until we add support for Crystal v0.25.0

eliasjpr:
See https://github.com/crystal-lang/crystal/issues/5449 😞

PR #856, opened 2018-06-16
[TravisCI] Use crystal specific version

eliasjpr:
Issue: https://github.com/amberframework/amber/issues/855

Until we can add support for the latest crystal version Travis should
run in the previous supported Crystal version. Currently some specs are failing.

We should expect specs to pass in the supported version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 dist: trusty
 sudo: required
 language: crystal
+crystal:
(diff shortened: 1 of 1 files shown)


eliasjpr:
This does not work since unfortunately Crystal team does not saver older builds of crystal. See https://github.com/crystal-lang/crystal/issues/5449

PR #857, opened 2018-06-16
Update to Crystal/0.25.0

robacarp:
This is an internal branch copy of the branch @bcardiff opened in #826

PR #861, opened 2018-06-16
Use pluralize method

### LINKED ISSUE #713: Refactor Inflector.pluralize
### Description

I know https://github.com/amberframework/amber/pull/673 was already merged, but I think a refactor would be helpful. The code `Inflector.pluralize(@name)` shows up in lots of places. Why not have this in the template:

```crystal
def plural_name
  @plural_name ||= Inflector.pluralize(@name)
end
```

Then `plural_name` can be used everywhere, and if amber ever swaps this inflector for another one (like what @faustinoaq mentioned) then it just has to change in this one place.

Also, the inflector being called that many times might incur some unnecessary overhead.

### Steps to Reproduce

Not applicable

### Versions

Master, see PR: https://github.com/amberframework/amber/pull/673

### Additional Information

This issue was opened to track a requested change in a merged PR

/cc @marksiemers

faustinoaq:
### Description of the Change

This PR simplifies pluralization by using `pluralize` method instead of `Inflector.pluralize`

### Alternate Designs

Keep `Inflector.pluralize`

### Benefits

Less verbose and fixes #713 

### Possible Drawbacks

No

### DIFF
--- a/src/amber/cli/templates/scaffold/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
+++ b/src/amber/cli/templates/scaffold/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
@@ -40,16 +40,16 @@ describe <%= class_name %>ControllerTest do
 
   it "renders <%= @name %> index template" do
     <%= class_name %>.clear
-    response = subject.get "/<%= Inflector.pluralize(@name) %>"
+    response = subject.get "/<%= pluralize(@name) %>"
 
     response.status_code.should eq(200)
-    response.body.should contain("<%= Inflector.pluralize(display_name) %>")
+    response.body.should contain("<%= pluralize(display_name) %>")
   end
 
   it "renders <%= @name %> show template" do
     <%= class_name %>.clear
     model = create_<%= @name.downcase %>
-    location = "/<%= Inflector.pluralize(@name) %>/#{model.id}"
+    location = "/<%= pluralize(@name) %>/#{model.id}"
 
     response = subject.get location
 
@@ -59,7 +59,7 @@ describe <%= class_name %>ControllerTest do
 
   it "renders <%= @name %> new template" do
     <%= class_name %>.clear
-    location = "/<%= Inflector.pluralize(@name) %>/new"
+    location = "/<%= pluralize(@name) %>/new"
 
     response = subject.get location
 
@@ -70,7 +70,7 @@ describe <%= class_name %>ControllerTest do
   it "renders <%= @name %> edit template" do
     <%= class_name %>.clear
     model = <%= create_model_method %>
-    location = "/<%= Inflector.pluralize(@name) %>/#{model.id}/edit"
+    location = "/<%= pluralize(@name) %>/#{model.id}/edit"
 
     response = subject.get location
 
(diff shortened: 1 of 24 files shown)


faustinoaq:
> Why not just create a variable called name_plural since its the only variable that gets pluralized?

@eliasjpr Nice idea :+1:

faustinoaq:
@eliasjpr Done! :tada:

src/amber/cli/templates/scaffold/controller/granite/spec/controllers/{{name}}_controller_spec.cr.ecr
```diff
@@ -40,16 +40,16 @@ describe <%= class_name %>ControllerTest do
 
   it "renders <%= @name %> index template" do
     <%= class_name %>.clear
-    response = subject.get "/<%= Inflector.pluralize(@name) %>"
+    response = subject.get "/<%= @name_plural %>"
 
     response.status_code.should eq(200)
-    response.body.should contain("<%= Inflector.pluralize(display_name) %>")
+    response.body.should contain("<%= pluralize(display_name) %>")
```
robacarp: missed one

faustinoaq: Thank you, Done! :+1:

src/amber/cli/recipes/recipe.cr
```diff
@@ -40,6 +40,7 @@ module Amber::Recipes
         exit 1
       end
 
+      @name_plural = Inflector.pluralize(word)
```
marksiemers: Where does `word` come from? Isn't this pluralizing `@name`?

Also, any reason not to create a getter, or write a custom one like this:
```crystal
def plural_name
  @plural_name ||= Inflector.pluralize(word)
end
```

faustinoaq: Hi @marksiemers :tada: 

You're completely right, let me fix that :100:

faustinoaq: > Also, any reason not to create a getter, or write a custom one like this:

Is a getter really needed? I think `initialize` method already does the job, no?

`@name_plural` is initialized just one time inside the constructor, so no memoization requirement

marksiemers: It can be done without the getter, assuming it will never be needed publicly.

If it will ever be needed publicly, then creating a private getter now will make things more consistent when that change comes.

I know this isn't ruby, but one of the ruby idioms that I follow is that you minimize directly accessing instance variables (ideally only in the constructor/initialize method).

faustinoaq: @marksiemers Yeah, nice suggestion :+1:

I think we can refactor this in a new PR (`@name_plural` and others)  :sweat_smile:

faustinoaq: BTW, nice to see you here again :smile: :+1:

marksiemers:
@faustinoaq - Thanks. I started a new job, moved, and got into rock climbing - so my OSS time has been very limited.

It looks like the CI is currently failing due to Crystal 0.25.0 incompatibilities - I'm assuming that is a bit of a priority - should this PR just wait for those fixes?

Also for this change, if you switch to a getter method, if anything changes in the implementation of that method, you won't have to change 25 files again.

Of course, if the name of the method changes, then we're still stuck with changing all the files.

faustinoaq:
> Thanks. I started a new job, moved, and got into rock climbing - so my OSS time has been very limited.

Hehe, nice, no problem :+1:

> I'm assuming that is a bit of a priority - should this PR just wait for those fixes?

Yeah, almost all current PRs depend on #857 :sweat_smile: 

> Also for this change, if you switch to a getter method, if anything changes in the implementation of that method, you won't have to change 25 files again.

Yeah, you're right, but I think would be nice to change all other ivars as well in a new PR. This would be a bigger refactor. We're using a lot of ivars in all templates :sweat_smile:

faustinoaq:
@marksiemers I guess I was a bit confused about this change (too many files :sweat_smile: )

Your're right, is better to use a `getter` here, I already added it :+1: Thank you!

PR #864, opened 2018-06-16
Added new exception page

faustinoaq:
### Description of the Change

This PR adds a new exception page based on Phoenix one, see: https://github.com/phoenixframework/phoenix/issues/1776

This PR will be required by next `amber watch` PRs

vokoscreen-2018-04-14_18-07-30

### Alternate Designs

Keep basic error pages

### Benefits

- Fixes #770 reported by @c910335
- Add code snippets right in the browser
- Add file navigation
- Add header inspector
- Add Cookies inspector
- Add new awesome design based on Phoenix one with responsive design :sparkles: Thanks to @rstacruz :tada: 

### Possible Drawbacks

[x] Reload (watch) checkbox depends on new `amber watch` configuration (based on `amber_reload.js` without reload pipe)

PR #865, opened 2018-06-16
New amber watch config

### LINKED ISSUE #476: Amber Watch - Configurable

### LINKED ISSUE #507: [CLI] Amber watch starts multiple processes
It appears that when one is running under 'amber watch', every save of a watched file triggers a new build without waiting or canceling eventual ones already running.

So when one saves a couple files in quick succession, multiple builds spin up, taking a lot of RAM and making machine unresponsive till they finish.

faustinoaq:
### Description of the Change

This PR depends on #864 and #857 

This PR adds new `amber watch` config using `.amber.yml` file

vokoscreen-2018-04-14_21-23-04

### Alternate Designs

Keep current `process_runner.cr` and force npm/webpack commands :sweat_smile: 

### Benefits

- Fixes #476 
- Fixes #507 
- Allow to add watch tasks
- Allow to add one-time tasks
- Allow to customize build and run commands
- Allow to customize client files to watch
- Removes `sentry.cr` (dead code)

### Possible Drawbacks

All tasks require `files` and `commands` fields, for one-time tasks (without files) you can use an empty array `files: []`

No cli flags, so, things like `amber watch -w "./config/**/*.cr"` are not supported /cc @damianham 

See: https://github.com/damianham/amber_render_module#usage


faustinoaq:
This PR is based on `new-exception-page` branch, see: #864

src/amber/cli/templates/app/.amber.yml.ecr
```diff
@@ -2,3 +2,27 @@ type: app
...
+  server: # required: the first command for this task is blocking
+    files:
+      - "src/**/*.cr"
+      - "src/**/*.<%= @language %>"
+      - "config/**/*.cr"
+    commands:
+      - "crystal build -o bin/<%= @name %> src/<%= @name %>.cr -p --no-color"
+      - "bin/<%= @name %>"
+
+  client: # required: these files changes trigger browser reloading
```
robacarp: should this be required? I generally do almost no front end code, what if I don't want watch to do anything?

faustinoaq: @robacarp You're right, actually this field can be optional, the current implementation can work without a `client` field

faustinoaq:
This new config works pretty nice and fix many issue, although, Does someone have any idea how to add specs for this? :sweat_smile:

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,217 @@
...
+        end
       end
     end
 
-    private def stop_all_processes
-      log "Terminating app #{project_name}..."
-      @processes.each do |process|
-        process.kill unless process.terminated?
+    private def check_directories
+      Dir.mkdir_p("bin")
```
robacarp: why does this create `bin/`?

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,217 @@
...
+    private def run_build_command(command, commands)
+      check_directories
+      next_server_commands_range = (1...commands.size)
+      info "Building project #{Helpers.settings.name.colorize(:light_cyan)}"
+      spawn do
+        error_io = IO::Memory.new
+        process = Helpers.run(command, error: error_io)
+        PROCESSES << {process, "server"}
+        loop do
+          if process.terminated?
```
robacarp: I think this loop contains far too much logic all in one place. It may help to refactor to use early-exits instead, but I don't know. It's hard to even see what's going on here, but something like this might be possible:

```crystal

loop do
   sleep 1
   next if process.terminated?
   next if error_io.empty?
   # etc
end
```

src/amber/cli/templates/app/.amber.yml.ecr
```diff
@@ -2,3 +2,27 @@ type: app
...
+  server: # required: the first command for this task is blocking
+    files:
+      - "src/**/*.cr"
+      - "src/**/*.<%= @language %>"
+      - "config/**/*.cr"
+    commands:
+      - "crystal build -o bin/<%= @name %> src/<%= @name %>.cr -p --no-color"
+      - "bin/<%= @name %>"
+
+  client: # optional: these files changes trigger browser reloading
```
eliasjpr: \# Optional: File changes triggers the browser to reload.

src/amber/cli/templates/app/.amber.yml.ecr
```diff
@@ -2,3 +2,27 @@ type: app
...
+    files:
+      - "public/**/*"
+    commands: []
+
+  webpack: # optional: compiles assets using webpack
```
eliasjpr: I thought the new Amber watch was going to decouple webpack meaning you can choose which ever front end builder you want? Or is this a simple task?

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,217 @@
...
+        context.response.headers["Client-Reload"] = [error_id]
+        context.response.print Amber::Exceptions::ExceptionPageServer.new(context, error_output, error_id).to_s
+      end
     end
 
-    private def get_timestamp(file : String)
-      File.info(file).modification_time.to_s("%Y%m%d%H%M%S")
+    private def handle_error(error_output)
+      kill_processes("server")
+      puts error_output
```
eliasjpr: User `CLI::Logger`

src/amber/cli/templates/app/.amber.yml.ecr
```diff
@@ -2,3 +2,27 @@ type: app
 database: <%= @database %>
 language: <%= @language %>
 model: <%= @model %>
+watch:
+  server: # required: the first command for this task is blocking
```
eliasjpr: What does this mean `required: the first command for this task is blockin`?

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,217 @@
...
     end
 
     def run
-      loop do
-        scan_files
-        sleep 1
+      if watch_object = CLI.config.watch
+        run_watcher(watch_object)
+      else
+        error "Can't find watch settings, please check your .amber.yml file"
```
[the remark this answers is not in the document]
faustinoaq: screenshot_20180621_133658

Yeah,  nice idea, I'm already working on that :+1:

src/amber/cli/templates/app/.amber.yml.ecr
faustinoaq: This is just a simple task, Webpack removal is addressed here: #866

src/amber/cli/templates/app/.amber.yml.ecr
faustinoaq: That means that other tasks are blocked until your amber app compiles

Maybe we should change the message, WDYT? :sweat_smile:

src/amber/cli/helpers/process_runner.cr
faustinoaq: Because bin is needed to store the generated binary

Useful when your tasks is like `crystal build src/myapp.cr -o bin/myapp`, so you don't need to execute `mkdir -p` first. `amber watch` already does this, just like `shards build`

src/amber/cli/helpers/process_runner.cr
faustinoaq: I added a new method `handle_terminaded_process`:

Now the `loop` looks like this:

```cr
        loop do
          if process.terminated?
            handle_terminaded_process(process, error_io)
            break
          end
          sleep 1
        end
```

src/amber/cli/templates/app/.amber.yml.ecr
faustinoaq: Done! :+1:

faustinoaq:
> There are a lot of special cases in here for the server and client watcher, but it seems like that shouldn't be necessary. The one switch that I can see which needs to be made is simply: run a command on the server, or send a message down a channel. What about changing the way the configuration is written to be more generic?

Nice idea, let me try :+1:

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,217 @@
...
         end
+        file_counter += 1
       end
       if file_counter > 0
-        log "Watching #{file_counter} files (server reload)..."
-        start_app
+        debug "Watching #{file_counter} #{key} files"
+        kill_processes(key)
+        commands.each do |command|
+          if key == "server" && command == commands.first?
```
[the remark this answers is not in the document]
faustinoaq: This means that other tasks are blocked until your amber app compiles,

Should we change that behaivior? :sweat_smile: 

I think your idea of running task on parallel would be nice :+1:

faustinoaq:
@robacarp I have a comment about your idea:

```yaml
watch:
    meaningless_name:
      files:
        - src/**/*.cr
        - # etc
      notify_client: true   # This is already done by "client" field
      commands:
         - shards build app # Should we run this on parallel? what about next command
         - bin/app               # This requires app build first, WDYT?
```

faustinoaq:
```
      notify_client: true   # This is already done by "client" field
```

I got it!, I think I can implement this by appending all fields (`meaningless_name`) files with `notify_client: true` to the client watcher :+1:

faustinoaq:
```yml
      commands:
         - shards build app # Should we run this on parallel? what about next command
         - bin/app               # This requires app build first, WDYT?
```

If we use parallel task, then I guess we should require something like this:

```yaml
      commands:
         - shards build app && bin/app
```

WDYT?

eliasjpr:
@faustinoaq I would say not to mess with the feature too much, and iterate i will open issues with additional thoughts so we can merge this

faustinoaq:
@eliasjpr Yeah, no problem, now v0.8.0 is already released by @robacarp , so I'll fix this feature for v0.8.1 :+1: 

I just realize @robacarp idea is simpler and nicer :sweat_smile:

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
-    private def scan_files
+    private def run_watcher(watch_config)
+      begin
+        server_config = watch_config["server"]
+        spawn watcher("server", server_config["files"], server_config["commands"])
+      rescue ex
+        error "Error in watch configuration. #{ex.message}"
+        exit 1
+      end
+      watch_config.each do |key, value|
```
eliasjpr: This should be its own method and be able to loop to all config keys including `watch_config["server"]`

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
+    rescue ex : KeyError
+      error "Error in watch configuration. #{ex.message}"
+      exit 1
     end
 
-    private def scan_files
+    private def run_watcher(watch_config)
+      begin
+        server_config = watch_config["server"]
+        spawn watcher("server", server_config["files"], server_config["commands"])
```
eliasjpr: Why is server is being spawn separately and not just loop over each key as you doing below? Does the order matter? if yes, this should be moved to its own method `run_server_watcher`

The `watch_config` should be an instance variable!

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
-      build_result = build_app_process
-      if build_result.is_a? Process::Status
-        if build_result.success?
-          stop_all_processes
-          create_all_processes
-          @app_running = true
-        elsif !@app_running
-          log "Compile time errors detected. Shutting down..."
-          exit 1
+      if watch_config = CLI.config.watch
```
eliasjpr: This should be moved to initializer as a guard clause. watch
```crystal
def initialize
  raise AMBER_YAML_ERROR unless CLI.config.amber_yml_exists?
  @amber_yaml =  CLI.config.amber_yaml
end
```

Actually the loading and check of this should be in Amber::CLI module and not here. 

And have an instance of ProcessRunner in Amber::CLI 

```crystal
module Amber::CLI
  def self.process_runner
    ProcessRunner.new if amber_yml_loaded?
  end

  def amber_yaml_loaded?
     raise AMBER_YAML_ERROR unless CLI.config.amber_yml_exists?
  end
end

```

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
+        commands = value["commands"]
+        @notify_counter += 1
+        spawn watcher(key, files, commands)
+      end
+      @notify_counter_channel.send @notify_counter
+      @notify_counter = @notify_counter_channel.receive
+      sleep
+    end
+
+    private def watcher(key, files, commands)
```
eliasjpr: Why I keep seeing a lot of `watcher` keywords, this is a ProcessRunner class it knows too much about `Watcher` :disappointed:

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
+      else
+        loop do
+          scan_files(key, files, commands)
+          @watch_running = true if key == "server"
+          sleep 1
+        end
+      end
+    end
+
+    private def scan_files(key, files, commands)
```
eliasjpr: This scan_files should be its own class/struct

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
-      processes.clear
     end
 
-    private def create_all_processes
-      process = create_watch_process
-      @processes << process if process.is_a? Process
-      unless @npm_process
-        create_npm_process
-        @npm_process = true
+    private def run_build_command(command, commands)
```
eliasjpr: Should be its own class/stuct

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
+          end
+          sleep 1
+        end
       end
     end
 
-    private def build_app_process
-      log "Building project #{project_name}..."
-      Amber::CLI::Helpers.run(@build_command)
+    private def handle_terminaded_process(process, error_io)
```
eliasjpr: Why not a TerminatedProcessHandler classs?

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
+      PROCESSES.each do |process, owner|
+        if process.terminated?
+          PROCESSES.delete(process)
+        elsif owner == key || key.nil?
+          process.kill
+        end
+      end
+    end
+
+    private def error_server(error_output)
```
eliasjpr: The body of this method should be moved to a ErrorServer class/module

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
+        context.response.content_type = "text/html"
+        context.response.status_code = 500
+        context.response.headers["Client-Reload"] = [error_id]
+        context.response.print Amber::Exceptions::ExceptionPageServer.new(context, error_output, error_id).to_s
+      end
+    end
+
+    private def handle_error(error_output)
+      kill_processes("server")
+      puts error_output
```
eliasjpr: CLI::Logger?

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
+    @file_watcher : FileWatcher
+
+    def initialize
+      @watch_running = false
+      @wait_build = Channel(Bool).new
+      @server_files_changed = false
+      @notify_counter = 0
+      @notify_counter_channel = Channel(Int32).new
+      @notify_channel = Channel(Nil).new
+      @host = Helpers.settings.host
```
eliasjpr: How come we are getting settings information from a Helper module and not Amber::CLI environment settings?

src/amber/cli/helpers/process_runner.cr
```diff
@@ -1,114 +1,233 @@
...
+      notify_counter.times { @notify_channel.send nil }
+      @notify_counter_channel.send 0
     end
 
-    private def create_npm_process
-      node_log "Installing dependencies..."
-      Amber::CLI::Helpers.run("npm install --loglevel=error && npm run watch", wait: false)
-      node_log "Watching public directory"
+    private def run_command(command, key)
+      if key == "server"
```
eliasjpr: Why is "server" special?

faustinoaq: Yeah, I did such a complex stuff here, I'm already fixing this :+1:

src/amber/cli/helpers/process_runner.cr
faustinoaq: Oh, this is using `include Enviroment`, I know this is a bad practice, because even I was confused about this when I tried to figure out where `Helpers.settings` is defined, So, I gonna use `Amber::CLI` :+1:

src/amber/cli/helpers/process_runner.cr
faustinoaq: Oh, I forgot this one :sweat_smile:

src/amber/cli/helpers/process_runner.cr
faustinoaq: I agree, let me split this file :+1:

src/amber/cli/helpers/process_runner.cr
faustinoaq: I think this useless now, I gonna use parallel processes (like @robacarp  suggested), no channel nor concurrency management would be required :+1:

src/amber/cli/helpers/process_runner.cr
faustinoaq: Nice suggestion, let me apply this refactoring as well :+1:

src/amber/cli/helpers/process_runner.cr
faustinoaq: Yeah, let me split this file :+1:

src/amber/cli/helpers/process_runner.cr
faustinoaq: I'm already using `FileWatcher` class to store file watcher logic, I guess I can create one class for this as well :+1:

src/amber/cli/helpers/process_runner.cr
faustinoaq: This won't be required, I just gonna use `Helpers.run` to run processes :+1:

src/amber/cli/commands/database.cr
```diff
@@ -55,7 +55,7 @@ module Amber::CLI
           when "create"
             Micrate.logger.info create_database
           when "seed"
-            Helpers.run("crystal db/seeds.cr", wait: true, shell: true)
+            Helpers.run("crystal db/seeds.cr", shell: true).wait
```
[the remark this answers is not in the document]
faustinoaq: Looks cleaner, Nice idea! :+1:

src/amber/cli/config.cr
```diff
@@ -1,28 +1,57 @@
...
+          "config/**/*.cr"
+        ],
+        "commands" => [
+          "shards build -p --no-color",
+          "bin/#{app_name}"
+        ]
+      }
+    }
 
     def initialize
```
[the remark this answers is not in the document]
faustinoaq: If I remove the initialize then I can't use `Config.new` because it complains about no matching parameters :sweat_smile: 

Let me add a comment about this :+1:

src/amber/cli/helpers/process_runner.cr
faustinoaq: I applied @robacarp suggestion here, althoght, the new refactoring won't require this :sweat_smile:

src/amber/cli/helpers/process_runner.cr
faustinoaq: I applied @robacarp suggestion here, although, the new refactoring won't require this sweat_smile

elorest:
Why was this closed? Looks a little complex, but there's some good stuff in here. 

@faustinoaq Will there be another PR with some of this? Also will autoreload be added back soon?

robacarp:
@elorest @drujensen This was closed for cleanup -- @faustinoaq did some great work, but doesn't have time these days to contribute as much. If someone wants to pick up where he left of, 👍

Issue #834, opened 2018-06-01
Github flagging hapijs for vulnerability


rishavs:
We need to update the package.json version for hapijs to escape this notification.

![image](https://user-images.githubusercontent.com/1210239/40838444-d2bdae88-65bb-11e8-86d8-dc0ad05bc656.png)

This is in the Amber CLI (amberframework.org) - v0.7.2

faustinoaq:
#825 is already merged :tada:

PR #866, opened 2018-06-16
Remove webpack

faustinoaq:
# DO NOT MERGE YET! :warning: 

This highly depends on #865 and #864 

Also please be aware of all the **breaking changes**

**No NPM means, we need to use vanilla JS and plain CSS.**

Maybe we can propose a `--npm` flag, WDYT?

### Description of the Change

> :warning: This PR is optional, because to remove webpack we require to use plain JS and CSS without npm and sass dependencies.

This PR depends on #865 

### Alternate Designs

Keep configurable/removable webpack/sass dependency implemented on #865 

### Benefits

- Removes `amber.min.js` dependency, see: #825 
- No more NPM vulnerabilities by default, see #834 
- Use plain JS and CSS without dependencies
- No opinionated frontend, so, front-end devs can use whatever tools they want (react, yarn, rollup, parcel, etc) without default bundler config.
- No require NPM in amber generated projects, so we'd remove npm from amber Dockerfile:

https://github.com/amberframework/amber/blob/ddcb6f26835cc84a67f2d50a074ed17d8dc4d81f/Dockerfile#L7-L9

### Possible Drawbacks

- The plain JS is using ES5 to be compatible with most browsers
- Sass file has been replaced by plain CSS (so, no color variables here)
- To copy assets into `public/` folder, `amber watch` is using auxiliar tasks with `cp` and `cat` commands


src/amber/cli/templates/app/config/webpack/common.js
```diff
@@ -1,69 +0,0 @@
...
-    'main.bundle.css': './src/assets/stylesheets/main.scss'
-  },
-  output: {
-    filename: '[name]',
-    path: path.resolve(__dirname, '../../public/dist'),
-    publicPath: '/dist'
-  },
-  resolve: {
-    alias: {
-      amber: path.resolve(__dirname, '../../lib/amber/assets/js/amber.js')
```
drujensen: We will need to figure out how to include this for anyone using web sockets.

src/amber/cli/templates/app/.amber.yml.ecr
```diff
@@ -2,3 +2,44 @@ type: app
 database: <%= @database %>
 language: <%= @language %>
 model: <%= @model %>
+watch:
```
drujensen: awesome! 💯

src/amber/cli/templates/error/src/pipes/error.cr.ecr
```diff
@@ -1,25 +1,25 @@
...
-        context.response.status_code = 404
-        error = <%= class_name %>Controller.new(context, ex)
-        context.response.print(error.not_found)
-      rescue ex : Exception
-        context.response.status_code = 500
-        error = <%= class_name %>Controller.new(context, ex)
-        context.response.print(error.internal_server_error)
-      end
-    end
+class <%= class_name %> < Amber::Pipe::Error
```
drujensen: This doesn't seem anything to do with removing web pack.

PR #882, opened 2018-06-20
Update crystal-redis to 2.0.0

katafrakt:
### Description of the Change

Redis 2.0.0 [has been released](https://github.com/stefanwille/crystal-redis/releases/tag/v2.0.0) today. This pull request updates the dependency.

### Alternate Designs

No

### Benefits

New release supports automatic reconnecting when connection to redis server was lost, which is pretty important. 

### Possible Drawbacks

Amber will stop working with other shards using Redis that were not updated (such as Sidekiq), but probably they will catch up soon.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -48,7 +48,7 @@ dependencies:
 
   redis:
     github: stefanwille/crystal-redis
-    version: ~> 1.11.0
+    version: ~> 2.0.0
 
   shell-table:
     github: jwaldrip/shell-table.cr


faustinoaq:
@katafrakt @robacarp Do you know if this dependency update has some breaking change?

> Add reconnect option that will automatically reconnect a Redis instance when the Redis server connection is lost. NOTE: Default value is true! - Thanks to @kostya !

^^ Is this a breaking change? :sweat_smile:

faustinoaq:
Well, specs are already passing, so I gonna merge this :+1:

robacarp:
@faustinoaq I suppose this is technically a breaking change because someone could have invented some obscene logic which is triggered off of redis disconnecting. I don't think it's worth worrying about though.

PR #891, opened 2018-06-22
[WIP][Router] Cleans RedisStore

eliasjpr:
### Description of the Change

Users of the framework have asked why do they need to include redis
shard for a simple project.

### Benefits

Cleans the Session::Store to make RadisStore more independent. This can
allow to have RedisStore adaptor as its own shard removing the
dependency of Redis shard.

### Alternate Designs

This does not change the current behavior of the RedisStore.


faustinoaq:
@eliasjpr Nice!, Should modify/add some documentation as well?

src/amber/router/session/redis_store.cr
```diff
@@ -4,17 +4,17 @@ require "uuid"
 module Amber::Router::Session
   class RedisStore < AbstractStore
     @id : String?
-    getter store : Redis
+    getter store : Redis = Redis.new(url: Amber.settings.redis_url.to_s)
```
Thellior: Isn't it a better idea to specify this in the initialize? Maybe even with a check if redis_url is specified. If it's not used than raise an exception stating so?

PR #893, opened 2018-06-24
Using non-parenthesis method calls in scaffolded controllers

### LINKED ISSUE #892: Consistent code style for amber scaffolds
### Description

I have generated multiple scaffolds and the one thing that I always change is the way methods are being called. Specifically `render` and `redirect_to`.

Everywhere, `render` is called using paranthesis :-

```
render("new.ecr")
```
But `redirect_to` is called without using paranthesis :-

```
redirect_to "/articles"
```
I've hunted down the issue and would like to fix it to use a consistent style.

Which is the preferred style for Amber?

### Steps to Reproduce

Simply create a scaffold.

```
$ amber g scaffold articles title:string body:text
```

**Expected behavior:** 

Consistent code style for method calls.

**Actual behavior:** 

Mixed style. `render` uses parenthesis and `redirect_to` does not.

**Reproduces how often:** Always.

### Versions

Amber CLI (amberframework.org) - v0.8.0

Crystal 0.25.0 (2018-06-20)

LLVM: 5.0.1
Default target: x86_64-apple-macosx

shobhitic:
### Description of the Change

Fixes #892. Using non-parenthesis method calls in scaffolded controllers.

### Alternate Designs

Another option was to use a parenthesis call. Check gitter and #892 for discussion.

### Benefits

Consistent code style.

### Possible Drawbacks

Some people might prefer parenthesis.

### DIFF
--- a/src/amber/cli/templates/scaffold/controller/crecto/src/controllers/{{name}}_controller.cr.ecr
+++ b/src/amber/cli/templates/scaffold/controller/crecto/src/controllers/{{name}}_controller.cr.ecr
@@ -6,62 +6,62 @@ class <%= class_name %>Controller < ApplicationController
   end
 
   def index
-    <%= name_plural %> = Repo.all(<%= class_name %>)
-    render("index.<%= @language %>")
+    <%= name_plural %> = Repo.all <%= class_name %>
+    render "index.<%= @language %>"
   end
 
   def show
-    render("show.<%= @language %>")
+    render "show.<%= @language %>"
   end
 
   def new
-    changeset = <%= class_name %>.changeset(<%= @name %>)
-    render("new.<%= @language %>")
+    changeset = <%= class_name %>.changeset <%= @name %>
+    render "new.<%= @language %>"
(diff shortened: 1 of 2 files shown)


faustinoaq:
@shobhitic This is totally fine and I like it, but just to point out, looks like **sometimes** parenthesis are required in crystal, see: https://github.com/crystal-lang/crystal/issues/6254

shobhitic:
@faustinoaq Yeah, not just at the link you mentioned, but if we are ever calling a method with multiple arguments and using its return value in another method's multiple argument call, we'd need parenthesis.

Example: 

```
def foo(a, b, c)
  # something
end

def bar(d, e, f)
  # returns something useful
end

# to use bar's return value in foo, we'll have to use paranthesis.

# Invalid

foo 1, bar 1, 2, 3, 3

# Valid

foo 1, bar(1, 2, 3), 3
```

So that is also something which we'll need to keep in mind. (And I am realizing that you might already know this 😅 )

PR #897, opened 2018-06-26
Added support for returning a default value while using flash

shobhitic:
### Description of the Change

Now `flash` supports returning a default value in case the key does not exist.

### Alternate Designs

Alternatively, we skip this and rely on the `exists` method to see if the value is present, and if it isn't, we set it ourselves.

### Benefits

Makes the `flash` closer to a Hash in implementation. One weird to catch bug would have been that `flash` persists even when `[]?` with String is used as we hadn't overridden it and so it wasn't adding keys to `@read` array.

Also, this let's us do

```
<%- if value = flash[:something]? -%>
<%= value %>
<%- end -%>
```

in place of 

```
<%- if value = flash["something"]? -%>
<%= value %>
<%- end -%>
```
which just reads nicer!

### Possible Drawbacks

People who have been using `[]?` with String previously might have relied on the fact that this does not make the flash go away in this request. They'll see things breaking.

### DIFF
--- a/spec/amber/pipes/flash_spec.cr
+++ b/spec/amber/pipes/flash_spec.cr
@@ -44,7 +44,7 @@ module Amber
           flash_store.fetch(:some_key).should eq "some_value"
         end
 
-        it "supports default value" do
+        it "returns the default value" do
           flashes = {"some_key" => "some_value"}
           json = flashes.to_json
           flash_store = Flash::FlashStore.from_session json
--- a/spec/amber/pipes/flash_spec.cr
+++ b/spec/amber/pipes/flash_spec.cr
@@ -44,11 +44,29 @@ module Amber
           flash_store.fetch(:some_key).should eq "some_value"
         end
 
+        it "supports default value" do
+          flashes = {"some_key" => "some_value"}
+          json = flashes.to_json
+          flash_store = Flash::FlashStore.from_session json
+
+          flash_store.fetch(:some_other_key, "default").should eq "default"
+        end
+
         it "marks the value as read" do
           flashes = {"some_key" => "some_value"}
           json = flashes.to_json
           flash_store = Flash::FlashStore.from_session json
           flash_store.fetch("some_key")
+          flash_store = Flash::FlashStore.from_session flash_store.to_session
+
+          flash_store.has_key?("some_key").should be_false
+        end
+
+        it "marks the value as read for default value implementation" do
+          flashes = {"some_key" => "some_value"}
+          json = flashes.to_json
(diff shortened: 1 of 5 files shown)


spec/amber/pipes/flash_spec.cr
```diff
@@ -44,6 +44,14 @@ module Amber
           flash_store.fetch(:some_key).should eq "some_value"
         end
 
+        it "supports default value" do
```
eliasjpr: It "returns the default value"

shobhitic: Changed it.

shobhitic:
Good to merge?

robacarp:
I'm not really clear on what the benefit of a default value is for a flash message. Can you provide a real world example?

shobhitic:
@robacarp Sure.

I recently deployed an app - https://jsonkeeper.com/ which is JSON hosting service.

Whenever someone enters a valid JSON on the box on home page and clicks Save, the request goes to a `HostingController` where it creates the db record and a slug to access the provided JSON. 

After that, I redirect the user back to root url and set a flash `flash[:host_url] = "https://jsonkeeper.com/b/{{ slug }}"`.

Now, on the home page, I had written :-

```
<% if host_url = flash[:host_url] %>
  <!-- print the host_url for user to save -->
<% end %>
```

Unfortunately, I was getting a run time error when `flash` didn't have the `:host_url` key. I had to use a `exists?` method to fix it

```
<% if flash.exists?(:host_url) %>
  <% host_url = flash[:host_url] %>
  <!-- print the host_url for user to save -->
<% end %>
```

Even this doesn't work because `exists?` method expects a String as a key as it wasn't overridden. So the correct way is 


```
<% if flash.exists?("host_url") %>
  <% host_url = flash[:host_url] %>
  <!-- print the host_url for user to save -->
<% end %>
```

In the proposed solution, this gets cleaner looking solution so we can simply do 


```
<% if host_url = flash[:host_url]? %>
  <!-- print the host_url for user to save -->
<% end %>
```

This is closer to how Crystal treats its hash and what users will come to expect, unless they know that the keys are supposed to be String type.

shobhitic:
I just used the Crystal's API and implemented the same here.

@robacarp Yes, the way you've written will still work, but now it'll let my way to be supported as well, as I use flash for other things as well and keep `.alert` divs to only two or three types.

robacarp:
@shobhitic thanks for explaining in depth what you're after here.

shobhitic:
@robacarp Thanks for the merge!

Issue #895, opened 2018-06-25
respond_with json method doesn't work


shobhitic:
### Description

I was trying to send a JSON response from controller, and so I used `respond_with` method and `json` method in the block. Compiler threw an undefined method error which looked like this :-

```
in src/controllers/app_controller.cr:18: undefined method 'json'

      respond_with { json "{}" }
```

This is actually the documented usage as well, so I am not sure what could have been different about my code.

### Steps to Reproduce

1. Create an application and a controller where in a get request you respond with JSON using the method `respond_with { json "{}" }`.
2. Try to compile/run the application with `amber w`.
3. You'll be greeted with this error.

**Expected behavior:** It should compile and return the string as a response.

**Actual behavior:** Doesn't compile.

**Reproduces how often:** Always when I've tried it.

### Versions

Amber CLI (amberframework.org) - v0.8.0

Crystal 0.25.0 (2018-06-20)

LLVM: 5.0.1
Default target: x86_64-apple-macosx

### Additional Information

According to our gitter chat -

@faustinoaq thinks that "new proc type in `respond_with` helpers just broke something"

eliasjpr:
This seems to be intentional as the `"{}"` is a string see https://github.com/amberframework/amber/blob/master/src/amber/controller/helpers/responders.cr#L38 you might want to pass an empty Hash or Array.

faustinoaq:
@eliasjpr I think we can use method overloading instead of union types.

Something like this:

```cr
def json(json : String)
  # ...
end

def json(json : ProcType)
  # ...
end

def json(json : Hash(Symbol | String, String))
  # ...
end
``` 

WDYT?

shobhitic:
My bad. I guess I can simply `JSON.parse(json_host.data)` here.

eliasjpr:
@shobhitic you must do `JSON.parse("{}")` and it should work see https://play.crystal-lang.org/#/r/4eh5

shobhitic:
@eliasjpr Yes, I had done that, but it didn't compile without the `.to_json` method call.

drujensen:
Fix merged.  closing.

elorest:
Tested original question above and it works.

PR #898, opened 2018-06-26
[Responders] Overloads json method

eliasjpr:
## Description

Removes conditional by overloading the JSON method. 
Ref https://github.com/amberframework/amber/issues/895#issuecomment-400134684

## Benefits
- Simplyfies logic
- Removes Duplication
- Uses Amber::Support::MimeTypes
- It only process the requested mime type

### DIFF
--- a/src/amber/controller/helpers/responders.cr
+++ b/src/amber/controller/helpers/responders.cr
@@ -2,115 +2,60 @@ module Amber::Controller::Helpers
   module Responders
     TYPE_EXT_REGEX         = /\.(#{TYPE.keys.join("|")})$/
     ACCEPT_SEPARATOR_REGEX = /,|,\s/
-    TYPE = {
-      html: "text/html",
-      json: "application/json",
-      txt:  "text/plain",
-      text: "text/plain",
-      xml:  "application/xml",
+    TYPE                   = {
+      "html" => "text/html",
+      "json" => "application/json",
+      "txt" =>  "text/plain",
+      "text" => "text/plain",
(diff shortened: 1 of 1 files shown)


elorest:
> @robarcarp One benefit is that previously it used to run all the respond formats for each request, for instance, if you had defined the following respond_with, all responder format would process. Not instead only the response format requested will run.

@eliasjpr That's not entirely true. This has been discussed multiple times in other PR's. If you pass a string in that's true but if you pass a proc in it only processed the matching response. This was due to performance issues with procs. @damianham and I put a lot of thought and discussion into this.

Don't refactors and non bug fixes generally require 2 approvals?

elorest:
A couple of the changes I'm fine with but a lot of it just feels like moving code around and renaming stuff arbitrarily. I'm worried that it will introduce an unforeseen bug or change to existing projects.

eliasjpr:
@elorest now it does not process it is a string it skips only small renaming happened and a made use of the Amber::Support::MimeTypes which allowed the class to be cleaned, none of the specs were changed to ensure the same functionality exists

elorest:
Good point on the specs...

As mentioned in slack. Strings are still being rendered with this change. But they're being thrown away and garbage collected immediately instead of used. This doesn't change performance though.

I'm a bit concerned that if I Curled `curl https://domain.com/json_endpoint` it wouldn't return anything in the case that I didn't specify a type or had it defaulted to html since the value was thrown away. In the previous version it feel throw to the first specified if the request didn't match.

PR #900, opened 2018-06-26
Fetch redirected urls

### LINKED ISSUE #899: --recipe flag doesn't work
### Description

Ref: https://github.com/amberframework/recipes/issues/32

### Steps to Reproduce

1. `amber new app -r default`

**Expected behavior:** Create an app based on default recipe (or any recipe)

**Actual behavior:** Get an error output:

```
➜  ~ amber new blog -r default
12:11:40 Generate   | (INFO) Rendering App blog in ./blog from default
12:11:41 Generate   | (ERROR) Could not find the recipe default : 302 Found
Environment file not found for ./config/environments/production
```

### Versions

Amber v0.8.0

### Additional Information

Looks like this is issue is due crystal `HTTP::Client` doesn't support redirects yet, see: https://github.com/crystal-lang/crystal/issues/2721

/cc @damianham

faustinoaq:
### Description of the Change

This PR adds the ability to fetch redirected urls like the used on [recipes release page](https://github.com/amberframework/recipes/releases/)

### Alternate Designs

Use redirect option on `HTTP::Client` (not supported yet, see: https://github.com/crystal-lang/crystal/issues/2721)

### Benefits

Fixes #899 

### Possible Drawbacks

No


faustinoaq:
This works but is pretty hacky, What would happens if the URL redirects 3 or more times?

I think we need to implement something like `HTTP::Client.get("<url>", redirects: true)` @asterite @RX14 @straight-shoota @ysbaddaden, WDYT?

asterite:
There's an open issue for that in the stdlib. Someone just has to implement it (and it needs to be discussed how to implement it, which will probably result in it becoming a monster of complexity)

drujensen:
@faustinoaq should we merge this or should this wait?

drujensen:
Closing this.  Recipes will no longer be hosted by Amber.  We will work on allowing a recipe to be provided by a shard in a git repository.

PR #902, opened 2018-06-28
[MimeTypes] Refactors Mime Type

eliasjpr:
## Description:

Mime types are only being used as part of the router. This moves the Support::MimeTypes into Router::Mime, this is for better organization.


src/amber/router/mime.cr
```diff
@@ -0,0 +1,49 @@
...
+    FORMAT_HEADER          = "Accept"
+    TYPE_EXT_REGEX         = /\.(#{TYPES.keys.join("|")})$/
+
+    # Returns the Mime Type for a given format or file extname.
+    #
+    # ```
+    # Amber::Support::Mime.type("json")                  # => "application/json"
+    # Amber::Support::Mime.type(".jpg")                  # => "image/jpeg"
+    # Amber::Support::Mime.type("unknown")               # => "application/octet-stream"
+    # Amber::Support::Mime.type("unknown", "text/plain") # => "text/plain"
```
shobhitic: Need to change this comment to show `Amber::Router::Mime.type`

drujensen:
> Mime types are only being used as part of the router.

@eliasjpr I think mime types can be used outside of the router for other backend activities like file handling, right?  If it was only for routing, I would agree with this refactor but I don't think that is the case.

eliasjpr:
> Why do mime types belong in a module called router? What about mime types is involved in routing requests to controller-actions?
The router represents a http request lifecycle from receiving the request and delivering the response.

> It might be true that the mime types can be used for back-end activities but in Amber concerns it is only used by the framework request life cycle and is an important piece since the mime types are used for content negotiation for when delivering the response. 

My intention was to couple mime types with the router and this is perfectly fine since the router cannot work without the mime types. Having it coupled would allow to better handle the content negotiation (See Reference https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation). The overall goal is to move the mime type within the router to do the content negotiation at the request level instead at the controller level as is being done now by the Responders class, the benefit of this would be to know ahead of time how the router(amber server) which mime type it can respond with, this will prevent from processing all the methods in the `respond_with` block, and just execute the necessary method (html, json, xml, text) it will also simplify the request and the code. 

Improvement 
`->request->content negotiation->controller->execute only the negotiated mimetype -> send response`

Current Implementation
`->request->controller->execute the block (all mimetypes) -> content negotiation -> send response`

How the improve way s possible? 

It turns out that we have all the information we need to do the content negotiation upfront, but the way, is currently written it is not possible to make the decision up front. We know for instance that with `respond_with` defines which content-type the given resource can respond to, but this information is not stored so the block has to be processed instead to capture the content types possible to respond. By moving the mime types does not make much difference of keeping it in the support directory to get this work, but the code call chain does get simplified make it easier to understand.

robacarp:
> The router represents a http request lifecycle from receiving the request and delivering the response.

@eliasjpr if that's what the router represents, then I have no problem with this change.

eliasjpr:
What would be better name for amber::router?

PR #903, opened 2018-06-28
Add API-only CLI option

sdzyba:
### Description of the Change

Now it is possible to tell `amber watch` to not to run npm process using `api_only` option.
`api_only` is a term borrowed from Rails' [configuration](http://guides.rubyonrails.org/api_app.html#changing-an-existing-application). It might be reused in future for other API-specific functionality.

Related Amber recipe: https://github.com/amberframework/recipes/pull/33

### Alternate Designs

Something like `no_npm` option in `.amber.yml` might be used as well, although I find `api_only` more common and extendible in the future.
Also `--no-npm` argument for `amber watch` could be used but I find it unintuitive to use and this way we have some cli options in `.amber.yml` and some as arguments and this is inconsistent.

### Benefits

User is able to run `amber watch` without npm dependency.

### Possible Drawbacks

+1 another cli option to maintain.

### DIFF
--- a/src/amber/cli/config.cr
+++ b/src/amber/cli/config.cr
@@ -11,6 +11,7 @@ module Amber::CLI
     property database : String = "pg"
     property language : String = "slang"
     property model : String = "granite"
+    property api_only : Bool = false
     property recipe : (String | Nil) = nil
     property recipe_source : (String | Nil) = nil
 
@@ -21,6 +22,7 @@ module Amber::CLI
       database: {type: String, default: "pg"},
       language: {type: String, default: "slang"},
       model: {type: String, default: "granite"},
+      api_only: {type: Bool, default: false},
       recipe: String | Nil,
       recipe_source: String | Nil
     )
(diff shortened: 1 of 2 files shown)


elorest:
I do like the no_npm option. However, I don't see the point of an api_only option. This has been discussed multiple times. The reason that Rails added an api only option was because they only have one middleware stack. So an api still has to run through all the same handlers as a web app which pointlessly slows stuff down. 

With amber just create an `api` pipeline and use that. It won't be any slower if you have a web pipeline as well since the handler stack is completely different.

Rails and Amber have different paradigms when it comes to middleware.

sdzyba:
Sounds reasonable.
If other people are ok with `no_npm` (or maybe the opposite: introduce `npm` option which is enabled by default?) I can change it this way.

drujensen:
@sdzyba I believe the new `amber watch` config that @faustinoaq is working on #865 will handle this option.  You will be able to remove or replace the npm process.  You will also be able to add other watch processes.

sdzyba:
Got it.

faustinoaq:
@sdzyba Yeah, thank your for your understanding, I have been a bit busy latest weeks I'll continue my work on #865 this weekend :sweat_smile:

PR #912, opened 2018-07-04
Fix typo in application.cr.ecr

pgeraghty:
### Description of the Change

Just fixing a typo I noticed in the comments relating to Port Reuse.


robacarp:
Thanks @pgeraghty

PR #913, opened 2018-07-05
Fix class capitalization in _form template

### LINKED ISSUE #907: Class names for referenced classes not correctly capitalized
### Description

When I scaffold a resource which has a reference to a class with more than one word as it's name it does not get capitalized correctly.

### Steps to Reproduce

1.  amber new someapp
2.  cd someapp
3.  shards install
4.  amber generate scaffold **SomeResource** sometext:text
5.  amber generate scaffold SomeOtherResource sometext:text **some_resource:reference**
6.  amber watch

**Expected behavior:**

The application should compile and the view "src/views/some_other_resource/_form.slang" should contain:

`SomeResource.all.map{...`

**Actual behavior:**

The compilation fails because the class in the template is not capitalized correctly

`Some_resource.all.map{...`

### Versions

Amber CLI (amberframework.org) - v0.8.0
Crystal 0.25.0 (2018-06-25)

LLVM: 6.0.0
Default target: x86_64-pc-linux-gnu

ZeroPointEnergy:
This fixes the class capitalization of reference fields in
the _form templates.

Fixes #907

### DIFF
--- a/src/amber/cli/templates/field.cr
+++ b/src/amber/cli/templates/field.cr
@@ -86,5 +86,10 @@ module Amber::CLI
     def reference?
       self.type == "reference"
     end
+
+    def class_name
+      Inflector.classify(@name)
+    end
+
   end
 end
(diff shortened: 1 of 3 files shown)


eliasjpr:
@drujensen I sent a message to @paulcsmith about taking the inflector shard that Lucky has so we can use for Amber.

eliasjpr:
Meanwhile, this fix looks good to me

robacarp:
thanks @ZeroPointEnergy

PR #917, opened 2018-07-13
Make logger accept auto-generated configuration

lstrzebinczyk:
I would like to begin by saying hello, and thank you for your awesome work on amber, it is a pleasure to tinker with : )

### Description of the Change

Auto-generated `config/application.cr` file contains this line:

```
  # settings.logging.context = %w(request headers cookies session params)
```

however uncommenting it produces a compile time error: 

```
instance variable '@context' of Amber::Environment::Logging must be Array(String | Nil), not Array(String)
```

It seems reasonable that:
  - uncommenting generated config should work out of the box
  - this config should accept an array of strings

I therefore went ahead and implemented this change.

### Alternate Designs

Alternative approach could be auto-generating the documentation as 

```
# settings.logging.context = %w(request headers cookies session params) as String?
```

But this, in my opinion, contains unnecessary knowledge and raises unneeded questions when reading the setting.

This could go even further and introduce some sort of enum. This would introduce more type safety at the cost of readability.

### Benefits

Uncommenting application config should always work, and this is fixing an issue where it was not the case. Also the suggested behavior presents the simplest api for interacting with the config.

### Possible Drawbacks

This issue seems like a very minor oversight rather than a design decision. If I am wrong and there is a reason behind current design, I am happy to learn and correct my thinking: )

### DIFF
--- a/src/amber/pipes/logger.cr
+++ b/src/amber/pipes/logger.cr
@@ -1,13 +1,13 @@
 module Amber
   module Pipe
     class Logger < Base
-      alias Params = Array(String?)
+      alias Params = Array(String)
       Colorize.enabled = Amber.settings.logging.colorize
       FILTERED_TEXT = "FILTERED".colorize(:white).mode(:underline)
 
       def initialize(@filter : Params = log_config.filter,
                      @skip : Params = log_config.skip,
-                     @context : Array(String) = log_config.context)
+                     @context : Params = log_config.context)
       end
 
       def call(context : HTTP::Server::Context)
--- a/src/amber/pipes/logger.cr
+++ b/src/amber/pipes/logger.cr
@@ -7,7 +7,7 @@ module Amber
 
       def initialize(@filter : Params = log_config.filter,
                      @skip : Params = log_config.skip,
-                     @context : Params = log_config.context)
+                     @context : Array(String) = log_config.context)
       end
 
       def call(context : HTTP::Server::Context)
--- a/spec/amber/amber_spec.cr
+++ b/spec/amber/amber_spec.cr
@@ -43,6 +43,8 @@ describe Amber do
           server.logger.level = ::Logger::INFO
           server.logging.colorize = false
           server.logging.context = %w(request headers cookies session params)
+          server.logging.filter = %w(password confirm_password)
+          # server.logging.skip = %w()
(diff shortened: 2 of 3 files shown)


src/amber/pipes/logger.cr
```diff
@@ -7,7 +7,7 @@ module Amber
 
       def initialize(@filter : Params = log_config.filter,
                      @skip : Params = log_config.skip,
-                     @context : Params = log_config.context)
+                     @context : Array(String) = log_config.context)
```
drujensen: Is this a problem for `filter` and `skip` as well? Maybe we need to do this for all the Params.  Is so, should we update the Params type instead?

lstrzebinczyk: I uploaded the code I have, which should work, really, but it throws me this exception:

```
Unhandled exception: Expected String, not  at line 8, column 6 (YAML::ParseException)
  from /usr/share/crystal/src/yaml/nodes/nodes.cr:30:9 in 'raise'
  from /from_yaml.cr:34:7 in 'parse_scalar'
  from /usr/share/crystal/src/yaml/from_yaml.cr:56:3 in 'new'
  from /usr/share/crystal/src/yaml/from_yaml.cr:88:11 in 'new'
  from /usr/share/crystal/src/yaml/from_yaml.cr:140:3 in 'new'
  from src/amber/environment/settings.cr:49:5 in 'initialize'
  from src/amber/environment/settings.cr:49:5 in 'new'
  from /usr/share/crystal/src/yaml/from_yaml.cr:2:3 in 'from_yaml'
  from src/amber/environment/loader.cr:9:7 in 'settings'
  from src/amber.cr:24:3 in 'env='
  from spec/spec_helper.cr:14:1 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:104:5 in 'main_user_code'
  from /usr/share/crystal/src/crystal/main.cr:93:7 in 'main'
  from /usr/share/crystal/src/crystal/main.cr:133:3 in 'main'
  from __libc_start_main
  from _start
  from ???
```

I am anable to figure out if this is a new error that surfaced with this, or an issue with this PR itself. Also, I am going on a vacations and will not be able to push this further after today, sorry: (. Next month I will be back.

PR #922, opened 2018-08-07
Update to crystal 0.26

bcardiff:
Note: This changes should be backward compatible with 0.25.1
Ref: https://github.com/crystal-lang/crystal/pull/6414

ivar initializers will be evaluated without `self` in context.

Disclaimer: I've only checked that the source build with crystal nightly together with this patch.

### DIFF
--- a/src/amber/cli/commands/exec.cr
+++ b/src/amber/cli/commands/exec.cr
@@ -6,8 +6,14 @@ module Amber::CLI
 
     class Exec < Command
       command_name "exec"
-      @filename : String = "./tmp/#{Time.now.epoch_ms}_console.cr"
-      @filelogs : String = @filename.sub("console.cr", "console_result.log")
+      @filename : String
+      @filelogs : String
+
+      def initialize(__previous, __argv)
+        @filename = "./tmp/#{Time.now.epoch_ms}_console.cr"
+        @filelogs = @filename.sub("console.cr", "console_result.log")
+        super(__previous, __argv)
+      end
 
       class Options
         arg "code", desc: "Crystal code or .cr file to execute within the application scope", default: ""


src/amber/cli/commands/exec.cr
```diff
@@ -6,8 +6,14 @@ module Amber::CLI
 
     class Exec < Command
       command_name "exec"
-      @filename : String = "./tmp/#{Time.now.epoch_ms}_console.cr"
-      @filelogs : String = @filename.sub("console.cr", "console_result.log")
+      @filename : String
+      @filelogs : String
+
+      def initialize(__previous, __argv)
```
Sija: Why not use `*args` and `super`?

Sija:
Also, there's one more:

```
Error in src/pet-tracker.cr:3: instantiating 'Amber::Server.class#start()'

Amber::Server.start
              ^~~~~

in lib/amber/src/amber/server/server.cr:17: instantiating 'Amber::Server#run()'

      instance.run
               ^~~

in lib/amber/src/amber/server/server.cr:50: instantiating 'start()'

        start
        ^~~~~

in lib/amber/src/amber/server/server.cr:59: undefined method 'tls=' for HTTP::Server

      server.tls = Amber::SSL.new(settings.ssl_key_file.not_nil!, settings.ssl_cert_file.not_nil!).generate_tls if ssl_enabled?
             ^~~
```

bcardiff:
I didn't detect that because building the specs does not stress that path 🤔 .

To fix that, either something like https://github.com/kemalcr/kemal/pull/477 can be done, or 
since crystal-lang/crystal#6500 is merged, a more opaque bind can be used.

src/amber/cli/commands/exec.cr
bcardiff: It was the minimum change. But it should be equivalent.

bcardiff:
I also noticed that the shell-table shard version will need to be updated to support 0.26.0 jwaldrip/shell-table.cr#2

robacarp:
@bcardiff thanks for this, I've picked your commit over to #929 and will finish the effort there.

PR #926, opened 2018-08-13
Adds ability to specify Logger.color & fixes exception w/ missing config settings

noahlh:
### Description of the Change

This PR adds the ability to specify a log colorizing color when creating an `Amber::Environment::Logger`.  This is done by setting `Logger.color` (similar in style to `Logger.progname`).  I added this because I wanted to be able to set the color for a plugin/related process that uses Amber's logging mechanism.

For example, in Granite's initialization (project/config/initializers/database.cr), the following two lines setup the logging mechanism:

````
Granite.settings.logger = Amber.settings.logger.dup
Granite.settings.logger.progname = "Granite"
````

With this change, one could now add a third line in order to have log lines stand apart from standard Amber processes:

````
Granite.settings.logger.color = :blue
````

While making this change, I also uncovered a small bug in the way settings parameters are loaded.  Currently, in `Amber::Environment::Settings`, the `YAML.mapping` specifies a type (`Logging::OptionsType`) and default (`Logging::DEFAULTS`) for those settings in an environment config.  Because of `Logging::OptionsType` being a NamedTuple, however, if any of the settings are  missing from the `logging:` section of an env yml file, an exception is thrown.

(This is, I believe, because the presence of a `logging:` section in the yml file means that the default is not used, but the lack of an expected key/value pair within that logging block throws an exception.  So you can leave out the entire `logging:` section and the defaults will be applied, but you cannot leave out a key within a section your provide).

So - I've changed `Logging::OptionsType` to be a Hash instead of a NamedTuple and merge the `Logging::DEFAULTS` upon load.  For some reason that I'm not entirely certain of (but having to do with how YAML.mapping works with Hashes vs. NamedTuples), now if you leave a logging setting out of the env yml file, no exception is thrown and the defaults can be applied properly.

(I did this because by adding a `color` option, I certainly didn't want to break all existing installations that don't explicitly specify the color key nor did I want to require it going forward.)

I also added appropriate tests to handle this new behavior & functionality.

### Benefits

This allows for finer-grained configuration of logging settings and for a user to specify the color of both amber and related processes' logging activity.  Hooray for a more colorful terminal!

### Possible Drawbacks

I do think that having `Logging::OptionsType` be an alias'd NamedTuple is quite elegant, and this change is slightly less memory-efficient and requires a bit more verbosity in `Amber::Environment::Logging#initialize`.  But without diving deep into how YAML.mapping works, I'm not sure how to avoid throwing an exception from a missing key when parsing an env config file.  

If I've taken the wrong approach and someone can point me in the right direction, I'm more than happy to make any & all changes!

Thanks in advance for any feedback as well.  This is all a learning process for me, so code/design critique is appreciated.

### DIFF
--- a/spec/amber/environment/logger_spec.cr
+++ b/spec/amber/environment/logger_spec.cr
@@ -24,5 +24,34 @@ module Amber::Environment
         end
       end
     end
+    describe "#color" do
+      it "logs messages with passed color attribute" do
+        IO.pipe do |r, w|
+          logger = Logger.new(w)
+          Colorize.enabled = true
+          logger.info "Test", "Amber", :blue
+          r.gets.should match(/\[34mAmber/)
+        end
+      end
+      it "logs messages with default color attribute" do
+        IO.pipe do |r, w|
+          logger = Logger.new(w)
+          logger.progname = "Amber"
+          Colorize.enabled = true
+          logger.info "Test"
+          r.gets.should match(/\[96mAmber/)
+        end
+      end
+      it "logs messages when #color is used" do
+        IO.pipe do |r, w|
+          logger = Logger.new(w)
+          logger.progname = "Amber"
+          logger.color = :green
+          Colorize.enabled = true
+          logger.info "Test"
+          r.gets.should match(/\[32mAmber/)
+        end
+      end
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


robacarp:
@noahlh thank you for this, and especially for fixing the NamedTuple bug. It's been on my radar ever since working with the smtp settings.

PR #930, opened 2018-08-18
Use Hash composition over inheritance in SessionHash and FlashStore

### LINKED ISSUE #920: Raven.cr integration build fail
Since amber `v0.8.0` [raven.cr](https://github.com/Sija/raven.cr) integration is failing to compile.

Below is error given when trying to compile amber's _pet-tracker_ project with raven.cr from `develop` branch and crystal `v0.26.0`:

```
$ ../crystal/bin/crystal build src/pet-tracker.cr --debug

Module validation failed: Function return type does not match operand type of return inst!
  ret %"(Array(AnyHash::JSONTypes::Value) | Bool | Float32 | Float64 | Hash(String | Symbol, AnyHash::JSONTypes::Value)+ | Int128 | Int16 | Int32 | Int64 | Int8 | JSON::Any | Set(AnyHash::JSONTypes::Value) | String | Symbol | Time | UInt128 | UInt16 | UInt32 | UInt64 | UInt8 | Nil)" %1, !dbg !28
 %"(Array(AnyHash::JSONTypes::Value) | Bool | Float32 | Float64 | Hash(String | Symbol, AnyHash::JSONTypes::Value) | Int128 | Int16 | Int32 | Int64 | Int8 | JSON::Any | Set(AnyHash::JSONTypes::Value) | String | Symbol | Time | UInt128 | UInt16 | UInt32 | UInt64 | UInt8 | Nil)" = type { i32, [3 x i64] } (Exception)
  from Crystal::CodeGenVisitor#finish:Nil
  from Crystal::Compiler#codegen<Crystal::Program, Crystal::ASTNode+, Array(Crystal::Compiler::Source), String>:(Tuple(Array(Crystal::Compiler::CompilationUnit), Array(String)) | Nil)
  from Crystal::Compiler#compile<Array(Crystal::Compiler::Source), String>:Crystal::Compiler::Result
  from Crystal::Command#run:(Bool | Crystal::Compiler::Result | Nil)
  from main
```

I'm trying to debug this since the very beginning but it just gets weirder and weirder and ATM I'm out of the ideas wtf is the cause of this...

Refs https://github.com/crystal-lang/crystal/issues/6253

Sija:
First and foremost inheriting from stdlib classes is considered a bad practice.
Secondly it triggers what I think is a Crystal bug, which happens to be seen in #920, making it unable to compile Amber with raven.cr.

Also, I've removed unused `SessionHash#find_entry` method which seems obsolete.

Fixes #920


src/amber/router/flash.cr
```diff
@@ -9,7 +9,10 @@ module Amber
         FlashStore.from_session(flash_content)
       end
 
-      class FlashStore < Hash(String, String)
+      class FlashStore
+        @raw : Hash(String, String)
```
robacarp: The name `raw` doesn't mean much to me, but I'm not sure I can think of a better name. Perhaps `@store`?

PR #940, opened 2018-08-26
Bumped version number to v0.9.0.

elorest:
### Description of the Change

Merging in 0.9.0 release.

### DIFF
--- a/src/amber/cli/templates/app/shard.yml.ecr
+++ b/src/amber/cli/templates/app/shard.yml.ecr
@@ -18,8 +18,8 @@ targets:
 dependencies:
   amber:
     github: amberframework/amber
-    branch: master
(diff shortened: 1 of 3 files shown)


src/amber/cli/templates/app/shard.yml.ecr
```diff
@@ -18,8 +18,8 @@ targets:
 dependencies:
   amber:
     github: amberframework/amber
-    branch: master
-    #version: <%= Amber::VERSION %>
+    version: <%= Amber::VERSION %>
+    #branch: master
```
drujensen: I take it, we will uncomment this later?

elorest: That should be changed before merge in. Let me do that now.

PR #946, opened 2018-08-31
Removes CLI from source in favor of Amber CLI shard

eliasjpr:
### Description of the Change

Remove Amber CLI from source. We can still prove the CLI by using crystal shards target builds and adding `amberframework/amber-cli` to the development section of shards.yml

### How to test

1. Generate new project `amber new test_project`
2. Scaffold 25 models `amber g scaffold Post{n} title:text`
3. Point to this branch 

```yams
dependencies:
  amber:
    github: amberframework/amber
    commit: a37b91990570b7308d8ed3e1d6f1cc7fcd6482d1
```
4. Install shards `shards install`
5. Run amber watch `amber w` (the logs prints time outputs)

### What I have found:

- Compilation time might be ~5 seconds faster between current branch and master this is not truly an improvement.
- Compilation time is incremental, more files more time, this is more of a crystal issue.
- Amber installation time improves since user only needs to download CLI
- Better organization. CLI maintenance lives on its own repo.
- No issues with db drivers as we previously saw. 
- Users can still compile the CLI locally per project using `shard build target` and adding amber cli dependency
- Clear separation of Amber Lib and Amber CLI
- Makes it easier to release Amber Lib since most of the external shard dependencies live in CLI and makes it easier to support new versions of Crystal, since the CLI does always need to be updated.

### Downsides

- Versioning another repo might bring some complexity, we can always mirror the same version number found in amber lib
- No real improvement over compilation time, this is a crystal issue.

## Conclusion

I personally don't think this improves compilation time, this is more a Crystal issue than an Amber issue. I do incline of having clear separation of the modules and with the added benefits breaking the shards dependency graph that we currently have. The CLI has most of the external dependencies and sometimes releases become very time consuming coordinating the shards that have to be updated, by splitting the repo we isolate the problem in regards to shards dependencies.

This is what I think. Please weight your thoughts. 

Please Thumbs up or down to (approve or disapprove) of this PR

### Dependent PRs 

Clean up and fixes that were missing in Amber CLI https://github.com/amberframework/amber_cli/pull/3


drujensen:
@eliasjpr This is great.  Can you confirm this improves the compile times?

elorest:
I won’t stand in the way of it if everyone thinks it’s a good idea. Personally I feel like it will just add complexity and make releasing and compatibility more difficult. I don’t actually see how having it together is causing any issues that wouldn’t exist regardless of where it’s at.

We’d also have to move out support libraries like our encryptor too that are used in both cli, environment and cookies.

I'm worried that changing all of this will inevitably lead to unforeseen bugs which must be fixed instead of working on known bugs, at a time when most of our development time is very limited.

As it currently stands in regard to binary size and compile time.

1. Cli doesn’t get compiled into amber projects so compile time wouldn't be affected.
2. CLI doesn’t compile in amber project either so in terms of compilation it’s still 2 different binaries with multiple shared libraries.
3. The same goes for binary size. Would decrease the repo size though. But since it would be a dependency in shards it would still take as long to download and compile.

drujensen:
@elorest Let's just make sure we are not accidentally including everything via splat and we should be fine.  

I think the other benefit here is the shards download time and the massive list of shards included in our projects.  Even if they aren't included, there is a perception that there is a lot of stuff required when there isn't.  This gives a bad impression for new users.

eliasjpr:
Ping!

eliasjpr:
@robacarp @faustinoaq @elorest please weight in on the status of these changes.

drujensen:
@robacarp I used to do this with Kemalyst, but it sucked when deploying.  But, now that we can add the shard as a `development-dependency`, the compilation will only happen for development.  I like it!  👍

elorest:
Back when we used to do this with amber and kemalyst it was really easy for the cli to get out of sync with the latest version of amber, which made releasing hard and led to a lot of confusion in gitter and issues. Sure these issues aren't the end of the world but I don't see what issues we're having how it currently is that it justifies creating even more repos and dependancies (amber, amber cli, amber support, and possible environment). This increases our complexity. Not to mention the confusion caused by calling a binary compiled from amber_cli `amber`; alternatively calling the binary `amber_cli` would suck more.

drujensen:
@Thellior just to confirm, did you use this branch instead?
```
dependencies:
  amber:
    github: amberframework/amber
    branch: ep/remove-cli
```

Thellior:
I personally preference that we have a cli repo and install that as command line binary.
I'm only concerned about the work we have to do to to remove the cli from the repository.

If i'm correct we also need to remove the "Makefile" from the amber repository. Don't we also have to change the installation process of the amber framework? Like the brew scripts / documentation etc?

Thellior:
@drujensen Yes i used this branch but i used the commit sha instead.

```yml
  amber:
    github: amberframework/amber
    commit: 2a47d0665cbd4a6112221d92dee7a7b4c503b650
```

drujensen:
> Don't we also have to change the installation process of the amber framework? Like the brew scripts / documentation etc?

yes, yes we do.  It's going to be a lot of work.  Is it worth it?  That is a hard question to answer.

Thellior:
I think it's good that have it removed from the main framework code. If you look at angular for example it has a node page "@angular/cli"(https://github.com/angular/angular-cli). It's separated from the "@angular/core". I like that behaviour because the main code is not including the CLI so if a new user of the amber framework checks the source code it's not bloated with the cli code

epergo:
About the conversation of compiling amber cli only for development, remember that amber cli is needed for production too.

Right now (well it's been some time since I did something in Amber) when you have to run migrations you do the following:

```crystal
amber db:migrate
```

If you don't have the cli compiled for production, how do you run migrations?

Thellior:
The Amber CLI was not compiled in your app in the first place. You had Amber installed by the whole framework by either make install or the brew install. In the new situation you install the binary from the cli repository. It Will Give you the binary so you can run amber db:create

epergo:
@Thellior Exactly, I have to install the binary too, always. So what's the point of split the project if I have to install both no matter what

drujensen:
@Thellior @epergo  I was envisioning that you would initially create the project using the apt-get or brew binary but we would generate a `bin/amber` in the project itself that matches the project.

@epergo you have a good point that maybe we need to keep this as a dependency instead of a development dependency. 

The good news is with this change, you can choose to include it or remove it in production.  Its up to you.

Thellior:
You only install the cli repository because that can be compiled to a binary. The original Amber framework is included in the shards file.  So On production you Have your project installed anyways. You only install the binary like you already had to

epergo:
@drujensen You are right, with the split the user has the option to choose.

@Thellior I'm not sure what you mean by `You only install the binary like you already had to` I don't have to install the binary right now, I just have 
```yaml
targets:
  sourcecode-api:
    main: src/sourcecode-api.cr

  amber:
    main: lib/amber/src/amber/cli.cr
```

And with that I have the amber cli.

Thellior:
@epergo What i mean that on your production server you had to install amber right? Or do i understand something incorrectly. So if you are on ubuntu you have to 

Old situation:
```shell
curl -L https://github.com/amberframework/amber/archive/stable.tar.gz | tar xz
cd amber-stable/
make
sudo make install
```

New situation:
```shell
curl -L https://github.com/amberframework/amber_cli/archive/stable.tar.gz | tar xz
cd amber-stable/
make
sudo make install
```

So the binary for you production only changes from the repository that is what i meant

epergo:
For production? Not at all, just with the configuration I wrote before and running:

`shards build production` 

And with that you have `bin/amber` executable.

Thellior:
@epergo O really i didn't know that. Thanks for the clarification :)

eliasjpr:
> About the conversation of compiling amber cli only for development, remember that amber cli is needed for production too.
>
> Right now (well it's been some time since I did something in Amber) when you have to run migrations you do the following:

This can be corrected easily by simply moving amber from development deps to deps. 

In the case of heroic build pack the amber cli is compiled for you.

eliasjpr:
> For production? Not at all, just with the configuration I wrote before and running:
> `shards build production`
> And with that you have bin/amber executable.

Should be able to do the same, just move the dependency. Even better we can modify `shards.yml` and use a post install script. Something like.

```
scripts:
  postinstall: ./lib/amber_cli/make && cp ./lib/bin/amber ./bin/
```

eliasjpr:
@Thellior a pr was opened on Amber CLI to fix the compile issues. See https://github.com/amberframework/amber_cli/pull/3

eliasjpr:
Live Reload dependency has been removed from Amber CLI and now lives in itw own shard `amberframework/live_reload`.

epergo:
Well, with more shards and more repos, more release processes, more issue
trackers to have an eye on and in general a bit more of work and with
little benefit.

Is just a concern, if you (as core team) don't think is a problem go ahead.
Is not criticism!

eliasjpr:
I really think that the major concern here is putting the cli code in a separate repo, the logical separation presented here is acceptable and beneficial.

If you look at the Rails repo they are using a mono repo so it might looks like the CLI is part of the Raisl lib project but looking closely the CLI is really separates/modular from the rails library. 

I can see the downside of having multiple repos but the logical separation that this Pr introduces is fine IMO. Maybe we should move to a monorepo style as well.

Issue #943, opened 2018-08-28
Missing and not a little important functionality: nested params


winhackua:
### Submitting questions

it would be great if we implemented parsing nestled query,

### Versions
0.9

elorest:
Json does allow for nesting.

https://github.com/amberframework/amber/blob/master/src/amber/router/params.cr#L57

For example `params.json(:user)` would return all of the nested information under `:user`.

eliasjpr:
@winhackua can you please elaborate more on this, we are considering enhancing this functionality and more detail will help to shed some light for an approach

elorest:
This is something we've spent a lot of time thinking about. Due to the way crystal handles memory differently from ruby, this is difficult to do without making: dealing with params substantially more difficult.

For instance if params[:key] could equal `String | Hash(String, String | Hash(String, Hash...))` recursively then the user has to cast it to the correct value every time. I'd appreciate any suggestions on this. I'm also going to explore @eliasjpr's closed PR linked above.

eliasjpr:
Update: I am looking to improve more the domain for contracts, agree that Contract name, please make some name suggestions. We will be porting this over to Amber as soon is ready for use.

Issue #770, opened 2018-04-22
400 instead of 500 when params validation failed


c910335:
### Description

`ValidationFailed` is apparently a client error due to passing invalid params.
Thus, server should response `400 Bad Request` instead of `500 Internal Server Error` when `ValidationFailed` is raised.

### Steps to Reproduce

1. Start a new amber project
2. Generate a scaffold with at least one field
3. POST it without params

**Expected behavior:** 400 Bad Request

**Actual behavior:** 500 Internal Server Error

**Reproduces how often:** always

### Versions

v0.7.2 and master

### Additional Information

I think @faustinoaq is ready to fix this.

faustinoaq:
@c910335 Thank you for helping me to open issues, very appreciated! :+1: :tada:

faustinoaq:
Fixed by #864

faustinoaq:
#864 is not merged yet, still work in progress :sweat_smile:

Issue #698, opened 2018-03-17
Better Validation system with built-in validations rules


diaafares:
I propose building a better Validation system with built-in validations rules and built-in i18n supoort. For example: 

- you can use built-in validation rules in Rails like this:

```ruby
class Person < ApplicationRecord
  validates :name, :login, :email, presence: true
  validates :name, length: { minimum: 2 }
  validates :bio, length: { maximum: 500 }
  validates :password, length: { in: 6..20 }
  validates :registration_number, length: { is: 6 }
end
 
Person.create(name: "John Doe").valid? # => true
Person.create(name: nil).valid? # => false

```

Rails have validation helpers for: acceptance, validates_associated, confirmation, exclusion, format, inclusion, length, numericality, presence, absence, uniqueness ..etc!

- And Laravel, provide three ways for validation:

1. using vaildate method in the any controller:

```php
$validatedData = $request->validate([
        'title' => 'required|unique:posts|max:255',
        'body' => 'required',
    ]);
```

2. Creating validator instance manually using the Validator facade in any controller:

```php
$validator = Validator::make($request->all(), [
            'title' => 'required|unique:posts|max:255',
            'body' => 'required',
        ]);

        if ($validator->fails()) {
            return redirect('post/create')
                        ->withErrors($validator)
                        ->withInput();
        }
```

3. For more complex validation scenarios, Laravel provides ability to create a "form request" classes which  are custom request classes that contain validation logic.

Laravel provides a lot of built-in validation rules! 
https://laravel.com/docs/5.6/validation#available-validation-rules
![laravel](https://user-images.githubusercontent.com/1811607/37553860-90901790-29d8-11e8-8c69-a850f97c9154.png)

Finally, I'm a beginner at Crystal and Ruby and came from Laravel/PHP background, I don't know If I can help in participating in re-implement Amber Validation system, but I will try to do my best to help you guys!

Useful links to check:
**Rails:**
[Rails Validation documentation](http://guides.rubyonrails.org/active_record_validations.html)
Rails Validation implementation:
http://api.rubyonrails.org/classes/ActiveModel/Validations/ClassMethods.html
https://github.com/rails/rails/blob/cdb9d7f481a6b299cd53a0f647397da60d806a21/activemodel/lib/active_model/validations.rb
https://github.com/rails/rails/blob/cdb9d7f481a6b299cd53a0f647397da60d806a21/activemodel/lib/active_model/validations/validates.rb
[...]

Issue #688, opened 2018-03-09
Field validators not working as expected


danielRomero:
### Description

I think field validators in controllers doesn't work as expected.

I'm trying to understand how it works with a new fresh installation following the sample application described in the guides

https://amberframework.org/guides/getting-started/quickstart/zero-to-deploy.md#zero-to-deploy

### Steps to Reproduce

create sample application
`amber new test-amber -d pg -m granite -t ecr`

create scaffold
`amber g scaffold Pet name:string breed:string age:integer`

migrate database
`amber db create migrate`

----------------------------

**Case 1: Validation optional without block**

`optional(:age)`

Result: Not compile

Message:
`in src/controllers/pet_controller.cr:71: undefined method 'optional'`

**Ok**, the block is mandatory and must return Bool
https://github.com/amberframework/amber/blob/master/src/amber/validators/params.cr#L59

----------------------------

**Case 2: Validation optional returning true**

`optional(:age) { |f| true }`

- Field empty in form

Result: Resource not created

Flash message: `Could not create Pet!`

- Field filled  in form

Result: Resource created

Flash message: `Created Pet successfully.`

----------------------------

**Case 3: Validation optional returning false**

`optional(:age) { |f| false }`

- Field empty in form
Result: Resource not created

Application error
`ERROR: Validation failed. [Amber::Validators::Error(@param="age", @value="", @message="Field age is required")] (Amber::Exceptions::Validator::ValidationFailed)`

- Field filled  in form

Result: Resource not created

Application error
`ERROR: Validation failed. [Amber::Validators::Error(@param="age", @value="32", @message="Field age is required")] (Amber::Exceptions::Validator::ValidationFailed)`

----------------------------

**Expected behavior:** When an optional param is missing in the request, the resource must be saved

### Versions

```
Crystal 0.24.1 (2018-01-27)

LLVM: 5.0.1
Default target: x86_64-apple-macosx
```

```
Amber CLI (amberframework.org) - v0.6.7
```

```
shards:
  amber:
    github: amberframework/amber
    version: 0.6.7
```

faustinoaq:
@danielRomero Thank you for reporting this issue!

BTW, @eliasjpr Why params validators and granite validators looks so different? :sweat_smile:

faustinoaq:
Related issue here: https://github.com/amberframework/amber/issues/698

drujensen:
This has been fixed.  The block is now optional.

PR #954, opened 2018-09-16
Ep/new params validator

eliasjpr:
Addresses

https://github.com/amberframework/amber/issues/943
https://github.com/amberframework/amber/issues/770
https://github.com/amberframework/amber/issues/698
https://github.com/amberframework/amber/issues/688

> More tests to come...

### Description of the Change

Introduces Contracts. A contract is an abstraction to handle validation of arbitrary data or object state. It is a fully self-contained object that is orchestrated by the controller.

The Contract macros helps you defining contracts and assists with instantiating and validating data with those contracts at runtime. The `contract` macro will define a Struct with properties and methods. 

Contract macro: 

```crystal
# name of the contract for, 
# key: param key to parse
# block contains param definitions, you can also define other methods as needed.
contract(name, key) do 
...
end
```
Param macro:

Defines properties types and validations for a contract. 
```crystal
  param email : String, length: (50..50), regex: /\w+@\w+\.\w{2,}/
  param! email : String, length: (50..50), regex: /\w+@\w+\.\w{2,}/ # raises error
```

Available validations are:
```crystal
eq - value must qual to
lt - value must be  less than
lte - value must be  less than or equal to
gte - value must be greater than or equal to
gt - value must be greater than
in - Values must be in collection
length - length (objects must respond to size)
regex - regex comparison
```
Example:

```crystal
  contract("User") do
    param email : String, length: (50..50), regex: /\w+@\w+\.\w{2,}/
    param name : String, length: (1..20)
    param age : Int32, gte: 58, eq: 24, be: "Age"
    param alive : Bool, be: false
    param childrens : Array(String)
    param childrens_ages : Array(Int32)
    contract("Address", "user.address") do
      param street : String, length: (10..35)
      param city : String, eq: "Jersey City"
      param state : String, in: ["NJ", "NY"]
      param zip_code : String, length: (5..5)
      contract("Location", "user.address.location") do
        param longitude : Int32
        param latitude : Int32
      end
    end
  end
end
```
The example above will create a `User` with properties defined by params with an Address sub struct that can only be accessed via the user. 

Available methods
```
#valid?
#valid!
#error(param : String, value : Any, message : String) 
#errors
#to_h
#validate
```
### Other usages

You can use Contracts outside of controllers by defining a class as follow:

```crystal
struct UserPOCO
    include Contract::Validation
    @raw_params : Amber::Router::Params

   # An instance method will the following signature will be created
   # Key: is a path to the nested param in form of `user.address.city`, `user.address.street`...
   def initialize(raw_params, key : String)
   end
end
```

### Benefits

- Strict Typed objects for params
- Params validation
- Easy/Simple DSL
- Contracts are modular and can be used to outside of controllers (Think POCO Plain Old Crystal Objects)
- Contracts can happen earlier in the request lifecycle not necessarily in a controller this is great if someone wants to validate/fail request earlier.

### Possible Drawbacks

- Breaking change (Removes old param validation
- Generators Need to be updated and tested


robacarp:
Interesting idea, I like it. Can you show how this is used in a controller action? Eg how is this supposed to be used to read parameters and build or update objects?

Sija:
- Validators should be easily extensible, preferably inheriting from some `abstract class Validator`
- Personally I find confusing and redundant specifying full sub-contract path, like in here:

```cr
contract("User") do
  # ...
  contract("Address", "user.address") do
    # ...
    contract("Location", "user.address.location") do
    # ...
    end
  end
end
```

I'd suggest using just the last path component, detecting the rest automagically:

```cr
contract("User") do
  # ...
  contract("Address", "address") do
    # ...
    contract("Location", "location") do
    # ...
    end
  end
end
```

robacarp:
Nice. This would probably pair well with a PR to Granite to accept a contract directly, rather than `.to_h`-ing the contract.

robacarp:
@Sija has a nice list of features, to which I'd just add that it should be natural for a developer to simply access the params as a raw Hash perhaps a (String => String | Int | Bool or something). This kind of contract / validation on params is going to be really useful _most_ of the time, but there will always be times when the contract paradigm breaks (or is unwanted) and the developer just wants unrestricted access to the params.

Sija:
@robacarp 👍 even better, such a contract could create a `record` (or `struct`) with typed fields, taken from properties defined using DSL - in such a way we'd have compile-time safety + lack of annoying Hash value unions.

eliasjpr:
@robacarp the `context.params` will still exist and be accessible.

PR #957, opened 2018-10-03
Dj/update granite

drujensen:
### Description of the Change

This updates granite to 0.14.0.  There are some breaking changes in this release to the `has_many` syntax.  Also, we have removed pluralization from Granite so you need to provide the table name if you want to match rails.  

The generator in Amber already generates the pluralized name and overrides the default name in Granite.

This also updates a couple template. 
- Declare the `primary id : Int64` instead of using a comment.  This makes it easier for users to understand how to replace the primary key if they need too.
- Move create method in the controller above the update method.  This was already done for Crecto and this is pretty standard order in rails and other frameworks.

### Alternate Designs

N/A

### Benefits

The `has_many` in Granite is now configurable to provide the foreign key and override the field name and class_name that may not match the one we try to generate.

### Possible Drawbacks

This is not backward compatible so if someone updates Amber, they will need to change their `has_many` declarations

previously: `has_many :posts` needs to be changed to: `has_many posts : Post` if you want the relationship to be pluralized or `has_many :post` if you choose to use the default singular name in granite.

### DIFF
--- a/spec/support/fixtures/cli_fixtures.cr
+++ b/spec/support/fixtures/cli_fixtures.cr
@@ -61,7 +61,7 @@ module CLIFixtures
 
       belongs_to :user
 
-      primar id : Int64
+      primary id : Int64
       field title : String
       field body : String
       field published : Bool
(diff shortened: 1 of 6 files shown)


elorest:
Should we release with this?

PR #961, opened 2018-10-15
Fixed responders extension routing.

elorest:
### Description of the Change

When `radix` was replaced with `amber_router` it broke the ability of `respond_with` to choose mime types with extensions. I.e. `listings.json` or `listings.xml`

### Alternate Designs

Change router to create additional routes. In benchmarks this was slightly slower in addition to being more cumbersome on developers.

### Benefits

It works again.


faustinoaq:
screenshot_20181019_015509
@elorest @amberframework/contributors Thank you! :100:

PR #963, opened 2018-10-19
Shard based recipes

damianham:
### Description of the Change

This change supports recipes as shards from github in addition to recipes in a zip file at a given URL.  If no recipe is given with 'amber new' the 'amber new' command will use the default recipe currently at github.com/damianham/amber_base.  This default shard recipe needs to be copied to the amberframework organisation by one of the organisation owners and then change the recipe command line argument default in src/amber/cli/commands/new.cr to 'amberframework/amber_base'

### Alternate Designs
No alternate designs

### Benefits
Packaging a recipe as a shard in a github repo is easier than packaging as a zip file.  The recipe is installed using the 'shards install' mechanism through a bootstrapping process.  Anyone will be able to easily create a new recipe.  Amber organisation only needs to maintain a single default recipe.

### Possible Drawbacks

None apparent


damianham:
@drujensen there are good reasons to move to a default recipe rather than built in templates. Maintenance of the default app template would then be in just 1 place rather than 2 places.  If we continue with the built in app templates then there is guaranteed to be a disconnect between the default recipe and the built in templates.  The end goal could be that eventually all of the generator templates move to the default recipe (this can happen once recipe hooks are implemented).  By disconnecting the code templates from the Amber binary we make it easier for people to update the code templates.  The recent upgrade to bootstrap4 is a good case in point.  Had the code templates been external in an independent recipe we would all have been able to test the upgrade so much easier simply by creating an app using the recipe rather than checkout a branch of Amber and building a new binary etc.

I am sure you are aware of Yeoman generators - http://yeoman.io/generators/ which brilliantly illustrates the benefit of independent code generation.  If we move exclusively to recipe based code generation I think we will see many more recipes being created by an enthusiastic community.  We would only need to provide a means for the recipe author to publish details of their recipe on amberframework.org as they do with Yeoman generators.

I think we will have a wider adoption of Amber if we have a whole myriad of recipes available for developers to kick start their Amber projects.  This will only happen if the use of recipes is championed by the core team.  Hopefully, one day I will be able to work for clients on Amber based contracts rather than just RoR.  Widespread adoption of Amber will bring that about.

drujensen:
@damianham WDYT if we move all of the floating variables into a class as properties?  Then we can provide methods that would be the hooks for the callbacks?

This would be the contract between the recipes and the generator.  We can then document this class to make it clear what each property represents? i.e. `fields`

damianham:
good points Dru.  I will revert the default to internal for now.

damianham:
@drujensen I will try to get some time to work on it.  I agree that we will make more progress on the recipes if that is where the focus is.  I am going to simplify the base recipe so it is just granite and move the crecto stuff to a crecto recipe.  That should make it easier for other people to create new recipes.

src/amber/version.cr
```diff
@@ -1,3 +1,3 @@
 module Amber
-  VERSION = "0.10.0"
+  VERSION = "0.10.1"
```
drujensen: We use `crelease` to handle this.
```suggestion
  VERSION = "0.10.0"
```

drujensen:
@damianham Do you want to merge this and we can clean up the recipes as a future task?

damianham:
merging is blocked - Ameba is throwing up an error 
https://travis-ci.org/amberframework/amber/jobs/447347142

elorest:
Wait... This doesn't mean that `amber new` uses recipes by default now does it?

damianham:
@elorest no it uses the internal templates by default

PR #964, opened 2018-10-19
Stub of upgrade command

allen-munsch:
### Description of the Change

- `amber upgrade`
- `amber upgrade v0.10.0`

### Alternate Designs

i'm not sure if something external would be better suited?

### Benefits

helps centralize amber version management

### Possible Drawbacks

maintenance, documentation, handling multiple versions currently might be out of scope.

it's more of a discussion pull request, but i'd be more than willing to take a crack at a working demo, plus documentation, etc.

### DIFF
--- a/src/amber/cli/commands/upgrade.cr
+++ b/src/amber/cli/commands/upgrade.cr
@@ -1,0 +1,48 @@
+module Amber::CLI
+  class_property color = true
+
+  class MainCommand < ::Cli::Supercommand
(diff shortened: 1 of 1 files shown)


allen-munsch:
Although after reading more of the documentation and getting more familiar with crystal, it looks like this might be something that `shard.yml` is all about.

Now, I'm thinking maybe just a bit of documentation might be okay.

PR #975, opened 2018-11-05
Update TFB to round 17

faustinoaq:
### Description of the Change

This PR updates Amber's README to include latest TFB round 17, see: https://www.techempower.com/blog/2018/10/30/framework-benchmarks-round-17/

### Alternate Designs

Remove benchmark section? :sweat_smile: 

### Benefits

Latest benchmarks

### Possible Drawbacks

Are benchmarks still relevant for amber users?

I like to see amber (crystal) performance being tested, WDYT?


drujensen:
@waghanza Agree.  

@faustinoaq Let's also filter out the MySQL, so only show PostgresQL to keep it apples to apples comparison.  If someone wants to add Amber-Pg and Amber-Mysql down the road, that would be cool.

https://www.techempower.com/benchmarks/#section=data-r17&hw=ph&test=json&c=6&d=b&o=e

PR #984, opened 2018-11-21
add some polyfills that may be required by Android WebView

damianham:
### Description of the Change

Add some polyfill functions to Object and Array that may be missing in some versions of Android WebView.  One such polyfill is used to convert a collection of HTMLNodes to an array (the HTMLnode collection may not support forEach) in the document on load event listener to attach click handler to nodes with data-method='delete'.

The Date.prototype.toGranite function was moved from the app template to a common polyfill file so that we can manage polyfills one one place. 

### Alternate Designs
None considered
### Benefits
Android WebView may not support some ES2015 functions 
Object.entries
Object.assign
Array.find
Array.from

This change adds these very useful functions as polyfills if they don't exist
### Possible Drawbacks
These polyfills will be redundant when all javascript engines support them


drujensen:
@damianham Why would we include polyfills directly like this?  There are lots of npm packages that handle this already i.e. https://babeljs.io/docs/en/babel-polyfill 

I am concerned this will conflict with other polyfills and I don't think developers would expect this to come out-of-the-box since we are not a JS framework.

drujensen:
@damianham WDYT of #988 instead?

damianham:
#988 makes more sense so I suggest closing this issue in favour of #988 which addresses the specific instance used by Amber.  Developers can provide their own polyfill solutions for other missing functions.

PR #985, opened 2018-11-22
Make optional params truly optional

anamba:
### Description of the Change / Benefits

1. Make optional params truly optional (either key is absent, or key is present but value is "")

This means that the optional param can be present on a form, but it is okay if the user does not fill it in. (Previously, it was only optional in the sense that it was considered okay if the param was not present on the form.)

2. Make block for optional params optional (no need to pass { true })

This always seemed weird to me. I want to get the value of this param from the validated params object, but I don't really care what it looks like, or whether it is blank.

### Possible Drawbacks

It doesn't break anything coded in the previous style, so there shouldn't be any (aside from the code bloat from adding these options).

### DIFF
--- a/spec/amber/validations/params_spec.cr
+++ b/spec/amber/validations/params_spec.cr
@@ -52,8 +52,19 @@ module Amber::Validators
 
       context "optional params" do
         context "when missing" do
-          it "is valid and there is no errors" do
+          it "is valid and there is no errors when param not present" do
             http_params = params_builder("last_name=&middle=j")
+            validator = Validators::Params.new(http_params)
+
+            validator.validation do
+              optional(:name) { |v| !v.nil? }
+            end
+
+            validator.valid?.should be_true
+            validator.errors.size.should eq 0
+          end
+          it "is valid and there is no errors when param present, but blank" do
+            http_params = params_builder("name=&last_name=&middle=j")
             validator = Validators::Params.new(http_params)
 
             validator.validation do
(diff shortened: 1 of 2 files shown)


drujensen:
@anamba This PR seems to change the meaning of `optional` to mean "validate if the field value is nil or blank".

If I understand the original intent, `optional` meant "run this validation block if the field exists".  The validation block will determine if its value is valid or not. 

@eliasjpr Can you take a look at this and chime in? I think the `optional` keyword may be confusing.

anamba:
Here's how I think of required and optional parameters.

The main reason to specify required and optional parameters is to get them back from the params object after validation is done. Sort of like ActionController permitted parameters. You get them back all in one safe hash, to hand off to the model/repo to be persisted to the database.

"Required" means that the (1) key needs to be present on raw_params, (2) the block needs to be present, and (3) and the block needs to return true.

By contrast, "Optional" means that we are still interested in receiving whatever value the user happens to enter on the form, but if they don't provide a value (or the form didn't have a field for it in the first place) we don't want to block the save.

If we do receive a (non-blank) value, then we can see if there is a block for this rule:
* If no block is provided, then that means we don't need to validate (same as passing `{ true }`)
* If a block is provided, then it should return true

drujensen:
@anamba That makes sense.  I imagine this is what the original intent was.  Maybe we should rename `optional` to `permitted`?  

Also, should we  allow for the block to be blank for `required` and assume it returns true as well?

anamba:
Not a bad idea to remove the block requirement for "required" as well. Most of the time, my required block is just `{ |f| !f.blank? }` anyway. So if we make the required check more similar to how the optional check now works (presence of key + value not blank) it would be more consistent, and also allow you to omit the block if you don't need anything fancier than that.

Should I update this branch/pull request or would it be better to start a new one?

drujensen:
@anamba A new PR would be best.  Leave it as `optional`.  Thanks!

PR #988, opened 2018-11-28
replace forEach with generic js loop

drujensen:
### Description of the Change

Replace `forEach` in favor of javascript loop.  I did this because some older browsers do not support `forEach` yet.

### Alternate Designs

polyfill the `forEach` function.  This may cause conflicts with other polyfill packages and create unexpected behavior.

### Benefits

Better compatibility with older browsers

### Possible Drawbacks

None that I am aware of


assets/js/amber.js
```diff
@@ -232,23 +232,47 @@ module.exports = {
...
-                input.setAttribute("value", "DELETE");
-                form.appendChild(input);
-                document.body.appendChild(form);
-                form.submit();
-            }
-            return false;
-        })
+  var elements = document.querySelectorAll("a[data-method='delete']");
+  var i;
+  for (i = 0; i < elements.length; i++) {
```
marksiemers: Is it preferred to have the declaration outside the loop?
What about:
```suggestion
  for (var i = 0; i < elements.length; i++) {
```

PR #989, opened 2018-12-04
Uppercase some scaffolded links

dwightwatson:
### Description of the Change

This makes some very minor changes to the templates used in the scaffolds. There was some inconsistency between links/buttons with a capital first letter and some that were all lowercase.

* This gives all the links/buttons in the resource scaffold a capital letter.
* This renames the "read" link to "show".

### Alternate Designs

I understand that "read" fits in with the CRUD, but if we are going that route we should probably rename "new" to "create" to line up. I think "show" makes much better sense in this context, especially to an end-user.

### Benefits

Just a little more polish in the scaffold which is a nicer experience.

### Possible Drawbacks

I suppose if you were running integration tests over your scaffolds that expected lowercase links then you would need to consider that going forward. It won't affect any existing scaffolds.


drujensen:
@dwightwatson 💯

PR #990, opened 2018-12-04
Improve use of Bootstrap grid, and default home page

dwightwatson:
### Description of the Change

This improves some use of the Bootstrap grid in the out-of-the-box template. There's some unnecessary things going on, like flash messages being nested in a `.row` and `.col-sm`, when they could be left without.

The existing template laid all the content in a base `.row` which unless all content is contained within columns then it's going to sit outside the main container. A fresh Amber install in fact has the logo sitting outside the container to the left. This adjust the home page content to work with it's row, centering the content.

The logo is now put on the page using an `img` tag which saves some CSS which is likely to be removed anyway. In addition I've updated the logo to match the preferred logo as suggested on the media page.

image

### Alternate Designs

For reference, here's the existing home page with Chrome's inspector showing that the logo sits outside of the container.

image

### Benefits

The default app uses the Bootstrap grid properly and content is contained correctly within the container. Developers have greater flexibility to use `.row` how they want without being in one by default.

Also means we ship with less custom CSS which probably going to be removed anyway, and the proper logo is used.

### Possible Drawbacks

If developers really liked being in a `.row` from the main template then they'll need to add it back. I don't expect that this is a common use-case though.


robacarp:
Thanks for this @dwightwatson, much appreciated.

PR #996, opened 2018-12-19
Configure watch via .amber.yml

anamba:
### Description of the Change

Allow user to configure `amber watch` from `.amber.yml`. Configuring this way is optional (there is still a default config, and it is the same as before), and you can continue to use the command line switches if you want (if you have a config in `.amber.yml`, command line switches will override it).

A sample `.amber.yml`:

```yml
type: app
language: ecr
model: granite

# list of tasks to be run by `amber watch`
watch:
  # NOTE: names that match crystal commands are special (e.g. run, spec)
  run:
    # commands will be joined with && (join them yourself if need || or ;)
    build_commands:
      - mkdir -p bin
      - crystal build ./src/rcwebcr.cr -o bin/rcwebcr
    run_commands:
      - bin/rcwebcr
    include:
      - ./config/**/*.cr
      - ./src/**/*.cr
      - ./src/**/*.ecr
    exclude: # NOTE simplistic implementation: (1) enumerate all includes and excludes; (2) return (includes - excludes)
      - ./src/dont_restart_when_i_am_changed.cr
  spec:
    run_commands:
      - crystal spec
    include:
      - ./spec/**/*.cr
    exclude:
      - ./spec/dont_run_specs_when_i_am_changed.cr
  npm:
    build_commands:
      - npm install --loglevel=error
    run_commands:
      - npm run watch
```

### Alternate Designs

The config is too complex to configure solely via command line switches/arguments, and leaving it non-user-configurable is creating unhappiness among users.

### Benefits

Developers are now free to choose `yarn` over `npm`, or even nothing at all. Additionally, they can use `amber watch` to run other processes as well.

### Possible Drawbacks

Mainly some code bloat. It's certainly possible that this is not something the amber project should handle directly.

### DIFF
--- a/src/amber/cli/commands/watch.cr
+++ b/src/amber/cli/commands/watch.cr
@@ -20,8 +20,6 @@ module Amber::CLI
 
       def run
         CLI.toggle_colors(options.no_color?)
-        options.watch << "./config/**/*.cr"
-        options.watch << "./src/views/**/*.slang"
         super
       end
     end
--- a/src/amber/cli/config.cr
+++ b/src/amber/cli/config.cr
@@ -1,20 +1,33 @@
 module Amber::CLI
   def self.config
     if File.exists? AMBER_YML
-      Config.from_yaml File.read(AMBER_YML)
+      begin
+        Config.from_yaml File.read(AMBER_YML)
+      rescue ex : YAML::ParseException
+        logger.error "Couldn't parse #{AMBER_YML} file", "Watcher", :red
+        exit 1
+      end
     else
       Config.new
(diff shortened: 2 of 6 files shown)


robacarp:
@anamba thanks for doing this, it's long overdue and will be a welcome change to the workflow.

> It's certainly possible that this is not something the amber project should handle directly.

This is a great thing to discuss. Can you explain what you mean by that? Are there existing tools which Amber can wrap or depend on that you have experience with? I think previously Amber depended on another shard, which I believe was abandoned/absorbed.

Issue #997, opened 2018-12-23
Sqlite cli issue


Willyboar:
If you try to create a new app using sqlite3 with  
`amber new app blog -d sqlite3` 
cli doesn't prevent you and giving false requirements at database.cr
`require "granite/adapter/sqlite3"` instead of 
`require "granite/adapter/sqlite"`

PR #1000, opened 2018-12-26
Provide supported options in CLI

dwightwatson:
### Description of the Change
This is meant to support the issue described in #997 where the new app generator would accept arguments that weren't valid - like `sqlite3` instead of the supported `sqlite`. It does this using the `any_of` of the [underlying CLI library](https://github.com/mosop/cli#shell-completion). It appears that this also populates the help text, so no need to include the options there.

### Alternate Designs
We could manually perform argument validation or add some sort of callback mechanism to perform pre-command logic, but it seemed better to utilise functionality already provided by an existing dependency.

### Benefits
Users will be forced to provide valid arguments and cannot generate an app that will not operate.

### DIFF
--- a/src/amber/cli/commands/new.cr
+++ b/src/amber/cli/commands/new.cr
@@ -7,12 +7,12 @@ module Amber::CLI
     class New < Command
       class Options
         arg "name", desc: "name/path of project", required: true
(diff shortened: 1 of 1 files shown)


drujensen:
💯

Issue #1004, opened 2018-12-27
Hide welcome page logo on small screen


faustinoaq:
### Description

Welcome page layout was improved on https://github.com/amberframework/amber/pull/990, although the logo seems a bit weird on small screens.

![vokoscreen-2018-12-27_13-25-48](https://user-images.githubusercontent.com/3067335/50490840-46b1f180-09dd-11e9-97cf-ba19f2792c37.gif)

### Steps to Reproduce

1. amber new project --deps
2. amber watch (wait build)
3. Go to 0.0.0.0:3000
4. Resize your window to 600x400

**Expected behavior:** Read all Welcome page without scrolling

![ezgif-3-6a87aec612](https://user-images.githubusercontent.com/3067335/30362934-e1ed8826-9823-11e7-84e1-11cf9e2471e8.gif)

**Actual behavior:** Need to scroll to see all welcome page because amber logo is mislocated

**Reproduces how often:** 100%

### Versions

Amber v11.2

### Additional Information

Related PRs & issues:

- https://github.com/amberframework/amber/pull/990
- https://github.com/amberframework/amber/pull/257
- https://github.com/amberframework/amber/issues/256

Ping: @dwightwatson

dwightwatson:
As part of that PR one of the things I wanted to improve was the amount of boilerplate CSS the framework ships with. There are already some custom styles for the navigation bar which are used by the scaffolds, but it seems silly having `#logo` specific styles when most apps are just going to have to remove it. To be honest, I'd actually like to keep working on that so Amber ships with the least front-end boilerplate possible.

So I suppose there's a couple of options (off the top of my head):

- Revert and bring back the custom CSS,
- Make the logo a little smaller so it takes up less space vertically, or
- Remove the logo entirely.

If I had to re-think it now I'd personally consider just removing the logo. It's nice to have there - but anyone starting with Amber already knows what the logo looks like, it's still there as a favicon, and it's just more files they need to remove to get to a clean slate.

Let me know what you think, happy to sort this out.

faustinoaq:
Hi @dwightwatson Thank you for your comment! :+1: 

> Remove the logo entirely.

I think this would be a nice idea, @amberframework/contributors WDYT?

drujensen:
I agree. The less boilerplate to remove, the better.

Sija:
> Make the logo a little smaller so it takes up less space vertically, or

[...]

PR #1005, opened 2018-12-28
Remove logo from app template

dwightwatson:
### Description of the Change
In reference to #1004 - this removes the logo from the app template and the home page views.

### Alternate Designs
These were discussed in #1004.

### Benefits
Solves an issue where the logo would be too large on smaller screens, and removes boilerplate that most developers will remove anyway.

### Possible Drawbacks
People that really like looking at the logo might be disappointed.


faustinoaq:
> I can't remember for Phoenix, but at least for Rails the home page it ships with isn't actually in the app template - it's just shown if there are no routes registered

Well, I just tried rails and phoenix:

vokoscreen-2018-12-28_08-21-32

vokoscreen-2018-12-28_08-31-55

I think Amber can use some responsive design here :sweat_smile: 

vokoscreen-2018-12-28_08-37-34

drujensen:
@dwightwatson Thanks for doing this!  I like using the bootstrap styles as much as possible so using the one built in would be great.  I think just the logo is fine but would like to hear what others think.

dwightwatson:
I've got this working and looking good, but I might wait until #1007 goes in, as I'll leverage the $primary brand colour that's provided to Bootstrap.

image

dwightwatson:
Yay, thanks!

Was just working on updating the existing generators to output links that are compatible with the Bootstrap navbar too. I'll get that into another PR shortly.

elorest:
Everyone else just has a larger logo that only shows up on the default index page. Why are we choosing to have a small logo in the default layout for all pages?

PR #1007, opened 2018-12-28
Load Bootstrap through Webpack

dwightwatson:
### Description of the Change
This adjusts the main template to load Bootstrap and its dependencies through NPM and Webpack - rather than using a CDN in the main application template.

* Sets a precedent for new users to prefer to use NPM and serve assets through a single bundle, rather than loading up additional dependencies through CDNs.
* It removes some additional boilerplate surrounding heading styles and containers and just leaves it up to Bootstrap. I use the existing brand colour variable as the `$primary` colour.
* For developers who do want to continue using Bootstrap it puts them in a good place to customise any variables or only import the parts of Bootstrap they want.
* For developers who don't want to use Bootstrap there is a *tad* more to remove in that they need to remove the references from the main CSS and JS files, as well as ditch the dependencies from `package.json`.

My main thought here is just to encourage developers to just adopt the Webpack pipeline and make it really easy to continue with Bootstrap if they decide to do so.

### Alternate Designs
Some alternate options:
* Leave Bootstrap as-is, no changes.
* Leave Bootstrap as-is, but reduce boilerplate CSS code so developers start with a cleaner slate. This could mean adopting Bootstrap's navbar component so we don't need to ship the custom `.masthead` component styles.
* Remove Bootstrap and ship with some bare-bone styles.

My general preference is to stick with Bootstrap as it's great for when you just want to experiment with something and it gives you a solid building block. However I totally appreciate that many will feel different about it.

### Benefits
Less CSS boilerplate, less reliance on CDNs (I actually came across this because I was playing with an Amber project while on a plane, nothing worked because I had no internet access and nothing could load) and means Amber apps will ship by default with a single CSS and JS file out of the box. Developers will be guided towards using NPM/yarn and Webpack to build their assets.

### Possible Drawbacks
People who don't like Bootstrap or don't intend to use it on a project will need to spend another minute to clear the boilerplate.

### DIFF
--- a/src/amber/cli/templates/app/src/assets/javascripts/main.js
+++ b/src/amber/cli/templates/app/src/assets/javascripts/main.js
@@ -1,4 +1,2 @@
-import 'jquery';
-import 'popper.js';
 import 'bootstrap';
 import Amber from 'amber'
--- a/src/amber/cli/templates/app/src/assets/javascripts/main.js
+++ b/src/amber/cli/templates/app/src/assets/javascripts/main.js
@@ -1,1 +1,4 @@
+import 'jquery';
+import 'popper.js';
+import 'bootstrap';
 import Amber from 'amber'
--- a/src/amber/cli/templates/app/package.json.ecr
+++ b/src/amber/cli/templates/app/package.json.ecr
@@ -23,5 +23,10 @@
     "webpack": "^3.12.0",
     "webpack-merge": "^4.1.2",
     "extract-text-webpack-plugin": "^3.0.2"
+  },
+  "dependencies": {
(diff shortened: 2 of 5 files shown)


src/amber/cli/templates/app/src/assets/javascripts/main.js
```diff
@@ -1 +1,4 @@
+import 'jquery';
+import 'popper.js';
```
robacarp: I think bootstrap will import jquery and popper itself, so you don't have to include these explicitly.

dwightwatson: Easy, updated.

PR #1009, opened 2018-12-30
Dj/amber extension

drujensen:
### Description of the Change

This change is to create a clear API for code generators.  This creates a base class `generator` that provides a foundation for others to create their own.   This also separates the generators from the templates.

The goal is to move the `auth` and `crecto` generators to their own shards.  This is the first step in being able to do that.

This also changes the options flag `--deps` to `--no-deps` and will automatically run `shards update`.  This makes the `new` command closer to rails.

### Alternate Designs

The current recipes is another approach to handling code generation. However, it doesn't allow you to extend amber and provide new generators, it only supports existing ones.

### Benefits

Ability to create your own generator.  There is still a hurdle of registering a generator, but I think this can be achieved with macros. 

### Possible Drawbacks

There is a lot of code changes in this PR.


src/amber/cli/generators/generator.cr
```diff
@@ -0,0 +1,81 @@
...
+      end
+    end
+
+    private def parse_fields(params : Array(String)?)
+      if params
+        fields = params.map { |field| Field.new(field, database: config.database) }
+      else
+        fields = [] of Field
+      end
+      return fields
```
eliasjpr: Is the return necessary?

src/amber/cli/generators/generator.cr
```diff
@@ -0,0 +1,81 @@
...
+    private def parse_fields_hash
+      fields_hash = Hash(String, String).new
+      @fields.reject(&.hidden).each do |f|
+        field_name = f.reference? ? "#{f.name}_id" : f.name
+        fields_hash[field_name] = default_value(f.cr_type) unless f.nil?
+      end
+      fields_hash
+    end
+
+    private def default_value(field_type)
```
eliasjpr: Should default values be give as a definition param instead of assumed by the generator?

src/amber/cli/recipes/recipe.cr
```diff
@@ -57,7 +57,7 @@ module Amber::Recipes
           app = App.new(name, options.d, options.t, options.m, recipe)
           app.fetch_recipe(directory)
           app.render(directory, list: true, color: true)
-          if options.deps?
```
eliasjpr: positive checks are more readable, no?

src/amber/cli/generators/generator.cr
drujensen: That would be a nice feature but not sure how you would do that.  This is used by the specs.  It would be cool if we had a `Faker` shard instead.

src/amber/cli/recipes/recipe.cr
drujensen: `if !option.no_deps?` is a double negative which I find harder to read myself.  This is one I think its acceptable to use the `unless`

drujensen:
@elorest I'm free tomorrow.  I will ping you.

PR #1018, opened 2019-01-01
Ability to bootstrap a project into an Amber project.

drujensen:
### Description of the Change

This change allows you to generate a project in the current directory.

It also removes `crecto` from the templates and cleans up the generators.

This changes the default behavior to enable interactive mode.  You can pass a flag `-y` or `--assume-yes` to disable it.

### Alternate Designs

This will allow bootstrapping an existing project.

### Benefits

This will allow you to initialize an existing project into an amber one.

### Possible Drawbacks

none


drujensen:
@elorest @robacarp @Blacksmoke16 @damianham  I finally had a couple hours to finish this up.  It kinda grew since I noticed we removed Crecto so this cleans up the generators and templates.  If you have a sec, can you review?  Thanks!

drujensen:
Anyone else available to review?  I know it's a big PR so if you want me to break it up, let me know.  Most of it is just removing the Crecto templates and generators.  This allowed us to collapse the file structure down.

src/amber/cli/commands/generate.cr
```diff
@@ -12,6 +12,7 @@ module Amber::CLI
         arg "name", desc: "name of resource", required: false
         arg_array "fields", desc: "user:reference name:string body:text age:integer published:bool"
         bool "--no-color", desc: "disable colored output", default: false
+        bool ["-y", "--assume-yes"], desc: "Assume yes to disable interactive mode", default: false
```
robacarp: 👍

src/amber/cli/commands/new.cr
```diff
@@ -22,24 +23,29 @@ module Amber::CLI
 
       def run
         CLI.toggle_colors(options.no_color?)
-        full_path_name = File.join(Dir.current, args.name)
+        if args.name == "."
+          name = Dir.current.split("/").last
```
robacarp: I wasn't expecting to see manually splitting the name here and using basename below. Is there a reason for the difference?

src/amber/cli/commands/new.cr
```diff
@@ -22,24 +23,29 @@ module Amber::CLI
...
         else
-          generator = Generators.new(name, "./#{args.name}")
+          generator = Generators.new(name, full_path_name)
         end
-        generator.generate("app", options)
+        generator.generate_app(options)
 
         # Encrypts production.yml by default.
-        cwd = Dir.current; Dir.cd(args.name)
+        cwd = Dir.current; Dir.cd(full_path_name)
```
robacarp: unrelated to your code, but I recently discovered Dir&cd takes a block to do this pop/push thing automatically.

src/amber/cli/generators.cr
```diff
@@ -27,23 +27,21 @@ module Amber::CLI
...
-          info "Rendering App #{name} in #{directory}"
-          App.new(name, options.d, options.t).render(directory)
-          unless options.no_deps?
-            info "Installing Dependencies"
-            Helpers.run("cd #{directory} && shards update")
-          end
-        end
+    def generate_app(options)
+      info "Rendering App #{name} in #{directory}"
+      App.new(name, options.d, options.t).render(directory, list: true, interactive: !options.assume_yes?, color: options.no_color?)
```
robacarp: > `!options.assume_yes?`

Maybe this is evidence that there's some drift between interfaces with `interactive` and `options`

src/amber/cli/commands/new.cr
drujensen: This allows you to bootstrap an existing project using `amber new .` I decided to grab the name of the project from the current directory.  Maybe there is a better way of getting the name?

src/amber/cli/commands/new.cr
drujensen: I will leave that for someone else to refactor

src/amber/cli/generators.cr
drujensen: We default to `false` so I think this makes sense.  `apt-get` uses this `-y` and `--assume-yes` flags, so it seemed to match.

src/amber/cli/commands/new.cr
robacarp: Sure, what I was wondering is why `File.basename Dir.current` doesn't work

src/amber/cli/generators.cr
robacarp: Yeah, not a to-do, just an observation that some tech debt might be stacking

src/amber/cli/commands/new.cr
drujensen: Aaah, gotcha.  I will fix that.

elorest:
👍

PR #1021, opened 2019-01-05
Don't present optional param if not present input params

### LINKED ISSUE #1019: Params validation should not present optional field when it not present in the input
### Description

Params validation should not present the optional field when it not present in the input.

### Steps to Reproduce

The following is an example:
```crystal
def user_params
  params.validation do
    optional :age
    required :name
  end
end
```

When use it validate the input `name=Don`, it returns:

```crystal
{"name" => "Don", "age" => nil}
```

**It will OVERWRITE the `age` field in the database.**

**Expected behavior:** [What you expect to happen]

Expect it returns:
```crystal
{"name" => "Don"}
```

**Actual behavior:** [What actually happens]
It returns:

```crystal
{"name" => "Don", "age" => nil}
```

Always.

### Versions

Amber 0.11.2
Crystal 0.27.0

cpunion:
Fix #1019

### DIFF
--- a/src/amber/validators/params.cr
+++ b/src/amber/validators/params.cr
@@ -7,14 +7,17 @@ module Amber::Validators
     getter predicate : (String -> Bool)
     getter field : String
     getter value : String?
+    getter present : Bool
 
     def initialize(field : String | Symbol, @msg : String?, @allow_blank : Bool = true)
       @field = field.to_s
+      @present = false
       @predicate = ->(_s : String) { true }
     end
 
     def initialize(field : String | Symbol, @msg : String?, @allow_blank : Bool = true, &block : String -> Bool)
       @field = field.to_s
+      @present = false
       @predicate = block
     end
 
@@ -29,6 +32,7 @@ module Amber::Validators
 
     private def call_predicate(params : Amber::Router::Params)
       @value = params[@field]
+      @present = !!params.has_key?(@field)
       @predicate.call params[@field] unless @predicate.nil?
     end
 
@@ -127,7 +131,7 @@ module Amber::Validators
           @errors << rule.error
         end
 
-        @params[rule.field] = rule.value
+        @params[rule.field] = rule.value if rule.present
       end
 
       errors.empty?
--- a/src/amber/validators/params.cr
+++ b/src/amber/validators/params.cr
@@ -32,7 +32,7 @@ module Amber::Validators
 
     private def call_predicate(params : Amber::Router::Params)
       @value = params[@field]
-      @present = !!params.has_key?(@field)
+      @present = params.has_key?(@field)
       @predicate.call params[@field] unless @predicate.nil?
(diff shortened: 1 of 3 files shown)


cpunion:
I'm not sure what means `optional` with `allow_blank: true`. I think `optional :name` will discard the blank value, and `optional :name, allow_blank: true` will keep the blank value? I need fix this.

drujensen:
@eliasjpr Any concerns with this fix?

src/amber/validators/params.cr
```diff
@@ -29,6 +32,7 @@ module Amber::Validators
 
     private def call_predicate(params : Amber::Router::Params)
       @value = params[@field]
+      @present = !!params.has_key?(@field)
```
drujensen: what is the `!!` for?

cpunion: `Amber::Router::Params#has_key?` returns `(String | Nil)`, `!!` converts to `Bool`.

See https://github.com/crystal-lang/crystal/blob/c9d1eef8fd/src/hash.cr#L147

cpunion: Oh. I think `Amber::Router::Params#has_key?` should be fixed to check the key exists.

cpunion:
Need more time to think about `allow_blank` of `optional`.

src/amber/validators/params.cr
eliasjpr: I like this and I think this is fine for now. I have opened this issue https://github.com/amberframework/amber/issues/1026 to refactor the `has_key?(key)` to return `Bool` to be consistent with Crystal.

eliasjpr:
@cpunion I was wondering if you can give this library a test drive since you are already on this scope of validations. Let me know if you do and if you see any issues with it.

src/amber/validators/params.cr
eliasjpr: You can make the change to the method instead of adding `!!` here https://github.com/amberframework/amber/blob/6c2889bafa57d44d1951a2dec60a519af2fad87a/src/amber/router/params.cr#L45

cpunion:
There are some other cases.

**1. `has_key?` can't accept null value by JSON format**

See the following methods:

https://github.com/amberframework/amber/blob/6c2889bafa57d44d1951a2dec60a519af2fad87a/src/amber/router/params.cr#L30-L33

https://github.com/amberframework/amber/blob/6c2889bafa57d44d1951a2dec60a519af2fad87a/src/amber/router/params.cr#L44-L46

Sometimes we pass a null value with a key by JSON format, so `has_key?` should correct for this.

**2. Params only handle `Hash(String, String)`, not full JSON support.**

----
I think those are not easy. We can solve this PR first, my project is blocked by this. I will refactor `has_key?` now.

cpunion:
@eliasjpr Great!

PR #1022, opened 2019-01-06
Register generators

drujensen:
### Description of the Change

This change removes the large case statement and allows the generator to
register itself.  Eventually, we will (hopefully) allow others to easily
create generators and register them with Amber.

### DIFF
--- a/src/amber/cli/generators/auth.cr
+++ b/src/amber/cli/generators/auth.cr
@@ -1,116 +1,19 @@
-require "./generator"
-require "./field"
-
 module Amber::CLI
   class Auth < Generator
-    property fields : Array(Field)
+    command :auth
+    property auth : Generator
 
     def initialize(name, fields)
-      super(name, nil)
-      @fields = all_fields(fields)
-    end
-
-    def pre_render(directory)
-      add_routes
-      add_plugs
-      add_dependencies
-      inject_application_controller_methods
-    end
-
-    private def add_routes
-      add_routes :web, <<-ROUTES
-        get "/profile", #{class_name}Controller, :show
-          get "/profile/edit", #{class_name}Controller, :edit
-          patch "/profile", #{class_name}Controller, :update
-          get "/signin", SessionController, :new
-          post "/session", SessionController, :create
-          get "/signout", SessionController, :delete
-          get "/signup", RegistrationController, :new
-          post "/registration", RegistrationController, :create
-      ROUTES
-    end
-
-    private def add_plugs
-      add_plugs :web, <<-PLUGS
-        plug Authenticate.new
-      PLUGS
-    end
-
-    private def add_dependencies
-      add_dependencies <<-DEPENDENCY
-      require "../src/models/**"
-      require "../src/pipes/**"
-      DEPENDENCY
-    end
-
-    private def inject_application_controller_methods
-      filename = "./src/controllers/application_controller.cr"
-      controller = File.read(filename)
-      append_text = ""
-
-      unless controller.includes? "property current_#{@name}"
-        append_text += current_method_definition
-      end
-
-      unless controller.includes? "def signed_in?"
-        append_text += signed_in_method_definition
-      end
-
-      unless controller.includes? "def redirect_signed_out_#{@name}"
-        append_text += redirect_signed_out_method_definition
-      end
-
-      append_text = "#{append_text}\nend\n"
-      controller = controller.gsub(/end\s*\Z/, append_text)
-      File.write(filename, controller)
-    end
-
-    private def current_method_definition
-      <<-AUTH
-
-        def current_#{@name}
-          context.current_#{@name}
-        end\n
-      AUTH
-    end
-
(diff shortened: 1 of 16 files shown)


src/amber/cli/generators/scaffold.cr
```diff
@@ -1,5 +1,6 @@
 module Amber::CLI
   class Scaffold < Generator
+    command :scaffold
```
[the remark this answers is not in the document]
drujensen: Yes, that was the thought.  This will (eventually) be a public API so I wanted to make this as clean as possible.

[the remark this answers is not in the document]
elorest: Fair enough. If someone needs to do something differently they can always look at what the macro is doing. :)

src/amber/cli/generators/auth_base.cr
```diff
@@ -0,0 +1,116 @@
+require "./generator"
+require "./field"
+
+module Amber::CLI
+  class AuthBase < Generator
```
eliasjpr: AuthGenerate?

eliasjpr: Maybe AuthGenerator?

src/amber/cli/generators/auth_base.cr
```diff
@@ -0,0 +1,116 @@
...
+        def current_#{@name}
+          context.current_#{@name}
+        end\n
+      AUTH
+    end
+
+    private def signed_in_method_definition
+      <<-AUTH
+
+        def signed_in?
```
eliasjpr: `def #{@name}_signed_in?` no?

src/amber/cli/generators/mailer.cr
```diff
@@ -2,6 +2,7 @@ require "./field.cr"
 
 module Amber::CLI
   class Mailer < Generator
+    command :mailer
```
eliasjpr: I think this is what @elorest means `class_getter registered_commands = :mailer`

src/amber/cli/generators/generator.cr
```diff
@@ -3,6 +3,12 @@ require "./field.cr"
 
 module Amber::CLI
   abstract class Generator < Teeplate::FileTree
+    class_getter registered_commands = Hash(String, Amber::CLI::Generator.class).new
+
+    macro command(name)
```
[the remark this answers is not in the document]
Sija:
```cr
abstract class AuthBase
  #...

  macro inherited
    {% factory_id = @type.name.gsub(/^Amber::CLI::Generator::/, "").underscore %}

    ::Amber::CLI::Generator.registered_commands[{{factory_id.stringify}}] = self
  end
end
```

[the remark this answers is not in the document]
drujensen: @Sija yeah, i thought of doing that but currently some of the generators are abstract base classes and would cause an error if someone tried to run them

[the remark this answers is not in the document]
Sija: There's still a limitation regarding naming the classes (`Amber::CLI::Generator::<CommandName>`), which doesn't exists for explicit declaration, so maybe it's better to keep the latter one avoiding potential confusion... just thinking out loud...

src/amber/cli/generators/mailer.cr
drujensen: @eliasjpr take a look at the conversation with @Sija .  We could simplify this using the class name as the generator name instead of using this macro.  I kinda like explicitly specifying the command.  I was also thinking of adding some type of description for help.

src/amber/cli/generators/auth_base.cr
drujensen: I prefer base myself.  There is an Auth that provides the command and picks granite or crecto.  When we move that out to its own shard (if that is possible) we will eliminate the need for this base class.

src/amber/cli/generators/auth.cr
```diff
@@ -1,116 +1,19 @@
...
-
-      unless controller.includes? "def signed_in?"
-        append_text += signed_in_method_definition
-      end
-
-      unless controller.includes? "def redirect_signed_out_#{@name}"
-        append_text += redirect_signed_out_method_definition
+      super(name, fields)
+      if config.model == "crecto"
+        @auth = CrectoAuth.new(name, fields)
```
eliasjpr: `def initialize(name, fields, orm : Class)` to make the class more flexible and extendable. Right now if someone would like to add auth generator for another orm (Think Jeniffer, LuckyRecord) will need to change this class.

src/amber/cli/generators/auth_base.cr
eliasjpr: I asked cause I saw Auth and AuthBase

src/amber/cli/generators/auth.cr
drujensen: This will be going away completely when we move to shards.  Each shard will either override the commands or they will provide new ones.

Issue #1014, opened 2018-12-30
Break up Amber::Router::Cookies module


dwightwatson:
### Description
The file for `Amber::Router::Cookies` contains multiple class and module definitions, and `Amber::Router::Cookies` may not even need to be a class itself. I think it would be beneficial for the classes and modules defined in the namespace to be broken up into individual files so that it's easier to find.

I suspect that there may be other instances of file/class inconsistencies that would be good to address as well. 

* `Amber::Router::Session::Store` located in `amber/router/session/session_store.cr`
* `Amber::Pipe` located in `amber/pipes` directory

drujensen:
@eliasjpr wdyt?

eliasjpr:
I thought of doing that previously. I think is a good idea

dwightwatson:
Was looking at renaming `Amber::Pipe` to `Amber::Pipes` versus renaming the directory to `amber/pipe`.

There might be room for discussion around naming and seeing if we can develop some consistency between the module names.

![image](https://user-images.githubusercontent.com/1100408/50810631-aa81d600-135e-11e9-94e8-2cb05342e1c2.png)

One option might be to aim for singular names. It may be better to go from `Amber::Validators` to `Amber::Validation`.

```
Amber::CLI
Amber::Controller
Amber::DSL
Amber::Environment
Amber::Exception
Amber::Extension
Amber::Pipe
Amber::Router
Amber::Server
Amber::Support
Amber::Validation
Amber::WebSocket
```

robacarp:
Some will always be singular, server for example. But I think any namespaces/folders which represent a collection of things should be plural. Validations, Extensions, Support, etc. 

Is Support plural? Whatever...

eliasjpr:
I was working on reorganizing the project into modules as in this screenshot, the team was hesitant at the time to make these changes. Not sure how the team feels now, to slowly modularized and document the code

![screenshot from 2019-01-08 20-34-18](https://user-images.githubusercontent.com/1685772/50870190-f6477600-1384-11e9-9c81-f4f1a62d758b.png)

eliasjpr:
I was  in the past proposing the idea to modularized the code further, to something maybe towards [this](https://github.com/eliasjpr/amber/tree/ep/cleanup-router-session/src/amber) I know the team was hesitant to do so then. I am not sure how the team feels now, to slowly organize and document the project so we can generate crystal docs.

drujensen:
[...]

PR #1024, opened 2019-01-07
Break up the cookies classes

dwightwatson:
### Description of the Change
This was done in response to #1014.

* The various cookie classes in the cookie module are broken out into their own files.
* The `ChainedStore` module is just rolled into `Amber::Router::Cookies` where it is used.
* The `SerializedStore` and `JsonSerializer` are removed as there don't appear to be references to them.

I'm sure there is still work that can be done to cleanup this code, but I think this is a good first step.

### Alternate Designs
Leaving it in a single file as it was, or breaking it out to multiple files but keeping the old stuff around as well.

### Benefits
It'll be much easier to reason with and work with the cookie related code. It will also make it easier to attempt to DRY up any of the code that can be.

---

There appear to be test failures - but I'm not sure they are related. Would love a second opinion/point in the right direction. I'm having trouble replicating test failures locally - I don't know if it's something with my setup or something that I'm doing wrong.


eliasjpr:
@dwightwatson the tests failures are related to these changes and it seems that the changes here might have broken the CLI. Try generating a new app with these changes and make sure the app runs successfully.

eliasjpr:
I like the breakdown of the classes and moving in the right direction. Maybe abstract the duplication of methods

dwightwatson:
I've merged master in after #1027 and the suite has gone back to green.

I've broken out an `AbstractStore` which matches the convention for session stores. It defines abstract methods `get` and `set` - which identified that the `PermanentStore` didn't actually conform to the protocol unlike the other stores - so I've updated it appropriately.

src/amber/router/cookies/abstract_store.cr
```diff
@@ -0,0 +1,20 @@
...
+
+    def [](name)
+      get(name)
+    end
+
+    def []=(name, value)
+      set(name, value)
+    end
+
+    abstract def get(name)
```
Sija: I'd use this opportunity to change it to `get?`. This way we have a free `#[]?` and a cheaper way to handle non-existent keys.

src/amber/router/cookies/store.cr
```diff
@@ -0,0 +1,128 @@
...
+    property host : String?
+    property secure : Bool = false
+
+    def initialize(@host = nil, @secret = Random::Secure.urlsafe_base64(32), @secure = false)
+      @cookies = {} of String => String
+      @set_cookies = {} of String => HTTP::Cookie
+      @delete_cookies = {} of String => HTTP::Cookie
+    end
+
+    def self.from_headers(headers)
```
Sija: This seems to be a constructor but instead it returns hash of `HTTP::Cookie`s.

src/amber/router/cookies/abstract_store.cr
dwightwatson: If possible, I'd rather tackle that sort of improvement in another PR. The focus of this one was to break up the classes, remove dead code and have it all still work.

src/amber/router/cookies/store.cr
dwightwatson: To be honest, I still find a lot of this code really confusing. Why is there a base Store class, but then we have other actual implementations of stores? I'd love to address further improvement here in other PR where I suspect we'd have some larger breaking changes.

Sija: Totally agree, that piece of the code screams for refactoring - in another PR that is.

PR #1028, opened 2019-01-07
Prefer JavaScript UMD modules spec

jackturnbull:
### Description of the Change

It appears that Webpack is already handling the UMD import/export specification correctly when compiling the assets within Amber. In the amber.js websocket file, both UMD and common JS are present, and non-identical as both `class Channel` and `class Socket` are exported in the UMD format but only Socket is exported as CommonJS. A side-effect of this is that the `module.exports` object is transpiled into the bundle as Webpack has already defaulted to using UMD.

This commit simply removes the CommonJS export definition.

### Alternate Designs

None / N/A

The CommonJS pattern could be chosen, but with Webpack support for UMD I don't believe this would be worthwhile.

### Benefits

This was spotted as I had changed my JavaScript bundler to [Rollup.js](https://rollupjs.org/guide/en) but the output was evidently not handled as graciously as Webpack's as I had a `module.exports` remaining in the output without `module` itself being defined.

The benefit being that this opens the floor to using other JS bundlers that also contain the same issue.

### Possible Drawbacks

None as far as I can tell. I've stepped through the output of the webpack bundles before and after this change and in both cases both the Socket class and the Channel class are the only two modules passed into the bootstrap function, although if anyone has more context for why the module.exports definitions remain that would help also!

Summary of the output bundles:

* Before
  * Defined on **module**: Socket
  * Defined on **exports**: Socket, Channel
  * Passed to boostrap function: Socket, Channel _(Socket deduped)_
* After
  * Defined on **module**:
  * Defined on **exports**: Socket, Channel
  * Passed to boostrap function: Socket, Channel


jackturnbull:
Waiting on https://github.com/amberframework/amber/pull/1027 to get the tests passing

jackturnbull:
Sorry, appears that the rebase has squashed the revew! Tests passing at least.

PR #1029, opened 2019-01-08
Webpack 4

jackturnbull:
### Description of the Change

Webpack 4 is now being used for both the amber.js websocket library and within project generation itself. There have been a few thing that have needed to be worked around, from least challenging to most:

* All packages updated in both the parent and CLI template package.json files.
* `babel-core` has became `@babel/core`.
* `webpack-cli` is now a required package.
* `babel-preset-env` has became `@babel/preset-env` and is reflected in the webpack config.
* Uglify can be removed as Terser is now built in can can be enabled with a boolean flag.
* `extract-text-webpack-plugin` has since been deprecated and is broken for Webpack 4. Their README recommends migration to `mini-css-extract-plugin`, which holds much of the same functionality. The primary difference being that Webpack 4 no longer graciously allows defining the CSS entry alongside the JS (you can, but it produces an additional unused JS file). The Webpack config has been reworked slightly to work with this new plugin.

### Alternate Designs

Mostly none, although I had considered not using the entry.js file within the webpack config and opting for requiring the stylesheets from within main.js.

Alternatives to `mini-css-extract-plugin` may exist although this has been chosen as it appears to be the accepted migration path.

### Benefits

Webpack 4 and all its charms.

### Possible Drawbacks

I don't have a substantial enough Amber project to test this with so I'm not sure what I'm missing. I suspect any bugs will likely be because of any change in the way webpack generates bundles between v3 and v4, and not the new Amber webpack config - but I'm unable to test.


src/amber/cli/templates/app/config/webpack/common.js
```diff
@@ -20,20 +19,13 @@ let config = {
   module: {
     rules: [
       {
-        test: /\.css$/,
+        test: /\.(sa|sc|c)ss$/,
```
Sija: `/\.(css|sass|scss)$/` plz

Sija:
Other than my one small comment looks goood 👍 (can't test it atm so I'll leave approval to someone else)

src/amber/cli/templates/app/config/webpack/common.js
jackturnbull: Agreed! I'd lifted this from the `mini-css-extract-plugin` readme but I prefer yours!

jackturnbull:
Sorted @sija 🙂

Issue #905, opened 2018-06-30
Unhandled exception on recipes with templates using .liquid extension


faustinoaq:
### Description

Hi, I'm creating a Vue.js recipe https://github.com/faustinoaq/vue-amber-recipe (not published yet).

And I'm using `.liquid` extension for my recipe templates

![image](https://user-images.githubusercontent.com/3067335/42126826-5ad56038-7c53-11e8-81fe-b4e2fb5740bd.png)

### Steps to Reproduce

```
➜  ~ mkdir -p myrecipe/app
➜  ~ touch myrecipe/app/shard.yml.liquid
➜  ~ amber new myapp -r myrecipe
```

**Expected behavior:**

(should be fine using `.lqd` or `.liquid` extensions)

```
➜  ~ amber new myapp -r myrecipe
10:56:57 Generate   | (INFO) Rendering App myapp in ./myapp from myrecipe
10:56:57 Generate   | (INFO) new       shard.yml
Environment file not found for ./config/environments/production
➜  ~ tree myapp                                       
myapp
└── shard.yml

0 directories, 1 file
```

BTW, not idea why `Environment file not found` message is printed :sweat_smile: 

**Actual behavior:**

```
10:53:39 Generate   | (INFO) Rendering App myapp in ./myapp from myrecipe
Unhandled exception: Capture group 1 was not matched (IndexError)
Failed to raise an exception: END_OF_STACK
[0x55e36525c326] *CallStack::print_backtrace:Int32 +118
[0x55e365224666] __crystal_raise +86
[0x55e365224dce] ???
[0x55e3652afa09] *Crystal::System::File::open<String, String, File::Permissions>:Int32 +201
[0x55e3652a492d] *File::new<String, String, File::Permissions, Nil, Nil>:File +45
[0x55e365253abd] *CallStack::read_dwarf_sections:(Array(Tuple(UInt64, UInt64, String)) | Nil) +109
[0x55e3652537fd] *CallStack::decode_line_number<UInt64>:Tuple(String | Nil, Int32, Int32) +45
[0x55e365252fa8] *CallStack#decode_backtrace:Array(String) +296
[0x55e365252e62] *CallStack#printable_backtrace:Array(String) +50
[0x55e3652a441d] *Exception+ +77
[0x55e3652a3868] *Exception+ +120
[0x55e36542ced8] *AtExitHandlers::run<Int32>:Int32 +488
[...]

PR #1033, opened 2019-01-10
Recipes: Fixes regex check of filename for liquid templates

sumwatt:
### Description of the Change

Issue #905 Removes the regex 'or' in favor of two distinct patterns using the (||) operator.

### Alternate Designs

Alternatively, we could just split the filename, prune the extension off and join it again rather than using $1

### Benefits

Forces $1 to be set

### Possible Drawbacks

Some minor overhead?

### DIFF
--- a/src/amber/cli/recipes/file_entries.cr
+++ b/src/amber/cli/recipes/file_entries.cr
@@ -44,15 +44,15 @@ module Amber::Recipes
           # process the filename with liquid
           template = Liquid::Template.parse filename
           filename = template.render @ctx.as(Liquid::Context)
-
-          if /^(.+)\.lqd$|^(.+)\.liquid$/ =~ filename
+	  
+          if /^(.+)\.lqd$/ =~ filename || /^(.+)\.liquid$/ =~ filename
             # process the file with liquid
             pack_liquid files, absolute_path, $1
           else
             # pack the file without processing
             pack_blob files, absolute_path, filename
           end
-        end
+	end
       end
     end
 
--- a/src/amber/cli/recipes/file_entries.cr
+++ b/src/amber/cli/recipes/file_entries.cr
@@ -44,7 +44,7 @@ module Amber::Recipes
           # process the filename with liquid
           template = Liquid::Template.parse filename
           filename = template.render @ctx.as(Liquid::Context)
-	  
+
           if /^(.+)\.lqd$/ =~ filename || /^(.+)\.liquid$/ =~ filename
             # process the file with liquid
             pack_liquid files, absolute_path, $1
@@ -52,7 +52,7 @@ module Amber::Recipes
             # pack the file without processing
             pack_blob files, absolute_path, filename
           end
-	end
+        end
       end
     end
 


damianham:
@drujensen we have already moved to shard based recipes.  I haven't had any free time lately to contribute more to the recipes and the recipe feature.  We need more recipes and also adding their info to the online docs or database driven recipe info that recipe maintainers can update themselves.

sumwatt:
> Can you add a spec for this regex?

I'll take a look and see if I can get something written up.

> Also, are you using recipes? We are working on moving recipes to shards (if possible, still questionable) but we would like to know if this feature is being used and if a breaking change would have an impact.

I was trying to find something to keep me occupied last night. I haven't used recipes so I'm not directly affected. I did see some of the other discussions going on around custom generators - what would be the difference between using creating a recipe and writing a custom generator?

sumwatt:
Need to reject this. It doesn't fix the issue. I thought I had tested it with both template extensions but it still fails on .liquid files. I'll keep poking at it.

damianham:
> @damianham We are close. We have github repos, but not shards. I'm refactoring the generators to see if this is possible. I'm trying to avoid making too many changes that would invalidate the existing recipes.

@drujensen the two existing recipes are in shard format.  One way or another you have to bootstrap the recipe to use as the app template.

> The goal would be that `-r` would add a shard too `shard.yml` and build a local version `/bin/amber` giving us the ability to add new generators or replacing existing ones.

Which shard.yml would you add that to ?  The binary I use is /usr/local/bin/amber and it is built from the amber git repo.  Yes we can add a shard to an application shard.yml and build a new amber binary but this is after the app has been created.  So there is a chicken and egg situation.  I think it is a great idea to be able to add new generators and replace existing ones just by installing a shard but I think it has to modify the global installation.

One option might be to change the global amber binary to a shell script that runs the app local amber binary in ./bin except for the new command which runs a binary specific to creating a new app e.g. /usr/local/bin/create_amber_app.  Using a shell script might make the whole installation and bootstrap process easier for users.  For example rvm and rbenv make it super simple to install ruby versions and gemsets, using a terminal command that a new user can copy and past from the rvm/rbenv website into their terminal.  It doesn't even need root privileges as the whole ecosystem is installed into the user's home directory.

Or -r adds a recipe shard to the global amber and rebuilds it.  However I think deferring CLI tasks to external programs might be a good option.  This would avoid rebuilding the global binary and it might allow shards to add new commands and command line args to the amber binary.

src/amber/cli/recipes/file_entries.cr
```diff
@@ -45,7 +45,7 @@ module Amber::Recipes
           template = Liquid::Template.parse filename
           filename = template.render @ctx.as(Liquid::Context)
 
-          if /^(.+)\.lqd$|^(.+)\.liquid$/ =~ filename
+          if /^(.+)\.lqd$/ =~ filename || /^(.+)\.liquid$/ =~ filename
```
eliasjpr: This change is fine, but let's try to make the regex constants for readability and a least one spec for the change.

drujensen:
@sumwatt will you be able to add a spec for this?

PR #1037, opened 2019-01-14
Treat mailer as first class citizen in new app

### LINKED ISSUE #1031: Mailer is not included by default, must be specifically required
To get the mailers working I had to add the following require statements: 

```ruby
require "quartz_mailer"
require "src/mailers/**"
```

I've submitted a PR https://github.com/amberframework/docs/pull/127 to update the documentation accordingly however I do not know if this instead should have been required by default.

westonganger:
Currently the mailer initializers and files are only added if you use the generator which is super confusing for people who do it by hand.

The mailer should be treated as a first class citizen. Lets make the mailer opt-out instead of opt-in. If an application does not need the mailer they can comment out the initializer or delete it.

Fixes #1031 

I will create a PR to update the docs accordingly afterwards.


westonganger:
Updated. Specs are now passing.

Issue #1042, opened 2019-01-15
Make it easier to use Params without the built-in validation API


westonganger:
Sometimes the validation API is nice but for many params I have no need for it and skip this step completely. Since I was not using this feature I was super confused why when debugging `params.to_h` was always empty no matter what the query string was. Turns out I must use `params.raw_params.to_h`. None of these details were mentioned in the docs so I had to search the source code.

PR #1043, opened 2019-01-15
Add method `Amber::Validators::Params#to_unsafe_h`

westonganger:
Solves #1042 

PR for documentation to be added after.

### DIFF
--- a/spec/amber/validations/params_spec.cr
+++ b/spec/amber/validations/params_spec.cr
@@ -345,5 +345,16 @@ module Amber::Validators
         validator.validate!.should eq result
       end
     end
+
+    describe "#to_unsafe_h" do
+      it "returns request raw_params as a hash" do
+        http_params = params_builder("first_name=elias&last_name=perez")
+        validator = Validators::Params.new(http_params)
+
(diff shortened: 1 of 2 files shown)


drujensen:
needs at least one spec.

westonganger:
Spec added

drujensen:
I'm hesitant to add this since it is dangerous to allow unexpected params to be processed. I agree with the name of the method as `to_unsafe_h` but worried this may be abused and expose a way to bypass the safety net.

I am interested what @eliasjpr as well.

eliasjpr:
Im fine with this. I would rather have people to call `raw_params.to_h` themselves.

anamba:
Guess I'd better rename my own #to_unsafe_h to avoid clobbering this new one.

Perhaps I'll re-implement as a boolean flag to #to_h that would allow it to return the values of all declared params, even those that have not passed validation. I would use it to update the model object in memory, to populate the form after an invalid submission (the user shouldn't have to re-enter everything from scratch).

Let me know if that sounds useful and I can open a new PR.

Issue #1041, opened 2019-01-15
Generated ErrorControlller does not inherit from ApplicationController but still uses the application layout


westonganger:
When generating the Error controller, by default uses the `application.slang` layout file. So suddenly the compiler started complaining about missing methods in my layout file that are defined in ApplicationController but not defined in ErrorController.

It turned out the errors popped up because the ErrorController does not inherit from ApplicationController. I am fine with errors happening after I change code myself but when a generator does it I am left questioning what the heck happened.

Here are two possible solutions
- Preferred (If Possible) - Somehow have the ErrorController inherit from ApplicationController by default
- Alternative - Make ErrorController have its own `error.slang` layout file by default. This way if the user decides to change the layout to the application layout they will have more of an understanding why the errors started occurring.

elorest:
Option 1 probably makes the most sense since it's nice if error pages still look the same and allow you to navigate out without having to go back.

elorest:
@westonganger Are the missing methods in your layout created by you or generated by default? Last time I checked application layout doesn't depend on any methods in application controller by default. If so then this is probably something that should be fixed by end users on a case by case basis. If it happens by default then it should be fixed in amber.

westonganger:
@elorest the missing methods were added by me but again this happens after a generator has been used making it super confusing.

robacarp:
I think it makes sense for ErrorController to inherit from ApplicationController

eliasjpr:
I think we can create a module for these error helper methods. WDYT?

westonganger:
Yes that sounds perfect

```ruby
class ErrorController < ApplicationController
  include Amber::Controller::Helpers::ErrorControllerHelpers

end

module Amber::Controller::Helpers::ErrorControllerHelpers
  LAYOUT = "application.slang"

  def initialize(@context : HTTP::Server::Context, @ex : Exception)
    super(@context)
    @context.response.content_type = content_type
  end

  def forbidden
    render("forbidden.slang")
  end

  def not_found
    render("not_found.slang")
  end

  def internal_server_error
    render("internal_server_error.slang")
  end

  def bad_request
    render("bad_request.slang")
  end

  private def content_type
    if context.request.headers["Accept"]?
      request.headers["Accept"].split(",").first
    else
      "text/plain"
    end
  end

  private def response_format
    case content_type
    when "application/json"
      {"error": @ex.message}.to_json
    when "text/html"
      html_response
    else
      @ex.message
    end
  end

  private def html_response
[...]

PR #1052, opened 2019-01-30
Make Generated ErrorController inherit from ApplicationController

westonganger:
Solves #1041


westonganger:
Fixed. Specs passing, just CI shards install failure on one.

Issue #1053, opened 2019-01-30
Allow respond_with methods to accept a block


westonganger:
Currently the `respond_with` format methods only allow a String or Proc. The Proc syntax is awkward and unexpected. It is also currently undocumented.

```ruby
respond_with do
  html -> {
    flash[:notice] = "Sucesss"
    render("my_action.slang")
  }
end
```

It would be alot nicer if we could do simply pass a Block like the following:

```ruby
respond_with do
  html do
    flash[:notice] = "Sucesss"
    render("my_action.slang")
  end
end
```

I tried creating a patch:

```
module Amber::Controller::Helpers::Responders
  def html(&block)
    @available_responses[TYPE[:html]] = block.call; self
  end
end
```

However I kept running into the following error:

```
undefined local variable or method 'html'

      html do
      ^~~~
```

elorest:
I'll look into this.

elorest:
I've submitted a PR to address this.

robacarp:
@elorest @westonganger fixed?

Sija:
@robacarp yep, fixed by #1054.

PR #1054, opened 2019-01-31
added blocks to respond with

elorest:
### Description of the Change

Adds blocks to `respond_with` as requested in https://github.com/amberframework/amber/issues/1053.

### Benefits

User code is easier to read.

### Possible Drawbacks

A bit more code.


elorest:
Tests **ALL** pass locally. Not sure why're they failing on travis. Especially since the failing ones don't seem to have anything to do with my changes.

src/amber/controller/helpers/responders.cr
```diff
@@ -24,32 +24,51 @@ module Amber::Controller::Helpers
       end
 
       def js(js : String | ProcType)
```
Sija: I'd suggest refactoring this with macros:

```cr
      {% for type in %w(html xml js json text) %}
        def {{type.id}}(value : String | ProcType)
          @available_responses[TYPE[:{{type.id}}]] = value
          self
        end

        def {{type.id}}(&block : -> _)
          {{type.id}}(block)
        end
      {% end %}

      def json(value : Hash(Symbol | String, String))
        json(value.to_json)
      end

      def json(**args : Object)
        json(args.to_h)
      end
```

elorest: Less readable but definitely a good idea. I'll do that. 

While your here any idea why the specs won't pass in Travis. They work perfectly on all my local and work computers.

Sija: I'm seeing `Error: formatting './src/amber/environment/settings.cr' produced changes`, could it be this?

Sija: After this refactor `add_response` method becomes obsolete.

elorest: Thank you. Interestingly Travis passed at the time. I guess format works differently in 0.27.1

I was still running format and specs with 0.27.0

elorest:
Unfortunately travis still doesn't pass. Still is just fine locally. :(

Sija:
@elorest Now for a change _Crecto_ specs are failing.

elorest:
@sija yeah I noticed that. Can't make them fail in osx or arch though locally. I'm using 0.27.1 now so it should be the same.

Sija:
@elorest That's odd... Do you run specs with Crecto test suite (same command as in Travis)?

elorest:
@Sija Yep. `crystal spec ./spec/build_spec_crecto.cr`

src/amber/controller/helpers/responders.cr
elorest: I've tried getting it to work with 0.27.2 now as well with no luck. Works locally but not on travis.

elorest:
I've tried getting it to work with 0.27.2 now as well with no luck. Works locally but not on travis.

Sija:
@elorest Does Crecto itself works with Crystal 0.27.1+?

PR #1057, opened 2019-01-31
Remove --deps flag from CLI header text

kevinelliott:
This option does not appear to work so it should not be shown to users in this text.

```
$ amber new myapp -d pg -t slang -m granite --deps
Parsing Error: The --deps option is unknown.
```

### Description of the Change

Removes flag from command line in text.

### Benefits

New users will be less confused.

### Possible Drawbacks

Perhaps this flag should work and needs to be fixed?

### DIFF
--- a/src/amber/cli/commands.cr
+++ b/src/amber/cli/commands.cr
@@ -31,7 +31,7 @@ module Amber::CLI
         defaults values shown above in this help message.
 
         Usage:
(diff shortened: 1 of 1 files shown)


elorest:
It seems to have been set to default recently in @drujensen's changes. We should discuss the future of that flag here before removing it. Thanks for bringing that up.

PR #1060, opened 2019-02-08
Align session nav links on the right of the nav bar

damianham:
* add session.[lang].ecr layout partials

* render the session partial in the application layout

* rename the grantite/crecto auth nav templates so they add to the session partial

### Description of the Change

When creating a new amber app with version 11.3 and running these commands
```
amber g auth User;
amber g scaffold Category title:string
amber g scaffold Product title:string description:text category:reference user:reference
amber g scaffold Comment body:text product:reference user:reference
```
The nav bar had the session items links in between the Home link and the Categories link rather 
than aligned on the right.  This change adds a second nav bar unordered list that is aligned on the right and renders a session partial into the unordered list.  The Auth generator nav link template partials for Granite and Crecto were renamed so that they add links to the new session partial rather than the main nav links partial.

### Alternate Designs
None considered

### Benefits
Session links will now be aligned on the right of the nav bar.

### Possible Drawbacks

An unordered list of nav links for the session links has been added to the nav bar to align on the right.  The session links partial will be empty until the developer runs the auth generator with;
```
amber g auth User
```
So an empty unordered list will be rendered in the nav bar.


src/amber/cli/templates/app/src/views/layouts/application.ecr.ecr
```diff
@@ -19,6 +19,9 @@
         <ul class="navbar-nav mr-auto">
           <%="<"%>%= render(partial: "layouts/_nav.ecr") %>
         </ul>
+        <ul class="navbar-nav">
+          <%="<"%>%= render(partial: "layouts/_session.ecr") %>
+        </ul>
```
drujensen: can we inject this in the auth generator instead?  I don't want to polluted the initial `app` with stuff from `auth`.

PR #1066, opened 2019-02-15
allow saving a blank value to an Optional param (to clear out a previously saved value)

anamba:
### Description of the Change

This is a small, but important fix for the optional param PR I submitted a couple months back.

I'm not sure how I didn't notice it all this time, but currently, you can't clear out the value of an optional param by submitting a blank value (e.g. empty string). You can only replace the current value with another non-blank string.

This PR fixes that, so that you can set the value of an optional param to an empty or blank string if you want to.

### Possible Drawbacks

I didn't make any acoommodation for people who might be depending on the existing behavior.

### DIFF
--- a/spec/amber/validations/optional_rule_spec.cr
+++ b/spec/amber/validations/optional_rule_spec.cr
@@ -11,6 +11,7 @@ module Amber::Validators
         end
 
         rule.apply(params).should be_true
+        rule.value.should be_nil
       end
 
       it "applies rule when field is present" do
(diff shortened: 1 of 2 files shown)


spec/amber/validations/optional_rule_spec.cr
```diff
@@ -21,6 +22,20 @@ module Amber::Validators
...
+      it "allows an empty value" do
+        params = params_builder("field=")
+        error_message = "You must provide this field"
+        rule = OptionalRule.new("field", error_message) do
+          true
+        end
+
+        rule.apply(params).should be_true
+        rule.value.should_not be_nil
+        rule.value.should eq ""
```
robacarp: `should_not be_nil` is redundant with `should eq ""` both here and above.

anamba: Agreed, I think I just put that in there to contrast with the original example, but that's not a great reason.

Issue #1064, opened 2019-02-13
Better 500 error logs?


alex-min:
Hi,

I've noticed the 500 errors aren't that descriptive in some cases, if the pg connection fails for example, you just get something like:

```
Error: 500 (ERROR)
```

This happens because the error class itself is what happens, there's no message in this case.
If we add more information in ```pipes/error.cr```, the error becomes more obvious:

```
Error: 500 | (ERROR) DB::ConnectionRefused  lib/pg/src/pg/connection.cr:16:9 in 'initialize'
lib/pg/src/pg/connection.cr:7:5 in 'new'
lib/pg/src/pg/driver.cr:3:5 in 'build_connection'
lib/db/src/db/database.cr:51:9 in '->'
lib/db/src/db/pool.cr:255:3 in 'build_resource'
lib/db/src/db/pool.cr:17:34 in 'initialize'
lib/db/src/db/pool.cr:15:5 in 'new:initial_pool_size:max_pool_size:max_idle_pool_size:checkout_timeout:retry_attempts:retry_delay'
lib/db/src/db/database.cr:50:7 in 'initialize'
lib/db/src/db/database.cr:43:5 in 'new'
[... more traceback stuff]
```

I've changed ```pipes/error.cr``` to add this:

```crystal
      rescue ex : Exception
        context.response.status_code = 500
        error = Amber::Controller::Error.new(context, ex)
        context.response.print(error.internal_server_error)
        if ex.backtrace && ex.class.name
          Amber.logger.error "#{ex.class.name} #{ex.message} #{ex.backtrace.join('\n')}", "Error: 500", :red
        else
          Amber.logger.error ex.message, "Error: 500", :red
        end
      end
```

However I'm not a Crystal developer, so that probably looks ugly to you guys. Is that a sensible approach? Does that make sense? If it does not, could you help me to correct the syntax so I can make a PR?

Sorry for bothering, I'm pretty new to Crystal & Amber.

alex-min:
closed by #1067

PR #1067, opened 2019-02-16
Adding more details in the 500 errors, issue #1064

alex-min:
### Description of the Change

I'm adding a little bit more information to the logs, in some cases the errors are not very descriptive.
 if the pg connection fails for example, you just get something like:

```
Error: 500 (ERROR)
```

This happens because the error class itself is what happens, there's no message in this case.

If we add more information in `pipes/error.cr`, the error becomes more obvious:

```
Error: 500 | (ERROR) DB::ConnectionRefused  lib/pg/src/pg/connection.cr:16:9 in 'initialize'
lib/pg/src/pg/connection.cr:7:5 in 'new'
lib/pg/src/pg/driver.cr:3:5 in 'build_connection'
lib/db/src/db/database.cr:51:9 in '->'
lib/db/src/db/pool.cr:255:3 in 'build_resource'
lib/db/src/db/pool.cr:17:34 in 'initialize'
lib/db/src/db/pool.cr:15:5 in 'new:initial_pool_size:max_pool_size:max_idle_pool_size:checkout_timeout:retry_attempts:retry_delay'
lib/db/src/db/database.cr:50:7 in 'initialize'
lib/db/src/db/database.cr:43:5 in 'new'
[... more traceback stuff]
```

### Alternate Designs

I'm a beginner developer in Crystal / Amber so suggestions would be appreciated if that change looks ugly to you guys.

### Benefits

We can now clearly see the Errors in the logs.

### Possible Drawbacks

I can't think any drawback.

### DIFF
--- a/src/amber/pipes/error.cr
+++ b/src/amber/pipes/error.cr
@@ -21,7 +21,11 @@ module Amber
         context.response.status_code = 500
         error = Amber::Controller::Error.new(context, ex)
         context.response.print(error.internal_server_error)
-        Amber.logger.error ex.message, "Error: 500", :red
+        if ex.backtrace && ex.class.name
+          Amber.logger.error "#{ex.class.name} #{ex.message} #{ex.backtrace.join('\n')}", "Error: 500", :red
+        else
+          Amber.logger.error ex.message, "Error: 500", :red
+        end
       end
     end
   end


eliasjpr:
Love it!

robacarp:
thanks @alex-min 👍

PR #1068, opened 2019-02-17
set AMBER_ENV to 'test' when testing

damianham:
During the process of creating a recipe I use a script to automate the process of creating a new app with the new recipe and to run the scaffold generator to create application artifacts.  The script populates the database with some test records for each model with 'amber db seed'.  Upon starting the application with 'amber watch' the database was empty.  This was due to the crystal spec command clearing out each database table because the AMBER_ENV was 'development' rather than 'test'.  This change ensures that by default the AMBER_ENV environment variable is 'test' when running the specs.

### Description of the Change

* prepend AMBER_ENV=test to the amber watch spec run_commands

### Alternate Designs
none

### Benefits
You don't wipe out your development database

### Possible Drawbacks

test environment is hard wired in amber.yml to 'test'

### DIFF
--- a/src/amber/cli/templates/app/.amber.yml.ecr
+++ b/src/amber/cli/templates/app/.amber.yml.ecr
@@ -25,7 +25,7 @@ watch:
     #  - ./src/some_irrelevant_file.cr
   spec:
     run_commands:
-      - crystal spec
+      - AMBER_ENV=test crystal spec
     include:
       - ./spec/**/*.cr
   npm:


anamba:
Ideally this shouldn't be necessary, you want to set AMBER_ENV in the spec_helper (otherwise you have to remember to set AMBER_ENV=test every time you run `crystal spec`.

But it looks like it is in there by default:
https://github.com/amberframework/amber/blob/master/src/amber/cli/templates/app/spec/spec_helper.cr.ecr

Wonder what went wrong then?

anamba:
After some discussion on gitter with OP, his guess was correct, Amber actually sets the AMBER_ENV env var itself on startup (not sure this is a good idea, but it might break things if we change it now):
https://github.com/amberframework/amber/blob/149f103ce264a21744c150e20dc677f3d850a33b/src/amber/environment/env.cr#L6

So that gets set by `amber watch`, then by the time it runs the spec_helper, ENV["AMBER_ENV"]  == development.

So we should probably merge this AND maybe also change the ||= to = in spec_helper.cr.ecr:
```
ENV["AMBER_ENV"] = "test"
```

That should be nice and safe. I didn't notice this bug because I have a failsafe in my own spec_helper that sets the database url to point to a test db. (Maybe I ran into this bug myself a long time ago, but didn't have the time to dig into it then...)

Issue #1070, opened 2019-02-23
i18n files do not reload on changes


alex-min:
Hi everyone,

I've noticed the i18n files do not reload when you save them, you have to run "amber watch" again to make the changes available.

Would it be possible to recompile them if they changed? I need to check the current code on how it's done.

robacarp:
@alex-min it is definitely possible. I think you can update your .amber.yml config to instruct the watch command to pay attention to the i18n files.

kvirani:
This is fixed in #1071 so can we close this issue?

alex-min:
yep sorry, forgot to close it

PR #1071, opened 2019-02-23
reload the i18n locale files automatically

alex-min:
fixing #1070

Currently, the 18n locale files do not reload automatically unless you specify them in the list of things to watch.
With this patch they are included by default in the template so that when you create a new project, you can edit them right away and see the i18n changes.

### Possible Drawbacks

I might have to submit the same patch for the recipes


robacarp:
thanks @alex-min

PR #1072, opened 2019-02-23
adding an option to print the routes as json

alex-min:
### Description of the Change

For some internal stuff I'm building, I would need to print the routes as json, this makes it easier to parse from external tools. I thought this change might be useful for other amber users, if the maintainers don't agree, you can just remove this PR otherwise :)

### Alternate Designs

If we had multiple formats, maybe we could do a ```--format=json``` ? since I've only added this one I thought just adding a ```--json``` would be enough, please let me know what you think of that :)

### Benefits

Being able to reuse the routes in a easier way from external tools. Just a JSON.parse from any language does the trick instead of trying to get the routes manually with some regex.

### Possible Drawbacks

I can't think of any.

### DIFF
--- a/spec/amber/cli/commands/routes/routes_spec.cr
+++ b/spec/amber/cli/commands/routes/routes_spec.cr
@@ -48,6 +48,34 @@ module Amber::CLI
           end
         end
 
+        describe "with the default routes as json" do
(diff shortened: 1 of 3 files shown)


alex-min:
@robacarp I've added some tests, let me know what you think

PR #1073, opened 2019-02-23
adding error validation to the authentication

alex-min:
### Description of the Change

This a followup of #1069, the changes done to the validator allowed me to improve the authentication template.

Currently, when you use the auth template and click on "Sign Up" without filling anything, you just have a 500 error because these validation errors were not caught.

This patch is rescuing the validation errors and putting them in the template files so it results in something like this instead of a 500 error:

screenshot_2019-02-23 abc using amber

### Alternate Designs

Maybe we could have removed the validation completely & let granite or crecto do it?

### Benefits

By default the sign up works without throwing a 500 error.

### Possible Drawbacks

I have to specify a not_nil! in the template, which is very strange since I do a if check just before, maybe someone more experimented in Crystal could help me to understand why.

### DIFF
--- a/src/amber/cli/templates/auth/crecto/src/controllers/registration_controller.cr.ecr
+++ b/src/amber/cli/templates/auth/crecto/src/controllers/registration_controller.cr.ecr
@@ -5,6 +5,8 @@ class RegistrationController < ApplicationController
   end
 
   def create
+    @validation_errors = [] of Amber::Validators::Error
+    registration_params.validate!
     <%= @name %> = <%= class_name %>.new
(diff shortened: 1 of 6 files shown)


alex-min:
Thanks, yeah it's not that great visually since it's a template in a template.

On the next PRs I would like to improve more stuff on the auth system, I would probably add:

-  specs, currently it just creates empty specs but we could add specs in the templates to verify that the authentication & profile pages are working
- reset password feature
- auth i18n support, maybe creating something like auth-i18n?

eliasjpr:
I thought we were going to start using slang as the default template engine for our generators.

PR #1077, opened 2019-03-01
spec improvements & spec formatting

alex-min:
I've noticed the project specs were not running due to this line in the build_spec_helper

```crystal
MainCommand.run ["generate", "error"]
```

It was missing the last parameter making the test suite not execute. I've fixed that and also added more generators to test.

This is now testing the whole project for crystal formatting so that the specs which are produced with the templates are valid.

I've also changed some of the templates so that some of the generated specs without parameters are compiling.

This breaks the crecto tests since crecto is not working on crystal 0.27.2, @elorest disabled the crecto tests on travis, the crecto tests could be enabled again when https://github.com/Crecto/crecto/issues/232 will be fixed.


spec/build_spec_helper.cr
```diff
@@ -15,15 +15,20 @@ module Amber::CLI
     options = ["user:reference", "name:string", "body:text", "age:integer", "published:bool"]
     temp_options = options - ["user:reference", "age:integer"]
     MainCommand.run ["generate", "auth", "User"] | (options - ["user:reference"])
-    MainCommand.run ["generate", "error"]
+    MainCommand.run ["generate", "error", "ErrorPage"]
```
alex-min: this was the line making the test suite not execute

elorest: Thanks. Possibly it should just be commented out with a TODO?

alex-min: I thought it was worth to test if the error pages are working (maybe the original intent here) that's why I added the name "ErrorPage"

alex-min: it was not executing because ```amber generate error``` just shows the help message since it's missing the error page name

alex-min:
Thanks @elorest, I did not see things changed since I opened the PR.
Does anyone need anything from my end in order to merge this PR? Anything missing?

elorest:
@alex-min I think it's good. There were a few changes we had to make to specs on the 0.28.0 release. Some of them related to the release, but many were bugs that didn't show up until we updated the pg, mysql shards. Lots of CLI/Deps related things. I'm just trying to verify that this doesn't change anything there. 

At the very least the specs pass so that's a good sign. If you wouldn't mind doing a quick scan of the current changes since 0.27.2 in relation to this I'd appreciate it. If everything looks good I'll merge tonight.

PR #1084, opened 2019-03-30
Fix settings and dep injection

elorest:
### Description of the Change

Because of the previous load order ENV Var injection into application.cr was being ignored because database, models, controllers, initializers, etc were being loaded before ENV vars were being added to settings. 

The fix was to move settings modification out into `config/settings.cr` and then load it at the correct location.

### Benefits

Projects work as expected.

### DIFF
--- a/src/amber/cli/helpers/helpers.cr
+++ b/src/amber/cli/helpers/helpers.cr
@@ -29,7 +29,7 @@ module Amber::CLI::Helpers
     deps = dependencies.split("\n").reject { |d| application.includes?(d) }
 
     replacement = <<-REQUIRES
-    #{injection_marker} 
+    #{injection_marker}
     #{deps.join("\n")}
     REQUIRES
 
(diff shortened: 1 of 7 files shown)


kvirani:
Nice! Will this also fix the issue I'm seeing with my seeds command talking about `Watcher` in its output?

```terminal
> amber db seed    
10:43:56 Watcher    | (INFO) Watching 11 files (browser reload)...
hello from seeds
10:43:56 Database   | (INFO) Seeded database
```

db/seeds.cr:
```crystal
require "../config/application.cr"
puts "Hello from seeds"
```

Or am I requiring the wrong file in order to require my models, etc in the seeds file?

elorest:
Probably related but different. Could you give me more information on that.

kvirani:
> Probably related but different. Could you give me more information on that.

@elorest see newly logged issue https://github.com/amberframework/amber/issues/1086

elorest:
Thanks. I think I have a solution for that.

PR #1090, opened 2019-04-07
Add constraints support for routes

### LINKED ISSUE #1089: Route Constraints
Since https://github.com/amberframework/amber-router/pull/12 was merged Amber supports adding constraints to its routes, like https://guides.rubyonrails.org/routing.html#segment-constraints.  

However there is not currently a method of defining these on a given route/resource.

Should be pretty simple like
```crystal
get "/user/:id", UserController, :show, constraints: {id: /\d+/}
```

Then just pass the constraints to the router when adding the route.

kvirani:
### Description of the Change

Constraints can now be passed into route definitions.

They can be passed in through the `get`, `post`, etc macros, and also the `resources` macros in the Amber router. 

This is done by passing constraints through to amber-router's RouteSet: https://github.com/amberframework/amber-router/blob/master/spec/amber_router/route_set/matching/routes_with_variables_spec.cr#L41

**Rails (_example from the RoR guides_):**
```ruby
get '/:id', to: 'articles#show', constraints: { id: /\d.+/ }
get '/:username', to: 'users#show'
```

**Amber**
```crystal
get "/:id", ArticlesController, :show, constraints: {"id" => /\d.+/}
get "/:username", UsersController, :show
```

Resolves #1089

### Alternate Designs

Only Hashes are supported at the moment. Not sure if we should also be supporting NamedTuple here? 

### Out of Scope

- Unlike with Rails, we have to explicitly pass in the constraints key/argument (`constraints: {...}`). In Rails there is shorthand for this but I'm not a fan of having too much shorthand support to begin with. I prefer a healthy balance of explicit vs implicit, and Rails sometimes goes too far to the right, imho. 
- Specific constraint keys like `subdomain` are not supported in this PR

### Benefits

Much like with [constraints in Rails](https://guides.rubyonrails.org/routing.html#segment-constraints), Amber devs can now pass request param value constraints using the `constraints` optional argument in the `routes.cr` file. 

### Possible Drawbacks

1. NamedTuple for constraints are not supported, but they should be, I think? Looking for actionable advice there.
2. I _think_ that the regex here is only looking for the pattern to be present in the request segment, but not fully matching the segment. This would result in different behavior than what is expected from constraints in the Rails dispatcher.

### DIFF
--- a/spec/amber/router/route_spec.cr
+++ b/spec/amber/router/route_spec.cr
@@ -22,6 +22,15 @@ module Amber
       end
     end
 
+    describe "#to_json" do
+      it "includes constraints key" do
+        route = Route.new("GET", "/", handler, constraints: { "id" => /\d/, "slug" => /\w+/ })
+        obj = JSON.parse(route.to_json)
+        obj["constraints"]["id"].should be_truthy
+        obj["constraints"]["slug"].should be_truthy
+      end
+    end
+
     describe "#call" do
       context "before action" do
         it "does not execute action" do
--- a/spec/amber/router/route_spec.cr
+++ b/spec/amber/router/route_spec.cr
@@ -24,7 +24,7 @@ module Amber
 
     describe "#to_json" do
       it "includes constraints key" do
-        route = Route.new("GET", "/", handler, constraints: { "id" => /\d/, "slug" => /\w+/ })
+        route = Route.new("GET", "/", handler, constraints: {"id" => /\d/, "slug" => /\w+/})
         obj = JSON.parse(route.to_json)
         obj["constraints"]["id"].should be_truthy
         obj["constraints"]["slug"].should be_truthy
--- a/spec/amber/router/router_spec.cr
+++ b/spec/amber/router/router_spec.cr
@@ -34,7 +34,7 @@ module Amber
           it "uses constraints for defining crud routes" do
             router = Router.new
             router.draw :web do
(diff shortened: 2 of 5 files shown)


elorest:
It would be nice to support named tuples. Even if it only supported them. All this gets rendered precompile anyway.

kvirani:
My understanding of NamedTuples is that they are better when devs (not just the compiler) knows what keys the Tuple is to expect. Are we sure that we _should_ support NamedTuple here? Yes it would be nice to support different "hash-like" syntax and therefore let it be more ruby-esque, but perhaps the more specific syntax isn't a bad thing, here?

wontruefree:
@elorest there is a PR for that https://github.com/amberframework/amber-router/pull/17

elorest:
That PR isn't really what I meant. I would expect for this to be converted before the macro's even add routes to amber-router.

`My understanding of NamedTuples is that they are better when devs (not just the compiler) knows what keys the Tuple is to expect` I would argue that since all routes are added at compile time that a named_tuple should be the default. Values never change on these during run time.

`{"id" => /\d.+/}` is a string key. `{id: /\d.+/}` is a symbol key. Neither are any easier or harder for the dev or system to understand. The named_tuple is faster for the system to use though and easier for the dev to type.

For now I'm just going to merge this.

kvirani:
@elorest thanks for merging, and I do see your latest point about the NamedTuple approach making sense since these don't mutate during runtime. 

Will look to add that support in.

PR #1091, opened 2019-04-16
Upgrade to Crystal 0.28.0

bcardiff:
I found the following changes needed to upgrade to the upcoming Crystal 0.28.0.

If some macro `compare_versions` are added it should be possible to support current versions as well.

Feel free to extract these changes and close the PR if you prefer to handle the migration in a specific way.

### DIFF
--- a/src/amber/support/message_encryptor.cr
+++ b/src/amber/support/message_encryptor.cr
@@ -63,7 +63,7 @@ module Amber::Support
     end
 
     private def sign_bytes(data : Bytes)
-      OpenSSL::HMAC.digest(@digest, @secret, data)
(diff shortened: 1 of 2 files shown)


bcardiff:
Note that db related shards are also in the process of been updated for Crystal 0.28.0.

elorest:
Thank you.

bcardiff:
FYI the crystal-pg/mysql/sqlite3 shards are already released with 0.28.0 support

elorest:
Thanks a lot for your help on this. Upgrading proved more difficult than most but it's been accomplished now.

PR #1092, opened 2019-04-24
initial pass at adding clear to amber

wontruefree:
### Description of the Change

This adds the ORM [clear](https://github.com/anykeyh/clear) to the generators.  Clear is an ORM built specifically for PostgreSQL in Crystal.  It offers a lot of features and is actively maintained.

### Benefits

Adding a new framework that is postgres specific and has a lot of super user features.

### Possible Drawbacks

This adds more cases to the generators and possibly more maintenance.

### DIFF
--- a/src/amber/cli/commands.cr
+++ b/src/amber/cli/commands.cr
@@ -46,7 +46,7 @@ module Amber::CLI
       help desc: "describe available commands and usages"
       string ["-t", "--template"], desc: "preconfigure for selected template engine", any_of: %w(slang ecr), default: "slang"
       string ["-d", "--database"], desc: "preconfigure for selected database.", any_of: %w(pg mysql sqlite), default: "pg"
-      string ["-m", "--model"], desc: "preconfigure for selected model", any_of: %w(granite crecto), default: "granite"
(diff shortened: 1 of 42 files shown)


wontruefree:
This is still very much a work in progress.  I am stuck though and wanted to see if anyone could help me.  I am getting conflicts in `shards install`.  This seems to be a conflict in transitive dependencies.  I can fork clear and just delete the dependency but obviously, it won't work as intended.  The dependency is not needed when interfacing with amber so it might not be a bad solution.

robacarp:
@wontruefree clear is awesome! I think it'll be more maintainable to move this into an amber recipe instead of using the built in generators. Generally the effort with the built in generators is to simplify and reduce complexity, and to move as much as possible into a recipe.

wontruefree:
I don't mind making this a recipe.
Is there any interest in the amber community in adding more ORMs?

wontruefree:
@robacarp the initial problem with shards is now fixed.  But it seems like the amber team would rather this as a recipe right?

robacarp:
@wontruefree yes, that way the amber generators don't have N complexity for N app skeletons and database managers

PR #1093, opened 2019-04-26
Builds with crystal 0.28.0.

hugopl:


hugopl:
Weird... granite tests pass locally on my machine, even after a "git clean -dfx".

elorest:
@hugopl thanks for your work on this. I ended up getting multiple PR's on this at around the same time. CLI tests were hard because of the database changes. That's most likely why they're failing here.

Issue #995, opened 2018-12-16
Error on type model


nbouliol:
### Description

Having a model called `Type` is not well supported

### Steps to Reproduce

I did

`amber g scaffold Type label:string`
other scaffolding, then
`amber watch`

**Actual behavior:** https://pastebin.com/P9En4Qsn

### Versions

```
→ amber --version
Amber CLI (amberframework.org) - v0.11.1

→ crystal --version
Crystal 0.27.0 (2018-11-04)

LLVM: 6.0.1
Default target: x86_64-apple-macosx
```

drujensen:
Yeah, there are several reserved words in Crystal and Amber.  Not sure the best way to deal with this except maybe check to see if someone is trying to use one and prevent it.

eliasjpr:
I think the keyword `Type` might be reserved

elorest:
Not sure if this is an issue we should worry about. There are multiple key words in most other frameworks including Rails, Phoenix that you just need to know not to use.

drujensen:
@robacarp Is there a list of reserved words we should check for?

robacarp:
Well, I'm not sure what all would be in an exhaustive list but it could start with Type, Array, Int32, etc. Plan for eventual completeness?

nsuchy:
The list from that issue at least looks like a good start. Do we want to check for these words in https://github.com/amberframework/amber/blob/master/src/amber/cli/generators/scaffold.cr or at a higher level in the commands chain? Should we return an exception or a pretty console output? Working on a PR to close out this issue.

PR #1096, opened 2019-05-06
Warn when using keywords as a name provided to a generator

nsuchy:
### Description of the Change

Issue #995 describes a bug where the generator will fail when using a Crystal keyword as a model name.  This changes the scaffold generator and adds a string array of keywords and returns a descriptive error if you attempt to use a keyword as a model name.

### Alternate Designs

0. We could just let it fail and expect the user to know to not use a keyword.
1. We could investigate if Crystal has a built in method to check if a word is a keyword rather than using a hard coded string array.

### Benefits

1. The error message is more descriptive to the user of what the exact issue is.
2. Prevents a situation where part of the necessary files are added before the compiler freaks out at improper use of a keyword.
3. Closes an issue 😛 

### Possible Drawbacks

There might be edge cases where using a Crystal keyword as a model name is allowed or required for a project. I'm not aware of any but if someone would like to chip-in this would be a good forum to do so.

There are not any unit tests for allowed / disallowed model names to test the generator. I plan to add a few tests for this functionality.

Chatlog on gitter: https://gitter.im/amberframework/amber?at=5cd0a3d7375bac747068e843


src/amber/cli/generators/scaffold.cr
```diff
@@ -7,17 +7,84 @@ module Amber::CLI
     property view : Generator
 
     def initialize(name, fields)
-      super(name, fields)
-      if config.model == "crecto"
-        @migration = CrectoMigration.new(name, fields)
-        @model = CrectoModel.new(name, fields)
-        @controller = ScaffoldCrectoController.new(name, fields)
+      # Keywords from https://github.com/crystal-lang/crystal/wiki/Crystal-for-Rubyists#available-keywords as of May 6th, 2019
+      keywords = [
```
Blacksmoke16: this could prob be a constant `KEYWORDS = [...]`

nsuchy:
Not a unit test, but example of how my solution works:
```
Nathaniels-MacBook-Pro% ../amber g scaffold Type
05:21:27 Generate   | (INFO) Generating Amber::CLI::Scaffold
Name cannot be a keyword.
Nathaniels-MacBook-Pro% ../amber g scaffold Post
05:21:42 Generate   | (INFO) Generating Amber::CLI::Scaffold
05:21:42 Generate   | (INFO) new       db/migrations/20190506172142811_create_post.sql
05:21:42 Generate   | (INFO) new       spec/models/post_spec.cr
05:21:42 Generate   | (INFO) new       spec/models/spec_helper.cr
05:21:42 Generate   | (INFO) new       src/models/post.cr
05:21:42 Generate   | (INFO) new       spec/controllers/post_controller_spec.cr
05:21:42 Generate   | (INFO) new       spec/controllers/spec_helper.cr
05:21:42 Generate   | (INFO) new       src/controllers/post_controller.cr
05:21:42 Generate   | (INFO) rewritten src/views/layouts/_nav.slang
05:21:42 Generate   | (INFO) new       src/views/post/_form.slang
05:21:42 Generate   | (INFO) new       src/views/post/edit.slang
05:21:42 Generate   | (INFO) new       src/views/post/index.slang
05:21:42 Generate   | (INFO) new       src/views/post/new.slang
05:21:42 Generate   | (INFO) new       src/views/post/show.slang
Nathaniels-MacBook-Pro% 
```

src/amber/cli/generators/scaffold.cr
nsuchy: It can't.
```

Syntax error in src/amber/cli/generators/scaffold.cr:11: dynamic constant assignment. Constants can only be declared at the top level or inside other types.

      KEYWORDS = [
```

Blacksmoke16: Would have to move it outside of `initialize`.

nsuchy:
Looking at the issue from a larger perspective should we also perform the same checks on other generators. Could we move keywords up higher and check the same constant on multiple generators? For example the controller and model generators.

src/amber/cli/generators/scaffold.cr
```diff
@@ -7,17 +7,84 @@ module Amber::CLI
...
+        "abstract",
+        "do",
+        "if",
+        "nil?",
+        "select",
+        "union",
+        "alias",
```
elorest: This list seems to be of method names even though constants being defined. It would technically be sort of ok to have a model called "Unless" other than what it would do to possible associations.

nsuchy: In the chat we discussed (https://gitter.im/amberframework/amber?at=5cd0b8f5375bac7470697ff2) potentially moving this logic over into `src/amber/cli/generators.cr` since some validations for the name already occur there. Would you be opposed to moving this logic into that file and outputting a warning (e.g. "It looks like you may be using a keyword.") but still allowing it to complete?

elorest: No that sounds good as long as it doesn't increase code complexity substantially.

nsuchy: Okay I've made changes that should address your concerns. Let me know if there's any other changes you believe are necessary. With the only logical change being a puts statement to output a warning do we need to add additional specs for this change?

nsuchy:
Updated the pull request's name and the commit message as the purpose has changed a bit.

damianham:
You might also want to include standard library classes.  I had an issue after upgrading to crystal 0.28 with the new Path class conflicting with my own Path model so I had to refactor.

nsuchy:
Could you provide an list? I fear this goes back to what @elorest brought up that it’s impossible for us to make a list of every last macro and class name. I think limiting to language keywords would be best for now.

damianham:
I think warning about the API classes would be more useful than warning about 'if' or 'else' as model class names.  It is highly unlikely that someone would generate model or scaffold for an 'else' class but there are classes in the API that one could conceivably use as names for models for a project, e.g. Signal is something that I might use as I am sending signals from my amber app to connected clients over websockets.

nsuchy:
@damianham How would you recommend implementing this? Is there a way to do this other than a hard coded list?

@elorest Did the changes I made meet your expectations?

elorest:
@nsuchy Yes. Adding a few stdlib classes to the list would probably be a good idea as well.

| Is there a way to do this other than a hard coded list?

There are ways with macro's but would slow down execution of generators considerably. 

@damianham Even if we had a list of std classes it wouldn't have solved your issue with existing names. Personally this whole thing seems a little too training wheels for me. Warning over obviously bad names is probably a good idea but I don't think people should become too reliant on it.

damianham:
@nsuchy I think the easiest implementation would be a hard coded list.
@elorest true enough that a list of existing classes wouldn't have helped me because the conflicting class was introduced in a later version of crystal after my model had been created.  I would venture that once we have a list of class names that generate a conflict warning we might add some class names that are not yet in the std lib but which could possibly be part of the std lib in the future e.g. std lib classes from other languages.

Yeah it is a bit of training wheels but if you remember one of the best features of gcc was the excellent and descriptive error/warning messages.  It made a big difference over using a standard compiler with cryptic hard to decipher error/warning messages.  So we would be improving the Amber developer experience without any kind of performance degredation.  That can't be a bad thing.

nsuchy:
> @nsuchy I think the easiest implementation would be a hard coded list.
> @elorest true enough that a list of existing classes wouldn't have helped me because the conflicting class was introduced in a later version of crystal after my model had been created. I would venture that once we have a list of class names that generate a conflict warning we might add some class names that are not yet in the std lib but which could possibly be part of the std lib in the future e.g. std lib classes from other languages.
> 
> Yeah it is a bit of training wheels but if you remember one of the best features of gcc was the excellent and descriptive error/warning messages. It made a big difference over using a standard compiler with cryptic hard to decipher error/warning messages. So we would be improving the Amber developer experience without any kind of performance degredation. That can't be a bad thing.

Let's keep in mind that our main "competitor" is Ruby/Ruby on Rails which has just that. An appeal of Crystal is ruby-like syntax (in fact it's nearly identical) this means new Crystal programmers (like me) can focus on learning the technical differences between the two languages and the crystal standard library rather than learning yet another syntax. Ruby on Rails came out in 2005 and is undeniable more mature than Amber. If we want to draw in new programmers (there are way more ruby gems than crystal shards) we need to create a better offering. While adding in helpful warnings is tedious, it does help new programmers starting out with Crystal/Amber prevent making basic mistakes. Please don't let this become a copy of Elixir/Phoenix where you need to be an senior expert engineer to understand the error messages and warnings.



> @nsuchy Yes. Adding a few stdlib classes to the list would probably be a good idea as well.
> 
> | Is there a way to do this other than a hard coded list?
> 
> There are ways with macro's but would slow down execution of generators considerably.
> 
> @damianham Even if we had a list of std classes it wouldn't have solved your issue with existing names. Personally this whole thing seems a little too training wheels for me. Warning over obviously bad names is probably a good idea but I don't think people should become too reliant on it.

Do you want all of the stdlib class names? If not could you provide a list of the ones you'd like included? I wouldn't opt for a macro if it'll slow things down, I want to make things better not worse.

nsuchy:
@elorest I came up with the following list of classes from the Crystal Standard Library. Any objections?
```crystal
KEYWORDS = [
  "adler32",
  "argumenterror",
  "array",
  "atomic",
  "base64",
  "benchmark",
  "bigdecimal",
  "bigfloat",
  "bigint",
  "bigrational",
  "bitarray",
  "bool",
  "box",
  "bytes",
  "channel",
  "char",
  "class",
  "colorize",
  "comparable",
  "complex",
  "concurrent",
  "concurrentexecutionexception",
  "crc32",
  "crypto",
  "crystal",
  "csv",
  "debug",
  "deprecated",
  "deque",
  "digest",
  "dir",
  "divisionbyzeroerror",
  "dl",
  "ecr",
  "enum",
  "enumerable",
  "env",
  "errno",
  "exception",
  "fiber",
  "file",
  "fileutils",
  "flags",
  "flate",
  "float",
  "float32",
  "float64",
  "gc",
  "gzip",
  "hash",
  "html",
  "http",
  "indexable",
  "indexerror",
  "ini",
  "int",
  "int128",
  "int16",
  "int32",
  "int64",
  "int8",
  "invalidbigdecimalexception",
  "invalidbytesequenceerror",
  "io",
  "ipsocket",
  "iterable",
  "iterator",
  "json",
  "keyerror",
  "levenshtein",
  "link",
  "llvm",
  "logger",
  "markdown",
  "math",
  "mime",
  "mutex",
  "namedtuple",
  "nil",
  "nilassertionerror",
  "notimplementederror",
  "number",
  "oauth",
  "oauth2",
  "object",
  "openssl",
  "optionparser",
  "overflowerror",
  "partialcomparable",
  "path",
  "pointer",
  "prettyprint",
  "proc",
  "process",
  "random",
  "range",
  "readline",
  "reference",
  "reflect",
  "regex",
  "semanticversion",
  "set",
  "signal",
  "slice",
  "socket",
  "spec",
  "staticarray",
  "string",
  "stringpool",
  "stringscanner",
  "struct",
  "symbol",
  "system",
  "tcpserver",
  "tcpsocket",
  "termios",
  "time",
  "tuple",
  "typecasterror",
  "udpsocket",
  "uint128",
  "uint16",
  "uint32",
  "uint64",
  "uint8",
  "union",
  "unixserver",
  "unixsocket",
  "uri",
  "uuid",
  "valist",
  "value",
  "weakref",
  "xml",
  "yaml",
  "zip",
  "zlib",
]
```

Blacksmoke16:
@nsuchy https://play.crystal-lang.org/#/r/6vq8 This macro would handle built in crystal types.  IMO i dont think it would have a big impact on performance since it would just be creating an array at compile time, but could test it if you wanted.

nsuchy:
> @nsuchy https://play.crystal-lang.org/#/r/6vq8 This macro would handle built in crystal types. IMO i dont think it would have a big impact on performance since it would just be creating an array at compile time, but could test it if you wanted.
> 
> EDIT: Actually seems to be missing some stuff, give me a few.

Any way you could have it generate a tuple instead?

damianham:
@nsuchy I am not so sure you need every std lib class in the list.  There are many entries in that list that are obvious that any programmer would not use as a model name, e.g. 'array' and 'int32'.  So personally I would limit the list to class names that are not primitive types and also less common so the list would be very small although I appreciate in practice it may not make much difference.

nsuchy:
@damianham I don’t have technical background or knowledge to cherrypick which class names users already know about / don’t know about.

Performance wise I don’t believe a few extra class names will make a significant differences.

Will you please provide a list of the names you want added? Thanks :)

nsuchy:
@elorest My opinion here is that since we're only warning the user and not stopping them from using the names it's okay to be a bit more aggressive with how many warnings we create. There's almost zero performance impact and the warning can still be useful and have users think twice about using something for a model name. I've updated by commit to use the big list, also ran `crystal tool format` so that's taken care of as well.

nsuchy:
> I'm in agreement with @elorest that maintaining this list is going to be a hassle. Is there a way to auto-generate this list of keywords?

@Blacksmoke16 Was trying that with a macro but it only generated a list of class names currently in use, not all class names.

It'll be a hassle yes, but it's only a warning, not an error so it won't be a huge issue.

drujensen:
@elorest @damianham are you ok with this change?

damianham:
@drujensen yes it is OK by me

Issue #1081, opened 2019-03-19
Drop support for crecto


Meldanor:
### Description

Crecto, an ORM, has decided to no longer maintain the project: https://github.com/Crecto/crecto/commit/4bf83cbfc1fdf6cf4efef05ac0fe6c72b826d90f

Amber should drop the support for this ORM - as a start by removing the documentation https://docs.amberframework.org/amber/guides/models/crecto from the website.

### Steps to Reproduce

Go to the Crecto Repository: https://github.com/Crecto/crecto/

### Versions

Does not work with crystal 0.27.1 or higher

### Additional Information

None

elorest:
Thanks for this. Please add submit PR to remove from `amberframework/docs`. I'll start working on removing from project. @drujensen do you see any issues with this? @samholst I know you were working on fixing specs for this.

samholst:
@elorest Thank you for the update, I’ll try and keep an eye out to see what happens with this.

elorest:
@amberframework/core-team Lets get this removed before the next release.

nsuchy:
@elorest If someone hasn't started already, is it okay for me to make a pull request that removes Crecto from the docs?

PR #1100, opened 2019-05-13
Drop support for crecto (#1081).

hugopl:
This removes the -m option form CLI, since now Amber only supports granite.


hugopl:
Should I rebase this or you guys can resolve the conflicts in github UI?

robacarp:
@hugopl it looks like the diff is more complicated that github can handle

hugopl:
Rebase done.

Issue #1110, opened 2019-05-29
Add except method to filters.cr


AndyRosenberg:
https://github.com/amberframework/amber/blob/master/src/amber/controller/filters.cr

Can someone add an except method to these filters? Currently, there's just all and only, which leaves you writing blocks like:

```
only [:new, :create, :edit, :update, :destroy] do
  #insert code
end
```

PR #1111, opened 2019-05-30
Add except method to filters.cr

AndyRosenberg:
Related to this issue:
https://github.com/amberframework/amber/issues/1110

### DIFF
--- a/src/amber/controller/filters.cr
+++ b/src/amber/controller/filters.cr
@@ -26,7 +26,7 @@ module Amber::Controller
     end
   end
 
-  record Filter, precedence : Symbol, action : Symbol, blk : -> Nil, excepts : Array(Symbol) | Nil = nil do
+  record Filter, precedence : Symbol, action : Symbol, blk : -> Nil, excepts : Array(Symbol) = [] of Symbol do
   end
 
   # Builds a BeforeAction filter.
@@ -100,9 +100,7 @@ module Amber::Controller
     end
 
     private def except_filter_and_has_action?(filter : Filter, except_action : Symbol | Nil)
-      return false unless filter.action == :except
-      excepts = filter.excepts || [] of Symbol
-      excepts.includes?(except_action)
+      filter.action == :except && filter.excepts.includes?(except_action)
     end
   end
 end
--- a/src/amber/controller/filters.cr
+++ b/src/amber/controller/filters.cr
@@ -26,7 +26,7 @@ module Amber::Controller
     end
   end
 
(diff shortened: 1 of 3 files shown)


src/amber/controller/filters.cr
```diff
@@ -70,8 +80,9 @@ module Amber::Controller
       filters[filter.precedence] << filter
     end
 
-    def run(precedence : Symbol, action : Symbol)
+    def run(precedence : Symbol, action : Symbol, except_action : Symbol | Nil = nil)
       filters[precedence].each do |filter|
+        next if filter.action == :except && filter.try(&.excepts).try(&.includes?(except_action))
```
robacarp: this is a lot on one line, can this block be refactored to avoid using .try and be more readable?

src/amber/controller/filters.cr
```diff
@@ -18,13 +19,14 @@ module Amber::Controller
...
           @filters.run(:after, action)
+          @filters.run(:after, :except, action)
           @filters.run(:after, :all)
         end
       end
     end
   end
 
-  record Filter, precedence : Symbol, action : Symbol, blk : -> Nil do
+  record Filter, precedence : Symbol, action : Symbol, blk : -> Nil, excepts : Array(Symbol) | Nil = nil do
```
robacarp: :+1:

src/amber/controller/filters.cr
```diff
@@ -18,13 +19,14 @@ module Amber::Controller
         if self.responds_to? :after_filters
           self.after_filters
           @filters.run(:after, action)
+          @filters.run(:after, :except, action)
```
robacarp: Thanks for following the pattern here.

It seems like all of these `.run` statements could just live together in a single run method. Is that a worthy refactor @drujensen ?

drujensen: yeah, they could.  Up to @AndyRosenberg if he has time to do it.  If not, we can do it later.

src/amber/controller/filters.cr
AndyRosenberg: Got it. Made it more readable and took the the calls to try away in latest.

src/amber/controller/filters.cr
AndyRosenberg: Don't think I'll be able to get to this one yet.

src/amber/controller/filters.cr
drujensen: What do you think about defaulting the `excepts` array as an empty array and remove the Nil union here.  That way you don't have to do this on line 104.

AndyRosenberg: makes sense

PR #1113, opened 2019-06-02
Clean up assets

smadeja:
### Description of the Change

Generating a new Amber application creates the `public/dist` folders with some assets in it. Those files are not necessary and cause a number of issues.

- `public/dist/main.bundle.css` and `public/dist/main.bundle.js` get overwritten the first time the server is started using the `bin/amber watch` command, so they are never used for anything.
- `public/dist/images/logo.svg` is the same as `src/assets/images/logo.svg`. There's no need to duplicate the file in the repo, copying it should be part of the webpack build process.
- The `public/dist` folder gets put under version control, which is not desirable.

This PR fixes those issues by:
- Removing the `public/dist` directory from the app template
- Adding the `public/dist` directory to `.gitignore`
- Including the Amber logo in the webpack build process


smadeja:
The main issue is that webpack's build folder gets put under version control so after I've compiled assets Git's going to expect me to commit the entirety of Bootstrap. And Bootstrap is used in the default views so the precompiled bundles seem pretty useless to me. This is what the home page looks like when rendered with those bundles:

Screenshot 2019-06-03 at 07 09 25

drujensen:
@smadeja looks like a bug.  It should render properly without requiring the assets to be compiled.

What if we add the `/public/dist' to the `.gitignore` but we fix and keep them for the initial project?

elorest:
> @smadeja looks like a bug. It should render properly without requiring the assets to be compiled.
> 
> What if we add the `/public/dist` to the `.gitignore` but we fix and keep them for the initial project?

@smadeja @drujensen That seems like a good solution to me. My use case is to compile binary for target machine and then move everything over to embedded linux system without amber or node.

smadeja:
I'm sorry, but I still think those bundles should be removed. They are going to be a nightmare to maintain. They already are — the default template keeps changing and no one remembers about the bundles. For the vast majority of users they are completely useless as they get overwritten the first time they run `bin/amber watch`. To get the home page to display properly with the default bundles you'd have to either include all of Bootstrap 4 in them, or manually put some styles together which is just not worth it.

I don't want to sound aggressive, but I really don't see any reason to keep those files lying around.

smadeja:
Webpack is part of the default app template. It is expected to compile assets defined in `src/assets`. This is the default. The template needs to make some assumptions and then act on them. If it assumes that webpack is used to manage assets, then webpack should be used to manage assets. And if there's something wrong with webpack installation, it should be visible right away. This is just the default, it should cover the most common use case.

smadeja:
@elorest I understand your use case, but I'm trying to think about the future here. ;-)

drujensen:
@smadeja I think I agree with you.  If you don't have webpack/node installed then you get the unstyled version.  Its similar to what rails generators produces.  @elorest Are you ok with this change?

elorest:
> @elorest I understand your use case, but I'm trying to think about the future here. ;-)

@smadeja As you pointed out they get overwritten the first time `amber watch` is run. The default template shouldn't be changing that much? Because of that I don't see what harm there is in having a working default bundle that may need to be updated once a year from new compiled changes. 

If we used crystal to compile this I would be happy with removing the compiled assets. Due to all the extra deps I think it's prudent to continue to include precompiled assets.

Worst case scenario they'll end up being a tiny bit out of date but be replaced as soon as someone runs `amber watch`.

If after this @drujensen still thinks it's worth removing completely I'll remove my change request and approve.

smadeja:
If you decide you want to keep the bundles, I can modify this PR to not remove them.

src/amber/cli/templates/app/public/dist/images/logo.svg
```diff
@@ -1 +0,0 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="502" height="566"><g fill="none" fill-rule="evenodd"><path fill="#F3C095" d="M.117895 142.08393L250.562495.91603l-184.7541 305.3151-65.5352 117.9262z"/><path fill="#FFFEFC" d="M250.8729974 565.0744177L.2832957 424.5200163l65.5325685-118.0244418L250.5625196.9045315l250.5896836 140.5544042.1552393 282.084939z"/><path fill="#FBE9DB" d="M65.808395 306.23113L250.562495.91603l170.6768 203.7692z"/><path fill="#F9DEC9" d="M.273195 424.15733l-.1553-282.0734 65.6904 164.1472z"/><path fill="#FEFBF6" d="M.273195 424.15733l65.5352-117.9262 185.0646 258.8319z"/><path fill="#FFFEFC" d="M339.250895 434.67373l-273.4425-128.4426 355.4309-101.5459z"/><path fill="#FDF3EA" d="M250.872995 565.06303l-185.0646-258.8319 273.4425 128.4426z"/><path fill="#FCF0E4" d="M339.250895 434.67373l81.9884-229.9885 80.0782 219.21z"/><path fill="#FBE9DB" d="M250.872995 565.06303l88.3779-130.3893 162.0666-10.7785z"/><path fill="#FFF" d="M501.317495 423.89523l-80.0782-219.21 79.9231-62.8635z"/><path fill="#FDFCFA" d="M421.239295 204.68523L250.562495.91603l250.5999 140.9057z"/></g></svg>
```
elorest: Is this svg dynamically generated as well?

smadeja: It's not, but there's [the same svg in the assets directory](https://github.com/amberframework/amber/blob/master/src/amber/cli/templates/app/src/assets/images/logo.svg), so I'm assuming it's meant to be managed by Webpack.

PR #1121, opened 2019-07-10
Add upload file multiple support, fix more problem about route scope.

wndfly:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 /.shards/
 /.vscode/
 .DS_Store
-lib/
+lib
 .shards/
 /amber
 /tmp
--- a/src/amber/dsl/router.cr
+++ b/src/amber/dsl/router.cr
@@ -28,7 +28,6 @@ module Amber::DSL
     macro namespace(scoped_namespace)
       scope.push({{scoped_namespace}})
       {{yield}}
-      scope.pop
     end
 
     {% for verb in RESOURCES %}
--- a/src/amber/dsl/server.cr
+++ b/src/amber/dsl/server.cr
@@ -12,4 +12,10 @@ module Amber::DSL::Server
       end
     {% end %}
   end
+
+  macro namespace(scoped_namespace, value = :web)
+    router.draw {{value}}, {{scoped_namespace}} do
+      {{yield}}
+    end
+  end
 end
--- a/src/amber/pipes/static.cr
+++ b/src/amber/pipes/static.cr
@@ -10,7 +10,8 @@ module Amber
       def call(context : HTTP::Server::Context)
         return allow_get_or_head(context) unless method_get_or_head?(context.request.method)
 
-        original_path = context.request.path.not_nil!
+        scope = get_current_scope(context)
+        original_path = context.request.path.not_nil!.gsub scope, ""
         request_path = URI.unescape(original_path)
 
         # File path cannot contains '\0' (NUL) because all filesystem I know
@@ -38,6 +39,10 @@ module Amber
         end
 
         call_next_with_file_path(context, request_path, file_path)
+      end
+
+      private def get_current_scope(context)
+        context.request.matched_route.payload.try { |r| r.scope.to_s } || ""
       end
 
       private def dir_path?(path)
--- a/src/amber/router/params.cr
+++ b/src/amber/router/params.cr
@@ -5,7 +5,9 @@ require "./file"
 module Amber::Router
   module Types
     alias Key = String | Symbol
-    alias Files = Hash(String, Amber::Router::File)
+    alias File = Hash(String, Amber::Router::File)
+    alias Files = Hash(String, Array(Amber::Router::File))
+    alias UploadFile = File | Files
     alias Params = Hash(String, String)
   end
 
@@ -15,7 +17,7 @@ module Amber::Router
     MULTIPART_FORM   = "multipart/form-data"
     APPLICATION_JSON = "application/json"
 
-    @files = Types::Files.new
+    @files : Types::UploadFile?
     @multipart : Types::Params?
     @json : Types::Params?
     @form : HTTP::Params?
@@ -34,7 +36,7 @@ module Amber::Router
 
     def files
       multipart unless @multipart
-      @files
+      @files ? @files.not_nil! : Types::Params.new
     end
 
     def []=(key : Types::Key, value)
--- a/src/amber/router/parsers/multipart.cr
+++ b/src/amber/router/parsers/multipart.cr
@@ -1,19 +1,40 @@
 module Amber::Router::Parsers
   module Multipart
-    def self.parse(request : HTTP::Request) : Tuple(Types::Params, Types::Files)
+    def self.parse(request : HTTP::Request) : Tuple(Types::Params, Types::UploadFile)
       multipart_params = Types::Params.new
-      files = Types::Files.new
+      files = Types::File.new
 
       HTTP::FormData.parse(request) do |upload|
         next unless upload
+
         filename = upload.filename
         if filename.is_a?(String) && !filename.empty?
-          files[upload.name] = Amber::Router::File.new(upload: upload)
+          files = get_upload_files(files, upload)
         else
+          next if multipart_params.has_key? upload.name
+
           multipart_params[upload.name] = upload.body.gets_to_end
         end
       end
       {multipart_params, files}
+    end
+
+    def self.get_upload_files(files : Types::UploadFile, upload : HTTP::FormData::Part) : Types::UploadFile
+      if files.has_key? upload.name
+        if files.is_a? Types::File
+          first_upload = files[upload.name]
+          files = Types::Files.new
+          files[upload.name] = [first_upload]
+        end
+
+        files = files.as Types::Files
+        files[upload.name] << Amber::Router::File.new(upload: upload)
+      else
+        files = files.as Types::File
+        files[upload.name] = Amber::Router::File.new(upload: upload)
+      end
+
+      files
     end
   end
 end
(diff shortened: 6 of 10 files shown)


src/amber/router/params.cr
```diff
@@ -34,7 +35,7 @@ module Amber::Router
 
     def files
       multipart unless @multipart
-      @files
+      @files.not_nil!
```
drujensen: Do we want an exception if this is nil or should it return an empty array?

src/amber/router/parsers/multipart.cr
```diff
@@ -1,16 +1,32 @@
...
-      files = Types::Files.new
+      files = Types::File.new
 
       HTTP::FormData.parse(request) do |upload|
         next unless upload
+
         filename = upload.filename
         if filename.is_a?(String) && !filename.empty?
-          files[upload.name] = Amber::Router::File.new(upload: upload)
+          if files.has_key? upload.name
```
drujensen: I'm struggling to follow this logic.  Can you refactor this down into smaller methods?

eliasjpr: I agree we can separate these?

src/amber/router/router.cr
```diff
@@ -29,9 +29,9 @@ module Amber
       end
 
       def draw(valve : Symbol, namespace : String)
+        scope = Scope.new
```
eliasjpr: Can we have a separate PR for the scope fix makes it easier to review and manage

alex-lairan: The scope is already created line `15`

https://github.com/amberframework/amber/pull/1121/files#diff-d4d79b3524306c32cadbb606d98afc53R15

src/amber/router/router.cr
```diff
@@ -29,9 +29,9 @@ module Amber
...
         scope.push(namespace)
         with DSL::Router.new(self, valve, scope) yield
-        scope.pop
```
alex-lairan: Why not poping this scope ?

src/amber/router/params.cr
wndfly: That's a good idea.

src/amber/router/parsers/multipart.cr
wndfly: I'll trying to refactor it. Thank you.

src/amber/router/router.cr
wndfly: Because, It's using in another instance.

alex-lairan: But by using another instance, you erase previous scopes, like 
```
namespace "/foo" do
  namespace "/bar" do
  end
end
```

Here, `/foo` vanished.

wndfly: @alex-lairan I pushed new commits, that contain code modified with namespace.

src/amber/router/scope.cr
```diff
@@ -1,9 +1,9 @@
 module Amber
   module Router
     class Scope
-      @stack = [] of String
+      property :stack
```
alex-lairan: Why leak the internal into the external world ?

No one should know `Scope` use an `Array`

src/amber/router/scope.cr
```diff
@@ -1,9 +1,9 @@
...
 
-      def initialize
+      def initialize(@stack : Array(String) = [] of String)
```
alex-lairan: Ok for injection, but `@stack : Array(String) = [] of String` should be simplified by

`@stack = [] of String`

src/amber/router/router.cr
```diff
@@ -25,13 +24,11 @@ module Amber
 
       # This registers all the routes for the application
       def draw(valve : Symbol)
-        with DSL::Router.new(self, valve, scope) yield
+        with DSL::Router.new(self, valve, Scope.new) yield
```
alex-lairan: Using `Scope.new`, you lose all benefits of nested scopes.

src/amber/router/router.cr
```diff
@@ -25,13 +24,11 @@ module Amber
...
       end
 
       def draw(valve : Symbol, namespace : String)
-        scope.push(namespace)
-        with DSL::Router.new(self, valve, scope) yield
-        scope.pop
+        with DSL::Router.new(self, valve, Scope.new([namespace])) yield
```
alex-lairan: Why not use the current scope ?

.gitignore
```diff
@@ -9,4 +9,7 @@ lib/
 /tmp
 shard.lock
 package-lock.json
-node_modules
\ No newline at end of file
+node_modules
+lib
```
alex-lairan: lib already ignored

.gitignore
```diff
@@ -9,4 +9,7 @@ lib/
...
+.dir-locals.el
+*~
```
alex-lairan: This is local system element, it should be ignored by your git.

src/amber/pipes/static.cr
```diff
@@ -10,7 +10,10 @@ module Amber
       def call(context : HTTP::Server::Context)
         return allow_get_or_head(context) unless method_get_or_head?(context.request.method)
 
-        original_path = context.request.path.not_nil!
+        scope = ""
+        context.request.matched_route.payload.try { |r| r.scope.to_s }
```
alex-lairan: This line seems complicated, and not very Demeter friendly.

Add methods that helps comprehension.

src/amber/router/router.cr
wndfly: When I used `amber`, I discovered [a problem at the method](https://github.com/amberframework/amber/blob/3465f1dced59426d1f01eeac84821a95429cfa18/src/amber/router/router.cr#L59) with route match which doesn't add `scope`, so I modified more about `scope`. 
Maybe I should modify it [here](https://github.com/amberframework/amber/blob/3465f1dced59426d1f01eeac84821a95429cfa18/src/amber/router/router.cr#L40), But I think keep `scope` in every `route `might be better.

.gitignore
wndfly: I had a small changed, ignore `lib` directory too when it's symbolic link.

.gitignore
wndfly: I removed it.

src/amber/pipes/static.cr
wndfly: I rewrites and fixed it here in new commit.

src/amber/router/scope.cr
wndfly: Thanks for your idea, I modified it in new commit.

src/amber/router/scope.cr
wndfly: I change it back.

src/amber/dsl/server.cr
```diff
@@ -12,4 +12,10 @@ module Amber::DSL::Server
       end
     {% end %}
   end
+
+  macro namespace(scoped_namespace, value = :web)
```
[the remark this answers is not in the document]
wndfly: I swapped them because I want used it like this.
`
namespace "/foo" do
  namespace "/bar" do
  end
end

namespace "/foo1", :web do
  namespace "/bar" do
  end
end
`

src/amber/pipes/static.cr
```diff
@@ -43,6 +41,11 @@ module Amber
         call_next_with_file_path(context, request_path, file_path)
       end
 
+      private def get_current_scope(context)
+        scope = context.request.matched_route.payload.try { |r| r.scope.to_s }
+        scope ? scope : ""
+      end
```
alex-lairan: Can be simplified by `context.request.matched_route.payload.try { |r| r.scope.to_s } || ""`

src/amber/router/scope.cr
```diff
@@ -1,9 +1,9 @@
 module Amber
   module Router
     class Scope
-      property :stack
+      @stack = [] of String
```
alex-lairan: We could simply remove this line I think.

src/amber/pipes/static.cr
wndfly: Thank you, I changed it.

src/amber/router/scope.cr
wndfly: Sure, that's good idea. I removed.

src/amber/dsl/server.cr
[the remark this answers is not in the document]
alex-lairan: And what happen for 

```
routes :web do
  namespace "/foo", :api do
    get "/bar" ...
  end
end
```

This is inconsistant and allow developpers to do crappy things.

src/amber/dsl/router.cr
```diff
@@ -28,7 +28,6 @@ module Amber::DSL
     macro namespace(scoped_namespace)
       scope.push({{scoped_namespace}})
       {{yield}}
-      scope.pop
```
alex-lairan: You push but not pop, this will add more and more namespaces.

eliasjpr:
@wndfly can you add the description to the PR and follow the template this helps the team evaluate the changes and intentions behind the changes. For instance I am not sure I understand the benefits of having then namespace 

```crystal
namespace "/foo" do
  # ... routes here
  namespace "/bar" do
  # ... routes here
  end
end
```

versus repeating scopes

```crystal
scope "/foo", :web do
# ... routes here
end

scope "/foo/bar", :web do
# ... routes here
end
```

`namespace` might remove duplication, but it is also more verbose IMO and harder to read and understand when there is multiple nested namespaces

Having the ability to upload multiple files is 💯 good with me, it would be great to have them on a separate PR (the files part would have already been merged by now).

drujensen:
I'm going to close this PR.  The file upload part would be nice as a separate PR.  Mixing the routing change to the namespace is killing this PR.

PR #1126, opened 2019-08-07
Crystal v0.30.0 changes

drujensen:
From @bcardiff original PR:

A Crystal::VERSION check around `URI.decode` should be added if you want amber to be warning free but also keep working in 0.29.0.

You can double check in crystal-lang/crystal#7997 which encode/decode method apply better.

A couple of dependencies need to be updated also as stated in https://github.com/bcardiff/amber/commit/f28b442ac1a598cc6377eb98b13a1ed5e6f0bd96 after each of them are released. Do not use my forks as dependencies.

* https://github.com/mosop/callback/pull/3
* https://github.com/mosop/cli/pull/19
* https://github.com/TechMagister/liquid.cr/pull/37
* https://github.com/will/crystal-pg/pull/178
* https://github.com/crystal-lang/crystal-mysql/pull/80
* https://github.com/crystal-lang/crystal-sqlite3/pull/43
* https://github.com/mosop/teeplate/pull/16
* https://github.com/amberframework/granite/pull/353
* https://github.com/amberframework/citrine-i18n/pull/6


drujensen:
I forked 3 of @mosop repos for now.  We will move back to his repo if/when he responds.

src/amber/controller/filters.cr
```diff
@@ -80,6 +80,12 @@ module Amber::Controller
       filters[filter.precedence] << filter
     end
 
+    def each(&block : {Symbol, Array(Filter)} -> _)
```
elorest: What was the point of adding this?

elorest:
Looks like this is a breaking change because of ruby 0.30.0 and the new version of granite correct?

src/amber/controller/filters.cr
drujensen: it's now required to implement abstract methods in Crystal.

drujensen:
@elorest yes, crystal 0.30.0 changes and the new DSL for Granite.  thanks for approving.

nsuchy:
> Looks like this is a breaking change because of ruby 0.30.0 and the new version of granite correct?

A bit late to reading over the changes. Did you mean ruby or crystal 0.30.0? Just making sure I don't need to read over something ruby specific :)

drujensen:
@nsuchy that would be a really old version of ruby if it was.  ;-)

PR #1140, opened 2019-09-23
Change flash message grammar

ashishbista:
### Description of the Change

Currently, if we create resources using `amber g scaffold` command, the controllers set flash messages like below:

```  
Created project successfully.
```

The pull request would change it to:
```
Project has been created successfully.
```

The same message in Rails scaffold would appear as:
```
Project was successfully created.
```
### Alternate Designs
```
Project has successfully been created.
```


ashishbista:
@robacarp I agree that "successfully" is not necessary. That was just some Rails legacy in me :). I have updated the PR. Thank you for reviewing the PR.

robacarp:
@ashishbista that is not a failure that could be caused here. I believe the build container needs to be updated somehow.

drujensen:
From @bcardiff https://travis-ci.community/t/fails-to-link-against-libgc-a-in-crystal-0-31-0/5197

maybe change target os in `.travis.yml`?

PR #1147, opened 2019-09-29
Fix specs

drujensen:
### Description of the Change

This fixes several of the specs by moving the generation and cleanup of the sample app inside the `it` statement.  It looks like the way that the code inside of a `describe` block no longer runs sequentially so the cleanup was running before the `it` blocks causing them to fail.

This also marks 6 test suites as Pending.  We will need to further investigate why these are failing.

Fixed by @samholst and @bbaragar 
  Amber::Recipes::RecipeFetcher
  Amber::CLI::MainCommand::New
  Amber::Pipe::Static
  Amber::Pipe::CORS
  Amber::Pipe::CSRF
  Amber::DSL::Server


elorest:
Have you found any notes on the change from sequential. It's possible that there's a flag or something to change that now?

Those pipeline's will hopefully all have the same solution. I'll look more into this tomorrow. Still dealing with jetlag from Germany.

drujensen:
I believe this is the issue causing some of the problems.

https://github.com/crystal-lang/crystal/issues/8235

elorest:
@drujensen I'm hoping we can find a solution that allows for setup and teardown after a group rather than each spec. Looks like this affected a lot of the crystal community. Sort of seems like a step backwards.

ashishbista:
With this PR https://github.com/crystal-lang/crystal/pull/8238 merged, does it have any impact here?

elorest:
Yay

PR #1150, opened 2019-10-04
updated crystal version and added 404 to static pipe spec

samholst:
### Description of the Change

As I fix a pending spec, I'll put up a PR for each one I get to.

### Alternate Designs

n/a

### Benefits

fixes static test 404 spec.

### Possible Drawbacks

n/a


elorest:
Probably not a good idea to do them one file at a time, since the tests will show as failing for all of them, since the the rests will fail.

spec/amber/pipes/static_spec.cr
```diff
@@ -22,7 +23,7 @@ module Amber
 
         response = create_request_and_return_io(static, request)
 
-        response.body.should eq "Not Found\n"
```
elorest: Did this change in this version of crystal?

samholst: Yeah looks like it was changed, the most recent version of master has it with a new method (https://github.com/crystal-lang/crystal/blob/master/src/http/server/response.cr#L144) which prepends the status code before the message.

PR #1151, opened 2019-10-08
Fix more specs

samholst:
### Description of the Change

Fixed some pending tests for:

Amber::Pipe::Static
Amber::Pipe::CSRF
Amber::Recipes::RecipeFetcher
Amber::DSL::Server

### Alternate Designs

n/a

### Benefits

Test are passing now.

### Possible Drawbacks

n/a

PR #1152, opened 2019-10-08
Bb/fix more specs

bbaragar:
### Description of the Change

Fixed some pending tests

### Alternate Designs

n/a

### Benefits

Tests are passing now

### Possible Drawbacks

n/a


spec/amber/cli/commands/init_spec.cr
```diff
@@ -35,19 +38,27 @@ module Amber::CLI
...
+    #     dirs("../../"+TESTING_APP).sort.should eq dirs(APP_TEMPLATE_PATH).sort
+    #     cleanup
+    #   end
+    # end
 
     describe "Database settings" do
       %w(pg mysql sqlite).each do |db|
         it "generates #{db} correctly" do
           scaffold_app(TESTING_APP, "-d", db)
+          Dir.cd("../../")
```
drujensen: checkout @samholst PR.  https://github.com/amberframework/amber/pull/1151/files#diff-6ffd50128311c1b52dee3f90523a73a1R14  Maybe you can use that method?

bbaragar: Pushed up new changes. It should be good to go.

drujensen:
Thanks @bbaragar 👏

PR #1157, opened 2019-10-13
Remove process_count option

### LINKED ISSUE #1148: Remove Cluster Option
# Remove Cluster Option
With the addition of the `-Dpreview_mt` crystal cluster option will not be necessary. Instead we can use the process count setting to easily set the `CRYSTAL_WORKERS` environment variable when starting the amber server.

ashishbista:
Fixes #1148


eliasjpr:
@elorest I think we should consider removing this ASAP and instead provide an alternative approach for the folks that still needs this. 

Example: Alternatively this can be implemented as:

```crystal
Amber.env = "production"
Amber::Server.start

System.cpu_count.times do |i|
  Process.fork do
    Amber::Server.start
  end
end

sleep
```

drujensen (approved):
lgtm

crimson-knight:
Since multithreading is so nowhere near being considered ready for production, I'm going to close this out and we'll leave the clustering mode the way it is currently.

PR #1159, opened 2019-10-20
Change log time to local rather than utc

### LINKED ISSUE #1154: Log time is UTC not local
### Description

When I run the server, the displayed time in logs doesn't match my local time. This is kinda annoying, and may led users to confusion.

### Steps to Reproduce

1. Just run `amber watch`

**Expected behavior:** The time in the logs should match the local time.

**Actual behavior:** The times is always UTC

**Reproduces how often:** Always.

### Versions
**Crystal**
Crystal 0.31.1 [0e2e1d067] (2019-09-30)

LLVM: 8.0.0

**Amber**
Amber CLI (amberframework.org) - v0.30.1

**OS**
Operating System: Linux Mint 18.3
Kernel: Linux 4.15.0-64-generic
Architecture: x86-64

MatheusRich:
### Description of the Change
Fixes #1154 
This is my first "code" PR. I think it's very simple. I just changed the log method to write `Time.local` rather than `Time.utc`.

### Alternate Designs
This seemed pretty straight forward to me.

### Benefits
Users from non-UTC timezone can now see logs on their local time :smile: 

### Possible Drawbacks
I can't see a possible drawback. Maybe add a configuration so users could choose in what timezone they want to see their logs?

### DIFF
--- a/src/amber/environment/logger.cr
+++ b/src/amber/environment/logger.cr
@@ -13,7 +13,7 @@ module Amber::Environment
 
     def log(severity, message, progname = nil, color = @color)
       return if severity < level || !@io
-      write(severity, Time.utc, "#{progname || @progname} |".colorize(color).to_s, message)
+      write(severity, Time.local, "#{progname || @progname} |".colorize(color).to_s, message)
     end
   end
 end


elorest:
Is there any precident for this? UTC is pretty much the standard for dealing with times in programming. I'd suggest adding a setting under logging for which timezone to log with.

MatheusRich:
@elorest I think it's useful to have local time in logs, but making this on a setting it's even better. Any guidelines to me to do this? Is there something similar so I can take a look at it?

robacarp:
I think I’d prefer to have the default be in UTC. In development I can see the thought behind wanting it to be in local time... but I prefer not to have the time in the logs at all in development.

elorest:
> @elorest I think it's useful to have local time in logs, but making this on a setting it's even better. Any guidelines to me to do this? Is there something similar so I can take a look at it?

You could add a setting in these files and then modify the logger that you've already been playing with to do something with the setting.

https://github.com/amberframework/amber/blob/master/src/amber/cli/templates/app/config/environments/development.yml.ecr#L5

https://github.com/amberframework/amber/blob/master/src/amber/environment/settings.cr

drujensen:
@MatheusRich I'm going to close this PR.  If you decide to create a new setting to log local time, please open a new PR.  Thanks for your contribution.

PR #1162, opened 2019-10-25
Updated minified js file amber.min.js

seanwatters:
### Description of the Change

Updated the minified amber.js file to match my most recent pull request for the main amber.js file (it also appears that all previous changes to the main js file before my recent pull request weren't updated).

### Benefits

It is consistent with out main js file. 

### Possible Drawbacks

It may be confusing if my other pull request isn't merged first or at all.


drujensen:
@seanwatters Thanks!

PR #1164, opened 2019-10-31
Added /api scope to API routes.

seanwatters:
### Description of the Change

The JSON API generator currently routes to the pluralized name of the model. The issue here is, that route can no longer be used for views.

i.e

Someone creates a 'Post' JSON API. They now want to use 'Posts' for the the view route, but because that route belongs to the API, they can't. 

Rails also uses the '/api/posts/` style. 

### Alternate Designs

Considered adding version number (/v1/api/posts) but the current fix resolves my main concern. (still totally can if we think that's a good idea)

There may be a better namespacing option as well, where we could just specify that in the routes file for APIs. I didn't dig too deep in the docs to see if we have namespacing functionality built out already, but it would be a super easy fix to update if we did. 

### Benefits

Doesn't lock down routes that could be used for views. 

### Possible Drawbacks

Don't see any.


seanwatters:
When/if merged, I'll go update the docs.

seanwatters:
I know why the tests are failing (checking a route that isn't there anymore for the API CRUD) and how to fix it, but it's unclear where the tests need to be updated. 

The test that's failing isn't included in the amber spec from the main repo. 

Haven't worked much with travis-ci before, so this may be super straightforward fix that I'm just not aware of.

seanwatters:
Got help from @elorest tracking down the test failure. Will add fix for the tests this evening. 

Also going to look deeper into the namespacing options.

seanwatters:
Okay, sorry for so many changes on this PR. 

Instead of hard coding '/api/post' for every API route, I changed it to use an "/api" scope. 

```
routes :api, "/api" do
    resources "/posts", API::PostController, except: [:new, :edit]
end
```

and changed the file structure for the API generator template to be 

```
|- controllers
        |- api
            |- post_controller.cr
```

With a `API::PostController` class declaration in the controller file.

alex-lairan:
I don't understand why add the '/api' to the route ?

If I don't have view (like almost all app with front and back separated) this would add useless information.

seanwatters:
@alex-lairan This is how rails sets up their API only applications. They also ad a version tag i.e `localhost:3000/api/v1/posts`. It helps to differentiate between what is JSON and what is a view, even if you're writing an API only amber app.

robacarp:
@seanwatters Thank you for your effort here. Rails compatibility isn't a directive of the amber project, but I agree that this is a good change. It introduces a sane default to the generators and establishes a convention. 

However, the amber project is also in the process of removing or minimizing the inbuilt generators and leveraging the [recipe system](https://docs.amberframework.org/amber/cli/recipes#recipes) for all generators. 

@elorest do you think this should be merged? or directed at an api recipe instead?

elorest:
@robacarp I'm fine with this being merged in theory. 

> However, the amber project is also in the process of removing or minimizing the inbuilt generators and leveraging the recipe system for all generator

Default generators aren't going to be removed. However I agree that anything more than default crud should be a recipe. Since API's already exist and are pretty standard right now in the development world right now. I vote we merge this. Since the original code was written by @drujensen though I'd like his opinion on this PR.

drujensen:
I'm good with this change as wll.  Thanks @seanwatters for your contribution.

src/amber/cli/generators/api_controller.cr
```diff
@@ -13,7 +13,7 @@ module Amber::CLI
 
     private def add_routes
       add_routes :api, <<-ROUTE
-        resources "/#{name_plural}", #{class_name}Controller, except: [:new, :edit]
+        resources "/#{name_plural}", API::#{class_name}Controller, except: [:new, :edit]
```
drujensen: I'm wondering if we can specify the namespace when we perform the generate.  That way if someone doesn't want the `API` namespace or prefers a different namespace, they can specify it.

What if we support name spaces i.e. `amber generate api API::User` instead of hard coding it?

This should work for all generators so `amber g scaffold Admin::User` would create the `/admin` directory and properly create the namespaced models, controllers, views, etc.

src/amber/cli/helpers/helpers.cr
```diff
@@ -1,11 +1,19 @@
 module Amber::CLI::Helpers
   def add_routes(pipeline, route)
     routes = File.read("./config/routes.cr")
-    replacement = <<-ROUTES
-    routes :#{pipeline.to_s} do
-    #{route}
-    ROUTES
-    File.write("./config/routes.cr", routes.gsub("routes :#{pipeline.to_s} do", replacement))
+    if pipeline.to_s == "web"
```
drujensen: not liking the hardcoded pipelines `web` and `api` with special logic here.

seanwatters: Would you prefer something like this?

```
def add_routes(pipeline, route)
    scope = "routes: #{pipeline.to_s == "web" ? pipeline.to_s ? pipeline.to_s + "\"/api\""} do"
    routes = File.read(".config/routes.cr")
    replacement = <<-ROUTES
    #{scope}
    #{route}
    ROUTES
    File.write(".config/routes.cr", routes.gsub(scope, replacement)
end
```

or do we not want `routes :api, "/api" do` at all?

src/amber/cli/generators/api_controller.cr
seanwatters: I can look into adding this functionality.

If (like @alex-lairan said) we are moving away from the dependence on generators, would it be better to drop the namespacing aspect in this PR altogether, and just keep the "/api" scope?

I see the merit for something like the API generator but if scaffolds are mostly going to depend on recipes, it may not be worth the effort for all other generators.

Though, the namespacing option could probably be dropped into the recipe generators too.

src/amber/cli/helpers/helpers.cr
drujensen: hhmm, good question. I am ok with the `routes :api do` since we have a different set of middleware.  I guess i'm in agreement with @alex-lairan that having the hard coded `/api` prefix is where I'm uncomfortable.  That seems inflexible to me.

Is it necessary to have the "/api" namespace here?  Is that to avoid conflicts?

src/amber/cli/generators/api_controller.cr
drujensen: I don't expect the generators will move to recipes anytime soon.  There is a lot of work needed to make them more flexible and robust.

It would be great if you could specify the namespace when generating resources.  This is a fairly big change though.  It will pretty much touch all the generated code including the tables generated for the models, etc.  The tests should help make sure you don't break existing functionality.

src/amber/cli/helpers/helpers.cr
seanwatters: So the problem I was running into with the [amber-vue](https://github.com/seanwatters/amber-vue) and [reactive-amber](https://github.com/seanwatters/reactive-amber) recipes, is that if I create an API with the generator, I have to go manually rewrite the routes file and manually name space the APIs every time I generate a new one. 

If I generate a default API for posts, and want to use a "/posts" route for vue or react router, I can't because it's been taken.

seanwatters: The primary goal is to make it easier for people to use progressive frontend frameworks, with Amber, without running a separate node project or using separate domain for their API (or manually setting the "/api" scope every time you create a new JSON API).

drujensen: >  I have to go manually rewrite the routes file and manually name space the APIs every time I generate a new one.

That is a bug.  You shouldn't have to do that. You should only have to set the scope once.  When the resource is added to the `api` pipeline, it shouldn't overwrite the scope.  If that is what is happening, we should fix that instead of adding a hard coded scope.

drujensen: @seanwatters I created a PR to fix the bug.  https://github.com/amberframework/amber/pull/1165

Let me know if this resolves your issue.

seanwatters: > @seanwatters I created a PR to fix the bug. #1165
> 
> Let me know if this resolves your issue.

That looks great! I think that actually solves the overall problem better, and I can work on the namespacing issue we discussed above. 

The change does look like it's failing the tests though and I'm not sure why. It looks like there may be an issue with the `backreferences = true` (?)

seanwatters: I do also have a PR up on the documentation for the API Generator (there wasn't anything on the generators docs) and made an update to my previous API example that uses the generated API controller code, instead of what I had handwritten (before I knew there was an API generator) https://github.com/amberframework/docs/pull/167/files

drujensen:
@seanwatters Can we close this PR?  Open another one if you decide to handle namespaces in the generators.

seanwatters:
@drujensen Yep! That works for me.

PR #1166, opened 2019-11-10
Replace mention to yaourt by yay in README.

hugopl:
yaourt is deprecated and discontinued, yay seems the closer replacement.


drujensen:
Thanks @hugopl and @alex-lairan

PR #1169, opened 2019-12-14
Leverage Amber routing for authenticated paths

drujensen:
### Description of the Change

This replaces the `Authenticate` pipe to remove the list of paths in the middleware with a new pipeline that will require authentication.  This includes a second pipeline to load the `CurrentUser` in both the :web and new :auth routes.

To authenticate your routes, add them to the `:auth` pipeline instead of the `:web` pipeline.

This also removes the `RegistrationController` in favor of using the `new` and `create` restful methods in the `UserController`.  

This also cleans up the `UserController` to match our other controllers closer.

### Alternate Designs

Keep it as is.

### Benefits

Take advantage of Ambers routing for authenticated paths

### Possible Drawbacks

None I can think of


drujensen:
@eliasjpr @elorest @robacarp @marksiemers Can you review?  This removes the array of paths from the middleware that require authentication and uses an `auth` pipeline/routes instead.

drujensen:
Thanks @robacarp

PR #1170, opened 2019-12-15
Add Pagination Support

irlcatgirl:
This pull request adds a partial for pagination. It requires the user to use the `?page_number=` query string and it requires that they include a `page_count` variable in their controller before loading the view. Unlike the other layouts, the user will need to include `<%= render(partial: "layouts/_pagination.ecr") %>` into the appropriate template on their own. From there they are responsible for modifying their Granite/ORM queries to limit the rows returned and to skip rows based on the limit and page number.

I did as follows in my controller and it works out well.
```ruby
page_number = params[:page_number]?.try &.to_i64 || 0
offset = page_number * 10
count = Model.where(CONDITION).count.run
page_count = (count.as(Int64) / 10).ceil.to_i64
results = Model.where(CONDITION).limit(10).offset(offset)
```

The resulting output is:
image
image
image

To-do:

- [ ] Slang templates
- [ ] Unit tests
- [ ] Discuss view variable names and potential improvements (it's sloppy at the moment)
- [ ] Update documentation

### DIFF
--- a/src/amber/cli/templates/app/src/views/layouts/_pagination.ecr
+++ b/src/amber/cli/templates/app/src/views/layouts/_pagination.ecr
@@ -1,3 +1,7 @@
+<%
+   page_number = 0 unless page_number
+   page_count = 0 unless page_count
+%>
 <nav aria-label="Pagination">
     <ul class="pagination pagination-circle pg-blue">
         <% unless page_number == 0 %>
--- a/src/amber/cli/templates/app/src/views/layouts/_pagination.ecr
+++ b/src/amber/cli/templates/app/src/views/layouts/_pagination.ecr
@@ -1,0 +1,46 @@
+<nav aria-label="Pagination">
+    <ul class="pagination pagination-circle pg-blue">
+        <% unless page_number == 0 %>
+            <li class="page-item"><a href="?page_number=0" class="page-link">First</a></li>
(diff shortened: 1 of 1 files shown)


drujensen:
@catgirlsec The partial for pagination won't compile without the variables being defined.  Crystal won't let you get away with that (unlike interpreted languages).

catgirlsec:
Could I add a default value in the application_controller so page_number and page_count default to 0?

drujensen:
@catgirlsec I am not ok cluttering up the `ApplicationController` with these variables.  That would be a pretty hacky solution IMO.  This needs more thought.

catgirlsec:
Is https://github.com/amberframework/amber/pull/1170/commits/806904f61b703830e37df68dcfdaa29556f1c930 an acceptable workaround?

catgirlsec:
Okay neither of my attempts passed unit tests. It looks like I cannot use the unless macro to test for a value. Any ideas to test inside the template without it becoming a mess?

eliasjpr:
@catgirlsec I would recommend having pagination as a separate shard given that Amber should be agnostic of pagination implementations given the data model

catgirlsec:
At that point I would just make a blog post on using this blob. I don’t have time at the moment to maintain a crystal shard :( - it’s not really a model just local variables. Feel free to close this request if it’s a bad idea to add into Amber.

PR #1172, opened 2020-01-22
Updates Generated APP Shard Dependencies Version

eliasjpr:
Upgrades dependents shards version 

Depends On: https://github.com/amberframework/granite/pull/377


faustinoaq:
Hi @eliasjpr is something wrong with `db`?
Annotation 2020-01-22 082518

eliasjpr:
Will have to look at the dependency tree to see where that error is coming from

PR #1173, opened 2020-01-22
Update CI/deps to support Crystal 0.32.1

eliasjpr:
### Description of the Change

New releases for DB, PG, MySQL and SQLite3 have been release that are
breaking building new amber apps.

This updates the shard.yml to use the latest versions for these projects


drujensen:
probably need to fix the specs.  Crystal now supports before and after callbacks for single tests or a suite of tests.

alvarosilvac:
any follow up about this issue?

eliasjpr:
The reason why this is failing is because the generated APP in the granite spec stilll points to Amber master. if we point it to the commit of this PR it works.

alvarosilvac:
Waiting on this

eliasjpr:
@alvarosilvac I am still working on this, to not keep you waiting you can always point your shard.yaml to this commit `commit: df064789498449a94db189b2aa0133cfefca843e` instead of branch/version.

faustinoaq:
Hi I tested it, amber compiles fine, but app generated with it still has `DB` dependency issue.

BTW, I think we should change `.crystal-version` to `0.32.1` like `shard.yml`

Annotation 2020-01-24 085752
Annotation 2020-01-24 090010
Annotation 2020-01-24 090020

eliasjpr:
@faustinoaq the tests are passing in CI. I wonder what could be different in your machine and I think you might need to recompile your `amber` cli tool based on this branch

I also updated the `.crystal-version` file to match the shard. Good catch!

spec/build_spec_granite.cr
```diff
@@ -2,66 +2,44 @@ require "./build_spec_helper"
 
 module Amber::CLI
   describe "building a generated app using granite model and slang template" do
+    cleanup
```
drujensen: these should go in a `before_suite` block.  https://crystal-lang.org/api/0.32.1/Spec.html#before_each(&block)-class-method

eliasjpr: I cleaned it up a little, I didn't use the Spec.before_each or before_suite because I didn't look necessary to get the tests running. Also the Spec.before_each will run for all specs so I was affriad it might break spec in other directories

PR #1179, opened 2020-02-15
Feature/add cache control header

VanyaNeytrino:
### Alternate Designs

I want to add caching for static files. I ran into a problem: there are no helpers for loading static files of the form: main.1ba318ef18f6a915354b. css. Do I understand correctly that writing a simple helper that will read files of this type from public/dist is too easy a solution? For example, in Rails, the implementation of such a helper is much more difficult.

### Benefits

There will be no unnecessary requests to static files. The browser will read them from the cache

### Possible Drawbacks

bugs related to the fact that the browser will read a file from the cache, and not make requests


src/amber/pipes/static.cr
```diff
@@ -124,6 +125,10 @@ module Amber
         return
       end
 
+      private def add_cashe_header(env : HTTP::Server::Context)
```
alex-min: There's some spelling mistake here

drujensen:
@alex-min Is this still a WIP?

alex-min:
@drujensen, I'm not the author of this PR so I don't know if it's still a WIP, I just added a comment in it because I saw some typo in it.

The concept looks good to me though.

Issue #1174, opened 2020-01-25
Password validation is failing


drujensen:
### Description

Password validation is failing when trying to login.

### Steps to Reproduce

```sh
amber new blog
cd blog
amber g auth user
amber db create migrate
amber watch
```

1. Access http://localhost:3000
2. Create a new account test@test.com / password
3. Logout
4. Login with test@test.com / password

**Expected behavior:** [What you expect to happen]

User is logged in

**Actual behavior:** [What actually happens]

Error saying email or password is incorrect

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions

crystal 0.32.0 / amber 0.32.0

Brodan:
I'm also seeing this currently. When I check the DB I see the user created and the `hashed_password` column is completely blank, thus re-logging in fails. This is when using postgres if that helps.

andrewc910:
It looks like `User#password=` is not being invoked with `user = User.new user_params.validate!`(user_controller#create). Actually, it looks like no manual setters are invoked like this. 

Adding `user_params.validate!["password"]` below the initialization of the `User` object ensures `User#password=` is invoked. 

I believe this fix will take a deep dive into Granite or Amber Router. I implemented a basic `email=` to see if it would be invoked and it is not. I tried something like the following which failed to hit:
```crystal
def email=(email)
  @email = email
end
```

And i tried(just to see logs) which again failed:
```crystal
def email=(email)
  pp "EMAIL INVOKED"
  pp "EMAIL INVOKED"
  pp "EMAIL INVOKED"
  pp "EMAIL INVOKED"
  @email = email
end
```

It looks like either the router or granite is bypassing these setters(some how). I will open a PR with quick fix i provided above but i do believe their is some bigger underlying issue.

drujensen:
This is a granite issue.  Granite should call the getters and setters when assigning values.

PR #1182, opened 2020-02-16
Quick fix for auth template, see issue #1174

andrewc910:
Description of the Change
This PR is meant to be a quick fix for issue #1174
This isn't meant to be a permanent change. You can read my reasoning in the issue.

Alternate Designs
Any alternates would have to be in either Granite or Amber Router but i believe it's a granite issue as Param#validate! successfully returns the proper hash.

Benefits
Users of amberframework have access to a working auth template. Currently, the auth template builds but fails to actually work.

Possible Drawbacks
No drawbacks that i am aware of, minus being ugly. I would love if someone more experienced can let me know if i am wrong. But we still call Params#validate! to ensure the params are safe.

### DIFF
--- a/src/amber/cli/templates/auth/src/controllers/{{name}}_controller.cr.ecr
+++ b/src/amber/cli/templates/auth/src/controllers/{{name}}_controller.cr.ecr
@@ -19,6 +19,9 @@ class <%= class_name %>Controller < ApplicationController
 
   def create
     <%= @name %> = <%= class_name %>.new <%= @name %>_params.validate!
+    pass = <%= @name %>_params.validate!["password"]
+    <%= @name %>.password = pass if pass
+
     if <%= @name %>.save
       session[:<%= @name %>_id] = <%= @name %>.id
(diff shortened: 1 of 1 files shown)


andrewc910:
I suggest we also add a validation for presence of `password_digest`. This will ensure the user cannot be saved unless `User#password=` is invoked thus digesting the password.

drujensen:
@andrewc910 thanks for looking into this.

drujensen:
@andrewc910 The tests are failing because Crystal 0.33.0 has been released and travis uses the latest version.

andrewc910:
I mentioned in gitter but ill say it here. I didn't understand the line "Can you rebase this to master?". I believe the idea was to rebase so the branch is up-to-date with master without making new commits. Is that correct? I only ask because this is my second PR outside of work. Still learning.

Just so you can know what happened, on my local branch i did the following commands:
`git rebase -i origin master` 
`git push -f` -f was because I was now apparently behind my remote branch.

If anything is wrong, i can just open a new PR to ensure i don't mess something up! :)

PR #1188, opened 2020-03-24
Bump Kilt to latest version

andrewc910:
### Description of the Change

Update Kilt's version in the shard.yml

### Alternate Designs

None?

### Benefits

Added jbuilder to kilt engine. Currently, you cannot use jbuilder in an amber project due to dependency issues between kilt & amber.

### Possible Drawbacks

None?

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -28,7 +28,7 @@ dependencies:
 
   kilt:
     github: jeromegn/kilt
-    version: ~> 0.4.0
+    version: ~> 0.6.0
 
   liquid:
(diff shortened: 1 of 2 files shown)


drujensen:
@andrewc910 can you bump the version of quartz-mailer to 0.6.0 in https://github.com/amberframework/amber/blob/master/src/amber/cli/templates/app/shard.yml.ecr to fix the specs?

andrewc910:
Quartz mailer has been bumped up to 0.6.0 in the app template :+1:

PR #1191, opened 2020-04-12
crystal 0.34.0 support

drujensen:
### Description of the Change

Support Crystal 0.34.0

### DIFF
--- a/src/amber/cli/templates/app/shard.yml.ecr
+++ b/src/amber/cli/templates/app/shard.yml.ecr
@@ -23,7 +23,7 @@ dependencies:
 
   granite:
     github: amberframework/granite
-    version: ~> 0.19.0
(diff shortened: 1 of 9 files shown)


src/amber/cli/templates/app/shard.yml.ecr
```diff
@@ -18,12 +18,12 @@ targets:
 dependencies:
   amber:
     github: amberframework/amber
-    branch: master
+    branch: dj-crystal-34-support
```
drujensen, line 21: changed this to use this branch so tests will pass.  we need to change this back to master.

drujensen:
So frustrating.  Why do the tests pass locally but fail on travis?  I want to move back to using Docker.  It removes the environment and provides a consistent testing platform.

PR #1195, opened 2020-04-19
updating crystal version file to 0.34.0

alex-min:


drujensen:
darn.  thanks @alex-min

PR #1197, opened 2020-04-27
Plugins implementation

damianham:
### Description of the Change
This change adds a plugin feature to Amber to allow Amber application developers to install ready made code plugins to provide additional features to their applications.

Plugins can be installed from a github repo, a local directory path or a downloadable zip file.  This adds 1 shard to shard.yml in the app template to support modular applications and plugins.  This makes 4 changes to config/application.cr in the app template;

- the modular shard is loaded before loading the plugins
- all plugin initializers are loaded 
- all plugin implementations are loaded
- all plugin hooks are loaded

### Alternate Designs
N/A

### Benefits
Being able to add in plugins to provide extra features will allow Amber application developers to create rich and powerful applications with greater ease.

### Possible Drawbacks

A plugin author may not namespace their plugin routes and classes, although this should be a mandatory requirement in the plugin documentation (coming soon).

A plugin route/class namespace may clash with an application's existing namespaces.  An Amber application develop should read the plugin documentation in the plugin README.md to ensure they are aware of the namespace used in the plugin to ensure installing it does not clash with their existing namespaces.


src/amber/cli/templates/app/config/application.cr.ecr
```diff
@@ -12,6 +12,11 @@ require "./settings"
 require "./i18n"
 require "./database"
 require "./initializers/**"
+require "amber_render_module"
```
drujensen, line 15: Is there anyway for someone to create a plugin without it being in the modular format?

src/amber/cli/templates/app/shard.yml.ecr
```diff
@@ -33,6 +33,10 @@ dependencies:
     github: amberframework/jasper-helpers
     version: ~> 0.2.5
 
+  amber_render_module:
```
drujensen, line 36: This goes against the standard layout that the generators currently use.  Either we go one way or the other. I will raise this request to the core team and provide feedback.  If approved, we will rewrite the current generators to use the module approach.  If not, we will stick with the current approach and request that the plugins comply.

drujensen:
@amberframework/core-team This PR requires plugins to conform to a different layout than we currently support.  You can read the differences here: 

https://github.com/damianham/amber_render_module

Basically, the plugins will need to use `render_module` instead of `render`.  They will need to comply with the module layout which uses the name of the object as the base directory and then the model, view, controller, mailer, etc are placed in this directory instead of all the controllers being together in one directory. so `src/user/controller.cr` instead of `src\controller\user_controller.cr`.  Anything shared will go in a shared folder.

If we want to move to this module layout, I will support the change and we will rewrite the current generators to use this new layout.  We will rewrite the `render` method to look for the controller and view in this new structure.  I dont' think we should support both but let me know if you think that should be considered as well.

Please provide your feedback, so we can determine how we want to move forward.

src/amber/cli/templates/app/config/application.cr.ecr
damianham: I think the danger with not using the modular format for plugins is that there could easily be conflicts with existing application artifacts.  By separating plugins into a specific plugins folder there is no possibility of any clashes between existing artifacts in src/controller, src/models etc.

src/amber/cli/templates/app/shard.yml.ecr
damianham: I personally don't see why the two formats cannot co-exist.  They have done so in my Amber applications from the outset.  The argument to remain with the existing layout is that people coming from a Rails environment will feel comfortable with the similarity in the layout.  The argument for modular layout is better code management, Principle of Proximity etc.  From my experience when working on large Rails projects the lack of modular structure and Principle of Proximity is very frustrating, particularly when coming into a project as a contractor with a very large codebase.

I don't think we need to throw the baby out with the bath water.  We could quite easily make the code layout an option with `amber new` so the developer could chose to create an application with the standard layout or chose a modular layout.  We don't even need to have 2 versions of the templates in the amber cli source.  We can use recipes to create an app either from a standard recipe or from a modular recipe, e.g
```
$ amber new myapp  # create an app with the standard layout 
$ amber new -m myapp  # create an app with the modular layout - from a pre-defined recipe
```

robacarp:
I like the idea of organizing code by module a lot. 

But it seems like introducing a second render method would make that a harder to discover feature. Why not just `render module: :foo`?

damianham:
> I like the idea of organizing code by module a lot.
> 
> But it seems like introducing a second render method would make that a harder to discover feature. Why not just `render module: :foo`?

To be honest I think the render() macro should support rendering views from paths relative to the calling controller in order to support modular layout and fallback to src/views if the view is not found locally then there is no need for a render_module() macro or a module: flag.

I am not sure about `module: :foo`  - does that mean render the view from src/foo ?
That is certainly more flexible allowing you to render views from anywhere, but the ability to render a local view without any extra option parameters would be better than needing to add any extra options.

elorest:
@damianham 
> To be honest I think the render() macro should support rendering views from paths relative to the calling controller in order to support modular layout and fallback to src/views if the view is not found locally then there is no need for a render_module() macro or a module: flag.

Currently the macro is fairly configurable if you override the defaults. It defaults to the controller now but allows you to load any path you want if you give it the whole path.

drujensen:
@elorest @eliasjpr @robacarp @marksiemers @faustinoaq thumbs up or down for supporting this layout for plugins.  It uses a different renderer that expects grouping by business concern (module) instead of by developer concern (models, views, controllers, etc.)

Options if we decide to move forward with supporting both layouts:
- support two different renderers
- support a module flag in the current renderer
- automatically determine the layout in the current renderer

eliasjpr:
> If we want to move to this module layout, I will support the change and we will rewrite the current generators to use this new layout. We will rewrite the render method to look for the controller and view in this new structure. I dont' think we should support both but let me know if you think that should be considered as well.

I like the idea about modules and it seems that we don't want to support of another macro just to render modules with that said I agree with @drujensen to support this PR, lets try to maintain a consistent API.

eliasjpr:
**Just a thought:** We can take advantage of crystal modules to the convention of the path module `SomeModule::Controller.class.name.underscore => some_module`

andrewc910:
@damianham Okay, so how do i use this? It looks like i do `amber plugin add X`. What is that X? I tried pointing it at one of my repos, I got a nice error message that a plugin could not be generated.

From a quick glance at the code, it looks like i make a .plugins folder? Let's say i modify the graphql lib to have a .plugins folder.

As i can't find any documentation, what should the file hierarchy look like? How does this all hook into the user's application? Going to give Graphql plugin a wack and see what happens.

damianham:
@andrewc910 sorry there isn't any documentation for plugins yet - it still hasn't been accepted so it may not be merged into the amber framework codebase - hence (being soooo busy) I am a little reluctant to create documentation for a feature that may not exist.  This is the first draft of the plugin feature so may change quite a bit once it is part of the codebase.  This PR is to establish a base to build from.

I have fixed up and DRY'd the code and tested creating an app from a recipe in a github repo and a zip file and installing a plugin from https://github.com/amberplugin/authorize with
```
amber plugin add authorize
```
Plugins without a '/' in their name are assumed to be in the amberplugin organisation so ```authorize``` resolves to https://github.com/amberplugin/authorize.  Other options are;
```
amber plugin add amberplugin/authorize
amber plugin add https://github.com/amberplugin/authorize
```
The issue with your plugin is that the plugin folder in your repo is called ```amberplugin``` rather than ```plugin```.  The name of the folder matters so we can be sure we are generating from the right kind of repo. Take a look at https://github.com/amberplugin/authorize to see how a plugin should be organised.

The initializers and hooks are separate from the main plugin code and located in src rather than plugins because the developer will modify those whereas they may not touch the plugin code (apart from things like mailers) however a plugin should generate default initializers and hooks files.

```
require "../src/plugins/initializers/**"  # user modified initializers
require "../plugins/plugins"     # src/plugins/plugins.cr that loads each plugin - created/edited by the developer
require "../src/plugins/hooks/**"   #  user modified hooks
```
So after adding some plugins I have ;
```
$ ls plugins
authorize/ messaging/ plugins.cr social/

$ cat plugins/plugins.cr
require "./authorize/plugin"
require "./messaging/plugin"
require "./social/plugin"
```

andrewc910:
Thank you for the information. I don't need a full doc, what you just gave is exactly what I was looking for. I don't expect full documentation on a PR.

Installed the new version. Renamed the folder to `plugin` instead of `amberplugin`. The recursion issue still persists. `.plugins/lib/amber-graphql-test/lib/amber...etc`. This also occurs with your `authorize` template.

Did you intend to generate files wherever the plugin author wanted? I just tried files like `plugin/src/controllers/test_controller.cr` and it merged into the main controller files. TBH, i kind of prefer this over something like `amber-project/src/plugins/controller/...` as that would lead to having models, controllers, possibly views in 2 different spots. 

I also tried `config/initializers` and that worked as well. I personally would prefer plugins to actually install into the project instead of being partitioned into a plugins section but that is just me.  Similar how the rails generator will install devise. It doesn't go into the plugin folder. Devise init is in the same initializer folder as everyone else. Devise default controllers go to `app/controllers`, etc.

I also just tried my project [mochi](https://github.com/andrewc910/mochi). I think this is a good example as it is an authentication shard with a lot of files to generate at first. Currently it uses a standalone cli. I am ripping that out this week to make as it's own shard OR for a plugin as the shard is specifically for Amber. One thing I like about the cli is different commands & args. I support both Granite & Jennifer ORMs along with having authenticable, confirmable, invitable modules. You can install only what you want. `mochi g auth jennifer` or `mochi g invitable granite`. Would plugins be an all or nothing install or can we stagger installs without 10 different repos?

If this gets merged and this is the base, how do you envision plugins? Do you except a single shard that the top level is `src/` for all business related logic and `plugin/` for generated files or would it be better to have a separate repo like `mochi-amber-plugin` that is solely related to the plugin? I am just thinking about plugin repo organization.

damianham:
> Installed the new version. Renamed the folder to `plugin` instead of `amberplugin`. The recursion issue still persists. `.plugins/lib/amber-graphql-test/lib/amber...etc`. This also occurs with your `authorize` template.

I wonder if you are running the plugin command in a folder that is not an amber app.  Here are the commands I ran to create an app from a recipe and add the authorize plugin and the plugin artifacts were added to the apps plugins folder 
```
$ amber/bin/amber new -r damianham/amber_material_kit materialkit
$ cd materialkit/
$ ../amber/bin/amber plugin add authorize
```
.plugins/authorize/lib is a symbolic link to '..' so that is why it appears as an infinite recursion.

> Did you intend to generate files wherever the plugin author wanted? I just tried files like `plugin/src/controllers/test_controller.cr` and it merged into the main controller files. TBH, i kind of prefer this over something like `amber-project/src/plugins/controller/...` as that would lead to having models, controllers, possibly views in 2 different spots.

Yes the plugin author can generate whatever they want within their plugin folder e.g.

amberproject/plugins/authorize/controllers
amberproject/src/plugins/initializers/authorize.cr

and initialisation and hook files which are intended to be modified by the developer
amberproject/src/plugins/hooks/__hook files for authorize plugin__

The point of this separation is to distinguish between plugin code that the developer will modify (initializers and hooks) and plugin code that they may not need to modify.

> I also tried `config/initializers` and that worked as well. I personally would prefer plugins to actually install into the project instead of being partitioned into a plugins section but that is just me. Similar how the rails generator will install devise. It doesn't go into the plugin folder. Devise init is in the same initializer folder as everyone else. Devise default controllers go to `app/controllers`, etc.

I purposefully avoided that for easier plugin management so that the code in src and config is code that you modify and the code is plugins is code that you don't modify.

> I also just tried my project [mochi](https://github.com/andrewc910/mochi). I think this is a good example as it is an authentication shard with a lot of files to generate at first. Currently it uses a standalone cli. I am ripping that out this week to make as it's own shard OR for a plugin as the shard is specifically for Amber. One thing I like about the cli is different commands & args. I support both Granite & Jennifer ORMs along with having authenticable, confirmable, invitable modules. You can install only what you want. `mochi g auth jennifer` or `mochi g invitable granite`. Would plugins be an all or nothing install or can we stagger installs without 10 different repos?

A plugin could support different ORMs in the way we used to support Crecto and Granite in the model/scaffolding generators by switching on the ORM identifier.

> If this gets merged and this is the base, how do you envision plugins? Do you except a single shard that the top level is `src/` for all business related logic and `plugin/` for generated files or would it be better to have a separate repo like `mochi-amber-plugin` that is solely related to the plugin? I am just thinking about plugin repo organization.

I envisage each plugin has it's own repo to make it easier for anyone to contribute a plugin.

drujensen:
Went with #1219 instead.  Closing this.

PR #1210, opened 2020-06-10
Crystal 0.35.0

drujensen:
### Description of the Change

These changes are needed to upgrade amber to support crystal 0.35.0 without dropping support for previous version.

JSON.mapping/YAML.mapping macros are deprecated. Either include the added dependency or migrate to Serializable to avoid warnings.

The updates of crystal-pg v0.21.1, crystal-mysql v0.11.1 should be applied to granite also.

The following dependecies need to be updated after the next release

Done:
cli - drujensen/cli#4
teeplate - mosop/teeplate#18
liquid.cr - TechMagister/liquid.cr#44


src/amber/cli/templates/app/.travis.yml
```diff
@@ -1 +1,6 @@
 language: crystal
+
+script:
+  - bin/ameba
```
Sija, line 4: IMHO they should be put in reverse order.

drujensen: Do you mean run `crystal spec` first, then `crystal tool format --check`, then `bin/ameba`?

Sija: Yes, that was my point. Because I reckon that the spec failures are more important then stylistic wrongdoings and you'd like to know about 'em as soon as possible.

drujensen: Interesting perspective.  Didn't think much about it to be honest.  Anyone else have opinions on this?  If not, I will reverse the order.

waghanza: Agree with @Sija, having `spec failures` is more important than coding conventions. Having any build `CI` notification before `ameba` run is better having them after (in term of response in case of bug)

PR #1217, opened 2020-06-26
Uses docker in travis build

coreycarvalho:
### Description of the Change

This change runs CI in Travis using docker. Unfortunately, there is a limitation in Travis that does not allow the pinning of a crystal version (it defaults to latest). This causes a mismatch in the version of crystal being run in CI vs the version of crystal that amber is expecting.

### Alternate Designs

An alternate design was to use docker-compose instead of building the image and using a bunch of `docker run` commands. I opted not to use docker-compose, as some of the configurations in `spec/support/config` seemed like they were hardcoding the redis_url, and passing `REDIS_URL` within the docker-compose file did not seem to overwrite it. This caused tests to not be able to connect to Redis.

### Benefits

Running the same version of crystal in CI that amber expects helps build more confidence in the test suite.

### Possible Drawbacks

A slight increase in build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 dist: xenial
 sudo: required
-language: crystal
 
 services:
   - postgresql
   - redis-server
+  - docker
 
 cache:
   directories:
@@ -20,9 +20,10 @@ before-install:
   - psql -U postgres -c "ALTER USER postgres WITH PASSWORD 'password';"
 
 script:
-  - bin/ameba
-  - crystal tool format --check
-  - crystal spec $TEST_SUITE
+  - docker build -t amber-test .
+  - docker run --rm amber-test /opt/amber/bin/ameba
+  - docker run --rm amber-test crystal tool format --check
+  - docker run --rm --network=host amber-test crystal spec $TEST_SUITE
 
 notifications:
   webhooks:


.travis.yml
```diff
@@ -20,9 +20,10 @@ before-install:
   - psql -U postgres -c "ALTER USER postgres WITH PASSWORD 'password';"
 
 script:
-  - bin/ameba
-  - crystal tool format --check
-  - crystal spec $TEST_SUITE
+  - docker build -t amber-test .
```
drujensen, line 23: consider using:
```
docker-compose run spec
```

coreycarvalho: Ah, good callout.

coreycarvalho: I attempted to use docker-compose here, but received the following errors when running tests:

```bash
Failures:

  1) database sqlite test creates the database when db migrate
     Failure/Error: File.exists?(db_filename).should be_true

       Expected: true
            got: false

     # spec/amber/cli/commands/database_spec.cr:35

  2) Amber::Pipe::Session session persist across different requests Redis Store sets session value in controller

       Socket::ConnectError: Error connecting to 'localhost:6379': Cannot assign requested address (Redis::CannotConnectError)
         from lib/redis/src/redis/socket_wrapper.cr:12:5 in 'initialize'
         from lib/redis/src/redis/connection.cr:8:3 in 'new'
         from lib/redis/src/redis.cr:157:19 in 'connect'
         from lib/redis/src/redis.cr:117:5 in 'initialize:url'
         from lib/redis/src/redis.cr:95:3 in 'new:url'
         from src/amber/router/session/session_store.cr:19:7 in 'redis_store'
         from src/amber/router/session/session_store.cr:10:31 in 'build'
         from src/amber/router/context.cr:30:18 in 'session'
         from spec/amber/pipes/session_spec.cr:41:13 in '->'
         from ../../usr/share/crystal/src/primitives.cr:255:3 in 'internal_run'
         from ../../usr/share/crystal/src/spec/example.cr:33:16 in 'run'
         from ../../usr/share/crystal/src/spec/context.cr:18:23 in 'internal_run'
         from ../../usr/share/crystal/src/spec/context.cr:330:7 in 'run'
         from ../../usr/share/crystal/src/spec/context.cr:18:23 in 'internal_run'
         from ../../usr/share/crystal/src/spec/context.cr:330:7 in 'run'
         from ../../usr/share/crystal/src/spec/context.cr:18:23 in 'internal_run'
         from ../../usr/share/crystal/src/spec/context.cr:330:7 in 'run'
         from ../../usr/share/crystal/src/spec/context.cr:18:23 in 'internal_run'
         from ../../usr/share/crystal/src/spec/context.cr:147:7 in 'run'
         from ../../usr/share/crystal/src/spec/dsl.cr:270:7 in '->'
         from ../../usr/share/crystal/src/primitives.cr:255:3 in 'run'
         from ../../usr/share/crystal/src/crystal/main.cr:45:14 in 'main'
         from ../../usr/share/crystal/src/crystal/main.cr:114:3 in 'main'
         from __libc_start_main
         from _start
         from ???
```

It looks like I would have had to modify tests and come configs. I will look into docker-compose in another iteration, I was mainly concerned with having the proper version of crystal run in CI at the moment.

drujensen: Ok, this is the `REDIS_URL` and `DATABASE_URL`.  I think we can support environment variables here.

drujensen:
@coreycarvalho thanks for fixing this!  Now we just need to solve why crystal 0.35.1 broke it.

PR #1222, opened 2020-07-15
Support granite migrations

drujensen:
### Description of the Change

This change adds support for the new granite migrations DSL that is similar to Rails.  This replaces the template generators to use the new syntax.

### Alternate Designs

Keep the existing SQL generators

### Benefits

One of the most requested features is support for a clean DSL for migrations.  This will make Rails developers feel right at home using Amber.

### Possible Drawbacks

The current implementation does not have all the field options yet.  A work around is to use `t.custom` and use SQL syntax for the field for now.


eliasjpr:
This is a great addition! Thank you @drujensen

Issue #1224, opened 2020-08-03
If the origin header doesn’t exist, the CORS pipe should ignore it and not validate against the allowed list.


samholst:
### Description

When making a GET request from the frontend, the Amber CORS pipe will block the request.

```crystal
  pipeline :api do
    plug Amber::Pipe::PoweredByAmber.new
    plug Amber::Pipe::Error.new
    plug Amber::Pipe::Logger.new
    plug Amber::Pipe::Session.new
    plug Amber::Pipe::CORS.new(origins: ["*", %r(), "http://localhost:3000", "http://localhost:8080"])
  end
```

### Steps to Reproduce

1. Open Firefox browser (version 79.0 64-bit)
2. Load http://localhost:3000
3. Make a GET request from JavaScript frontend
4. Will be blocked due to CORS

**Expected behavior:** [What you expect to happen]

If the origin header doesn’t exist, the CORS pipe should ignore it and not validate against the allowed list.

**Actual behavior:** [What actually happens]

Is validated and request failed using for example a GET request coming from the frontend to the API pipeline.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions

Version 0.34.0 of Amber and Crystal

### Additional Information

As something else (I could be implementing it wrong), using http://localhost:8080 (npm server) and requesting a GET endpoint in the API pipeline on http://localhost:3000 (amber server) produces a `missing allow origin` error on the frontend. In the controller's action, I have

```crystal
response.headers.add("Access-Control-Allow-Origin", "*")
```

which I believe should allow the request, but doesn't seem to?

eliasjpr:
Fixed with  #1225

PR #1225, opened 2020-08-05
[CORS]  Process Requests As Normal for Non CORS

eliasjpr:
**Issue:** https://github.com/amberframework/amber/issues/1224

### Description of the Change

When using CORS pipe all requests are being validated as CORS.

If validation fails 403 forbidden is thrown

### Changes

- If origin header is not present process the request as a normal one

### Side Effects

- Non CORS  requests are processed requests as normal when CORS handler is present

### DIFF
--- a/spec/amber/pipes/cors_spec.cr
+++ b/spec/amber/pipes/cors_spec.cr
@@ -8,10 +8,10 @@ module Amber::Pipe
       assert_cors_success(context)
     end
 
-    it "does not return CORS headers if Origin header not present" do
+    it "process requests as normal for non CORS requests" do
       context = cors_context("GET")
       cors_add_next(CORS.new).call(context)
-      assert_cors_failure context
+      assert_non_cors_request context
     end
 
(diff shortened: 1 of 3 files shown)


spec/amber/pipes/cors_spec.cr
```diff
@@ -8,10 +8,12 @@ module Amber::Pipe
       assert_cors_success(context)
     end
 
-    it "does not return CORS headers if Origin header not present" do
+    it "process requests as normal for non CORS requests" do
       context = cors_context("GET")
+      puts context.request.headers
```
drujensen, line 13: ✂️

src/amber/pipes/cors.cr
```diff
@@ -134,7 +136,7 @@ module Amber
         @origins.includes? "*"
       end
 
-      private def origin_header?(request)
```
Blacksmoke16, line 137: Might be able to make this `protected` if you wanted to keep it more internal.

PR #1231, opened 2020-09-08
Explicitly declare the UTF-8 charset for json responses

damianham:
### Description of the Change

I was getting incorrect decoding of UTF-8 characters in a Flutter application pulling data from an Amber web application that contained text strings with accented characters and this Dart issue explains why https://github.com/dart-lang/http/issues/175

### Benefits

Dart HTTP clients will be able to correctly decode json data from an Amber app

### Possible Drawbacks
None - it conforms to the json spec


eliasjpr (dismissed):

eliasjpr:
@damianham after reading https://www.w3.org/International/questions/qa-choosing-encodings and this https://www.w3.org/International/questions/qa-choosing-encodings#http I think it is best to have all responses to default to UTF-8. I wonder if there would be any implications for currently running Amber Apps. Thoughts

Issue #1221, opened 2020-07-05
Client reload doesn't work out-the-box in Chomium


faustinoaq:
### Description

Chromium Browser doesn't update the code changes in `main.js` or `main.scss`

### Steps to Reproduce

1. `amber watch`
2. Open http://0.0.0.0:3000/ in Chromium Browser
2. Change `$primary` in `main.scss`
3. See triggered reload in Chromium Browser

**Expected behavior:** color changes

**Actual behavior:** colors doesn't change

**Reproduces how often:** 100%

### Versions

```
$ shards --version
Shards 0.10.0 (2020-05-03)
$ crystal -v
Crystal 0.35.0 (2020-06-09)

LLVM: 10.0.0
Default target: x86_64-pc-linux-gnu
$ amber --version
Amber CLI (amberframework.org) - v0.35.0
$ npm -v
6.14.5
$ node -v
v14.3.0
$ lsb_release -a
LSB Version:    n/a
Distributor ID: ManjaroLinux
Description:    Manjaro Linux
Release:        20.0.3
Codename:       Lysia
```

### Additional Information

The issue goes away when I check the option `Disable cache (when DevTools is open)`

![image](https://user-images.githubusercontent.com/3067335/86541072-8cebe900-becf-11ea-812a-dfd7b5ef0b87.png)

I opened this issue to confirm Is this the right workaround or should I change some cache settings in my app?

Happy Crystalling! :heart: :sunglasses:

eliasjpr:
@faustinoaq do you think you can take a look at this?

midhun-at-vineti:
Looks like this has to do with how Chromium and Chrome behave on 'Cache-Control' header, which is set by [pipes/static.cr](https://github.com/amberframework/amber/blob/bb1db0b9536249c7c8e5a6ba9804c3742bd02d9b/src/amber/pipes/static.cr#L134) to private, max-age=3600. 

I tried reproducing this in Chrome but Chrome for some reason doesn't cache the main.bundle.css file but Chromium caches it and is reproducible. 

@eliasjpr @faustinoaq Do you think it would be a good idea to make the static response headers configurable? Would be happy to create a PR.

eliasjpr:
@midhun-at-vineti Yes lets make those headers configurable if possible have some default for Chromium

midhun-at-vineti:
@eliasjpr on it. 👍

PR #1233, opened 2020-09-30
1221 configurable static response headers

midhunkrishna:
Ability to configure response headers from <environment>.yml file
with the following configuration,

```yaml
public_file_server:
  headers:
    "Cache-Control": "private, max-age=7200"
```
### Description of the Change

When files are served by the static file server, via [pipes/static.cr](https://github.com/amberframework/amber/blob/master/src/amber/pipes/static.cr), and does not expose a way to modify the [hard coded Cache-Control headers](https://github.com/amberframework/amber/blob/master/src/amber/pipes/static.cr#L134)

### Alternate Designs

This enhancement is similar to the static file middle-ware headers configuration of Rails. [Original PR](https://github.com/rails/rails/pull/19135/files)

### Benefits

Ability to modify response headers for files served by static pipe. 

### Possible Drawbacks

Since this PR is an enhancement to the existing functionality, no changes are introduced to the behavior of the framework.

### DIFF
--- a/spec/support/config/with_public_file_server.yml
+++ b/spec/support/config/with_public_file_server.yml
@@ -1,0 +1,17 @@
+port: 3000
+port_reuse: true
+process_count: 1
+redis_url: "redis://localhost:6379"
+database_url: "mysql://root@localhost:3306/development_settings_test"
+
+session:
+  key: "myapp.session"
+  store: "encrypted_cookie"
+  expires: 10
+
+secrets:
+  description: "Store your test secrets credentials and settings here."
+
+public_file_server:
+  headers:
+    "Cache-Control": "private, max-age=7200"
--- a/spec/amber/environment/settings_spec.cr
+++ b/spec/amber/environment/settings_spec.cr
@@ -35,7 +35,7 @@ module Amber::Environment
 
     describe "#static_file_server" do
       it "sets default headers value as empty map" do
-        test_yaml = File.read(File.expand_path("./spec/support/config/test.yml"))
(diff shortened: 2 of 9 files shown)


midhunkrishna:
@eliasjpr This should fix #1221 . Can you take a look and add your thoughts?

spec/support/config/with_public_file_server.yml
```diff
@@ -0,0 +1,17 @@
...
+
+session:
+  key: "myapp.session"
+  store: "encrypted_cookie"
+  expires: 10
+
+secrets:
+  description: "Store your test secrets credentials and settings here."
+
+public_file_server:
```
drujensen, line 15: This name `public_file_server` doesn't relate to the `static` pipe which is confusing to me.  Also, It may not be publicly facing.  Maybe rename it to `static_file_server`
...or...
What if we create a section for `pipes` settings?
```
pipes:
  static:
    headers:
      "Cache-Control": "private, max-age=7200"
```
We can move other pipe specific settings here.



Also move this section above `secrets` since it should be last.

drujensen (dismissed):
💯  Looks good.  Thanks for the contribution.  I have a small nit-picky naming concern.

spec/support/config/with_public_file_server.yml
midhunkrishna: @drujensen Makes sense. I like the idea of per-pipe configuration. Let me update this.

drujensen (approved):

eliasjpr (approved):

PR #1234, opened 2020-10-18
Credentials

andrewc910:
### Description of the Change

Changes .encryption_key to "./config/master.key".


drujensen:
@andrewc910 This looks like an ambitious project.  Would love it if instead of copying the project, you would consider contributing to it instead.  

You might have the NIH syndrome. I'm not sure what the cure is for that though.  ;-)

andrewc910:
@drujensen I plan to contribute any way I can. PR's will be opened for any changes that can affect both repos. #1230 was one such PR. What I am changing without opening PR's are changes I believe the core team would have no interest in, for example changing the default ORM to Jennifer.

PR #1235, opened 2020-10-19
Plugin: Add support for a config file

andrewc910:
### Description of the Change

This adds the ability for plugin authors to create a `config.yml` file. The file is should be located in the plugin folder. Currently, I have only added the ability to inject routes into the users `route.cr` file but I have no problem adding extra functionality if people bring up ideas.

The code expects the YAML file to be in the following format:

```yml
routes:
  pipelines:
    web:
      - get "/", HomeController, :index
      - post "/", HomeController, :index
```

### Benefits

Right now, the benefit is routes can be auto-generated. 

### Possible Drawbacks

More code. 

A current limitation is it breaks if you don't have the full information. The cli will not read the file if the file doesn't exist. It doesn't break if the file is empty. It does break when the file is just:

```yml
routes:
  pipelines:
```

or

```yml
routes:
  pipelines:
    web:
```

But plugin authors shouldn't be doing that anyways.

### DIFF
--- a/spec/amber/cli/plugins/installer_spec.cr
+++ b/spec/amber/cli/plugins/installer_spec.cr
@@ -1,0 +1,24 @@
+require "../../../spec_helper"
+require "../../../support/helpers/cli_helper"
+
+include CLIHelper
+
+module Amber::Plugins
+  describe Installer do
+    Spec.after_each do
+      cleanup
+    end
+
+    it "should add routes when installing plugin" do
+      scaffold_app(TESTING_APP)
+      Dir.mkdir_p("#{Dir.current}/lib/test/plugin")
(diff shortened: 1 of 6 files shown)


drujensen (changes requested):
Can you add test coverage?

andrewc910:
@drujensen A spec has been added. 

I had to read the routes file my self. I tried doing what routes_spec does with:

```crystal
output = ""
Amber::CLI::MainCommand.run %w(routes) { |cmd| output = cmd.out.gets_to_end }
output_lines = route_table_rows(output)
```

But it would hang indefinitely for me.

drujensen (approved):

PR #1240, opened 2020-11-17
Unify AbstractStore#update API

### LINKED ISSUE #1239: Inconsistent Amber::Router::Session::AbstractStore API
The current state is the following:

```cr
abstract class AbstractStore
  abstract def update(other_hash)
end

class CookieStore < AbstractStore
  def update(hash : Hash(String | Symbol, String))
  end
end

class RedisStore < AbstractStore
  def update(hash : Hash(String, String))
  end
end
```

With 1.0 that will no longer compile due to https://github.com/crystal-lang/crystal/pull/9634.

What is the expected API? `Hash(String | Symbol, String)` or `Hash(String, String)` or any `include Enumerable({K, String})`?
Whatever the choice it would be need to be reflected in the `AbstractStore`, or remove the type restrictions in the implementation classes letting the type error show in the call-site.

bcardiff:
Fix #1239

`Redis#hmset` already uses `key.to_s` to convert keys to strings.

### DIFF
--- a/src/amber/router/session/abstract_store.cr
+++ b/src/amber/router/session/abstract_store.cr
@@ -3,7 +3,7 @@ module Amber::Router::Session
   abstract class AbstractStore
     abstract def id
     abstract def destroy
-    abstract def update(other_hash)
+    abstract def update(other_hash : Hash(String | Symbol, String))
     abstract def set_session
     abstract def current_session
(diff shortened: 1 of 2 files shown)


drujensen (approved):
thanks @bcardiff

PR #1268, opened 2021-08-10
Add México to the PHONE_FORMAT locale list

renich:
### Description of the Change

It's a simple check but enough for now until we figure out the exact regex. It will be complicated, though, since I couldn't find a full list online and asking Telmex is going to be a pain...

### Alternate Designs

If Telmex provides a full list, we can devise the exact RegEx required for a true validation. 

### Benefits

We will have a mexican phone locale available. 

### Possible Drawbacks

The check will allow any 10 digit  number with or without +52 (the country code). You can, still, enter bogus phones and they will pass unnoticed.


crimson-knight:
Wouldn't Mexico be covered by the North American phone number format? The US, Canada and Mexico all use the same formatting for phone numbers.

crimson-knight:
Nevermind, I see that isn't in here now that I look more closely.

renich:
Yeah, it wouldn't. I'm, actually, having issues with this. This is why I've decided to go for a simplified RegEx for now.

drujensen (approved):

PR #1279, opened 2022-05-17
Changed discord as the primary way to get in touch with the comminuty

crimson-knight:
Updated the README link to refer new users to the discord server


robacarp (approved):
Great start

PR #1289, opened 2022-07-12
Feature/change default orm to jennifer

crimson-knight:
### Description of the Change

This PR changes the default ORM to Jennifer, and integrates the CLI commands for Jennifer into the `amber` CLI tool.

A couple of nice changes:
- Detects the version of node and updates the node-sass dependency versions so that styling properly compiles if you're using Node 14+
- Changed the 3rd link in the auto generated Amber app to link to the discord
- Adds the `Sam` CLI utility to make running tasks easy.

Potential breaking change:
- The I18n shard that Jennifer relies on removes the `force_locale: context.locale` feature. I think this means we no longer have "dynamic" language detection. I don't see any tests currently for this, but it's worth trying to add this back in and I think using URL based language detection. The default language can be configured easily, and all the translation capabilities work it is just not dynamic.

### Benefits

Not having to solo-maintain an ORM and using an ORM that has more of the db-specific data types working (mostly looking at Postgres)

### Possible Drawbacks

Currently, this tightly couples the Amber CLI tool with the Sam CLI tool provided by Jennifer. The Sam tool is pretty handy for rake/make-like jobs, so I'm pretty happy with doing this. I think there will be a lot of situations where other devs will want to run tasks (data migrations/alterations, etc). so this will be a handy improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ brew install amber
 git clone https://github.com/amberframework/amber.git
 cd amber
 git checkout stable
+shards install
 make
 sudo make install
 ```


drujensen:
Hi Seth.  Thanks for submitting this PR.  This is a very large PR and will be difficult to review.  I will give it a spin when I get a chance.  Will you be able to update the documentation as well?

drujensen (changes requested):
I tried using `amber g auth user` and I'm getting an error: `class User < Granite::Base`.  You will need to update all of the scaffolding to use Jennifer.

drujensen:
Also, looks like the `Dockerfile` is still using crystal 1.0.0 and doesn't build.

crimson-knight:
Thanks @drujensen for the review, you're right I haven't update the auth generator yet. I had forgotten about the docker file, but I will need to update that in a different PR since that's a different scope from this one.

I'll definitely get the documentation updated once this PR is done and I know the full extent of the changes.

crimson-knight:
Oh yeah, FWIW a lot of the file changes are just tweaks suggested by Ameba so I could get the `bin/amber_spec` to pass.

src/amber/cli/commands/new.cr
```diff
@@ -38,17 +39,19 @@ module Amber::CLI
...
           generator = Generators.new(name, full_path_name)
         end
+
         generator.generate_app(options)
 
         # Encrypts production.yml by default.
-        cwd = Dir.current; Dir.cd(full_path_name)
-        MainCommand.run ["encrypt", "production", "--noedit"]
-        Dir.cd(cwd)
+        #cwd = Dir.current; Dir.cd(full_path_name)
```
eliasjpr, line 52: Is this intentional? Maybe you want to uncomment these lines

src/amber/cli/commands/pipelines.cr
```diff
@@ -113,7 +113,7 @@ module Amber::CLI
         table.border_color = :dark_gray unless options.no_color?
 
         if options.no_plugs?
-          result.map { |pipes_and_plugs| pipes_and_plugs[:pipes] }.flatten.uniq.each do |pipe|
+          result.flat_map { |pipes_and_plugs| pipes_and_plugs[:pipes] }.uniq!.each do |pipe|
```
eliasjpr, line 116: 👍

src/amber/cli/helpers/process_runner.cr
```diff
@@ -79,7 +79,7 @@ module Sentry
     private def check_processes
       @processes.each do |task, procs|
         # clean up process list and restart if terminated
-        if procs.any?
+        if !procs.empty?
```
eliasjpr, line 82: Maybe: `unless procs.empty?`

crimson-knight:
Closing this out in favor of a different approach that allows for any ORM to be used and connected to the generators/CLI tool

PR #1292, opened 2022-08-23
Quality of life improvements

crimson-knight:
### Description of the Change

- Added github actions for CI instead of TravisCi
- Updated ameba and improved the codebase to pass all ameba checks
- Fixed a few failing tests for cookie functionality that relied on outdated crystal behavior
- Changed the `amber new` command to check which node version a user has installed and adjust the node-sass sources based on the node version found
- Bumped version from 1.2.1 to 1.3.0 because this is intended to be released so I can start creating content and publishing docs for the Amber community

### Benefits

New Amber adopters will get a smoother experience when getting started using `amber new`

### Possible Drawbacks

May impact pre 1.0.0 crystal users

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,17 @@ jobs:
           --health-retries 5
         ports:
           - 6379:6379
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
 
     steps:
       - name: Update & install sqlite3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,8 @@ jobs:
       image: crystallang/crystal
 
     steps:
-      - name: Remove sqllite3 dev library
-        run: sudo apt uninstall libsqlite3-dev
+      - name: Install sqllite3 dev library
+        run: apt install libsqlite3-dev
       - name: Download source
         uses: actions/checkout@v3
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,12 +2,14 @@ on:
   push:
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     container:
       image: crystallang/crystal
 
     steps:
+      - name: Remove sqllite3 dev library
+        run: sudo apt uninstall libsqlite3-dev
       - name: Download source
         uses: actions/checkout@v3
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,6 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-        ports:
-          - 6379:6379
 
     steps:
       - name: Update & install sqlite3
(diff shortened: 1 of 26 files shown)


src/amber/pipes/cors.cr
```diff
@@ -132,8 +132,8 @@ module Amber
         end
       end
 
-      def any?
-        @origins.includes? "*"
+      def empty?
```
drujensen, line 135: I think `any` is a better name here.  The `*` symbol represents "any origin" as in the wildcard symbol.  Empty is a common method on a string and may confuse someone into thinking it is checking for an empty string.

src/amber/router/cookies/store.cr
```diff
@@ -90,6 +90,9 @@ module Amber::Router::Cookies
             expires : Time? = nil, domain : String? = nil,
             secure : Bool = false, http_only : Bool = false,
             extension : String? = nil)
+      name = URI.encode_www_form(name)
```
drujensen, line 93: Is this a breaking change?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
+on:
```
drujensen, line 1: This doesn't seem to be running all the different combinations of tests.  I thought we ran against postgres, mysql and sqlite.  Also, I thought we tested different templating options using docker compose.

drujensen (approved):

src/amber/pipes/cors.cr
crimson-knight: That's a good point, I hadn't given it enough thought since I was just going through and fixing what Ameba said was wrong. I'll probably change from `#any?` to `#any_origin?` to stop Ameba from complaining and make it clearer what we are checking for with this method.

src/amber/router/cookies/store.cr
crimson-knight: No this is preserving the existing expected behavior. I had to check the std lib, and 2 years ago the behavior for setting cookie names changed from implicitly encoding names and values to not encoding them at all. So I added this to preserve the expected behavior.

Now that I've said that I need to check and see if this means Amber will only work as expected on Crystal 1.0+

crimson-knight: Yeah looks like this behavior changed as of 1.0 https://github.com/crystal-lang/crystal/pull/10485

.github/workflows/ci.yml
crimson-knight: I'll look into expanding the combination of tests more. I'm not very familiar with GHA and I'm just trying to make it work to start.

crimson-knight: And FYI I still have 10 failing tests to fix, so I won't merge this until at least everything is green on my repo

drujensen:
@crimson-knight Anyway to test this locally first?  I'm getting a lot of emails asking for reviews.

crimson-knight:
@drujensen sorry about that! As far as the GHA go, I haven't found a way to test that workflow locally. I did take a look at the test suite and it appears within that is where we are testing the different db adapters (mostly SQLite and PG). It looks like the tests that are failing on GHA are possibly related to differences in how CI runs the tests so I may have to configure this to use a docker image to run the actual tests to simulate a local env.

My local tests all pass, 480/480 now so I can merge this any time and stop email bombing you for a re-review :)

crimson-knight:
@drujensen I have this at a satisfactory level for now. Github actions probably need a little more tweaking, but locally I get 480/480 tests passing, and some of these tweaks will help users on newer systems to use Amber smoothly.

I'd like to get this reviewed and cut a release today if possible.

crimson-knight:
@robacarp can I get your review when you get a few minutes?

robacarp (approved):

PR #1293, opened 2022-08-27
Remove crystalshards.xyz link on template view

bananatron:
### Description of the Change

Removes the crystalshards.xyz link. Perhaps it should be https://crystalshards.org (idk where the best spot is rn, I'm super ignorant) but the domain is compromised for the other link and it seems bad.

### Alternate Designs

N/A

### Benefits

The template page looks more professional.

### Possible Drawbacks

N/A


crimson-knight (approved):
LGTM. Thank you for contributing!

PR #1296, opened 2022-09-22
fix to cli scaffold controller spec deprecated garnetspec module

rrevi:
### Description of the Change

At some point in the history of the GarnetSpec shard, the `GarnetSpec::Controller::Test` module was deprecated, instead the `GarnetSpec::SystemTest` module is to be used.

### Benefits

Fixes an error when running cli scaffold controller spec files in more current versions of the GarnetSpec shard.

### Possible Drawbacks

Will not work in older versions of GarnetSpec shard.


crimson-knight (approved):
Nice catch, looks like there are also 2 other places this needs to be updated. I can open a separate PR for that.

PR #1304, opened 2022-11-11
feat!: remove webpack from dependencies

drujensen:
### Description of the Change

This change removes webpack from the dependencies. It also upgrades bootstrap to 5.0 using cdn links in the main template.

### Alternate Designs

Keep it as is

### Benefits

Removing webpack, nodejs, and npm dependencies simplifies amber templates and decouples from the developers choice on frontend solutions.

### Possible Drawbacks

This is a BREAKING CHANGE and requires a MAJOR version upgrade.


crimson-knight (commented):
I like it, thanks for doing this!

Do you think it's really advantageous to change the default db to sqlite from PG?

I don't think this is a breaking change, it only effects newly generated apps. Existing apps won't see a difference. I have this change listed under the roadmap for v1.5

drujensen:
You might be right about the breaking change.  It still a significant change that we should look through the documentation and see if we need any updates.  I will work on fixing the tests and adding htmx for delete functionality next.

drujensen:
@crimson-knight are the tests passing in master?  I'm getting all kinds of errors:
```
crystal spec spec/amber/pipes/powered_by_amber_spec.cr:19 # Amber::Pipe::PoweredByAmber Adds X-Powered-By: Amber to response should contain X-Powered-By in response
crystal spec spec/amber/pipes/pipeline_spec.cr:34 # Amber::Pipe::Pipeline with given server config routes
crystal spec spec/amber/pipes/pipeline_spec.cr:41 # Amber::Pipe::Pipeline with given server config responds to GET request
crystal spec spec/amber/pipes/pipeline_spec.cr:48 # Amber::Pipe::Pipeline with given server config responds to PUT request
crystal spec spec/amber/pipes/pipeline_spec.cr:55 # Amber::Pipe::Pipeline with given server config responds to PATCH request
crystal spec spec/amber/pipes/pipeline_spec.cr:62 # Amber::Pipe::Pipeline with given server config responds to POST request
crystal spec spec/amber/pipes/pipeline_spec.cr:69 # Amber::Pipe::Pipeline with given server config responds to DELETE request
crystal spec spec/amber/pipes/pipeline_spec.cr:76 # Amber::Pipe::Pipeline with given server config responds to HEAD request
crystal spec spec/amber/pipes/client_ip_spec.cr:11 # Amber::Pipe::ClientIp IP from headers gets first client IP from default header
crystal spec spec/amber/pipes/client_ip_spec.cr:28 # Amber::Pipe::ClientIp IP from headers gets client IP from first custom header found
crystal spec spec/amber/cli/commands/init_spec.cr:94 # Amber::CLI::MainCommand::New Database settings generates pg correctly
crystal spec spec/amber/cli/commands/exec_spec.cr:26 # amber exec within project executes multi-lines from the command-line argument
crystal spec spec/amber/cli/commands/exec_spec.cr:39 # amber exec within project executes a .cr file from the first command-line argument
crystal spec spec/amber/cli/commands/exec_spec.cr:47 # amber exec within project opens editor and executes .cr file on close
crystal spec spec/amber/cli/commands/exec_spec.cr:53 # amber exec within project copies previous run into new file for editing and runs it returning results
crystal spec spec/amber/cli/commands/exec_spec.cr:62 # amber exec outside of project executes outside of project but without including project
crystal spec spec/amber/cli/commands/database_spec.cr:57 # database postgres test has test  connection settings
```

For example the PoweredByAmber Pipe is not passing and it looks like its not configured.  Let me know if I need to fix these first.

crimson-knight:
@drujensen yes I have 100% passing tests on master on my local when I run `bin/amber_spec`. The GithubActions fail for the redis related tests because I couldn't get the redis configuration to work.

drujensen:
which version of crystal are you running?

crimson-knight:
The latest 1.6.2. All tests passed originally when I updated everything on 1.5 though

drujensen:
ok, maybe its crystal 1.6.2?  I can't get them to pass locally.  I will track them down.  Thanks.

crimson-knight:
I think a large part of it is around the DB being changed from PG to sqlite. There are a lot of tests around PG being the default DB and relying on that being present.

drujensen:
not related. I'm trying to get master to pass locally. no worries.  I will figure it out.

crimson-knight:
No worries. I'll be putting around tomorrow working on my bulldozer if you need anything, just message me in Discord

drujensen:
Looks like the redis tests are failing in github actions. I've open an issue #1305 to address that.  Please review and merge.

bin/amber_spec
```diff
@@ -3,8 +3,11 @@
 echo "Running './bin/ameba':"
 ./bin/ameba
 
-echo "\nRunning 'crystal tool --format --check':"
+echo "\nRunning 'crystal tool format --check':"
 crystal tool format --check
 
 echo "\nRunning 'crystal spec':"
 crystal spec
+
```
drujensen, line 11: We still need to fix these.  This is throwing an error in selenium driver being used by the template.

spec/amber/controller/render_spec.cr
```diff
@@ -29,11 +29,11 @@ module Amber::Controller
         RenderController.new(context).render_with_layout.should eq layout_with_template
       end
 
-      it "renders a form with a csrf tag" do
-        reuslt = RenderController.new(context).render_with_csrf
-        reuslt.should contain "<form"
-        reuslt.should contain "<input type=\"hidden\" name=\"_csrf\" value="
-      end
+      # it "renders a form with a csrf tag" do
```
drujensen, line 32: these run locally both directly using crystal spec and in docker using docker compose.  Also tested on an ubuntu host machine.  They still fail in github actions with same configuration.  It's related to redis not keeping the session key/value pair when requested.  Commenting out for now but we will need to revisit this.

marccremer:
You might want to move everything in [assets](https://github.com/amberframework/amber/tree/master/src/amber/cli/templates/app/src/assets) into [public](https://github.com/amberframework/amber/tree/master/src/amber/cli/templates/app/public)
at the moment we do not have the amber logo when we start a new project

marccremer:
This is what you see when you start a new project
2022-11-14-162635_1055x482_scrot

drujensen:
@marccremer should be fixed:
Screenshot 2022-11-14 at 7 47 43 AM
Screenshot 2022-11-14 at 7 48 10 AM

marccremer:
Did you push your changes?
These commits do not touch the relevant files

drujensen:
strange. checking `.gitignore`.  maybe i need to explicitly add them.

marccremer:
lgtm

crimson-knight:
@drujensen FWIW I had added the CLI binary to the .gitignore because it's compiled as arch specific, and not everyone is on the same architecture. I've had a fair amount of people on linux/intel mac and myself on M1 mac trying to run the cli tool in dev, so my thought was to ignore the CLI binary and always recompile after pulling changes to ensure you have the correct binary.

May not matter since you'd get an error trying to run a binary for an arch you don't have and you'd know to re-compile anyway

drujensen:
@crimson-knight makes sense.  The `cli` entry also excluded the whole cli directory so everything that was being added under that directory was being ignored.  probably just need to specify the binary file instead.

spec/support/helpers/cli_helper.cr
```diff
@@ -40,7 +40,7 @@ module CLIHelper
   def expected_db_url(db_key, env)
     case db_key
     when "pg"
-      "postgres://#{`whoami`.strip("\n")}:@localhost:5432/#{TEST_APP_NAME}_#{env}"
```
crimson-knight, line 43: I just added the `whoami` interpolation last week because it makes generating a new app and getting started pretty fast. Does this effect the tests on docker and that's why you added the `postgres:password` back? I had wondered, but I don't use docker and I forgot to check (I'm not a docker wizard).

src/amber/cli/templates/app/config/environments/development.yml.ecr
```diff
@@ -19,7 +19,7 @@ redis_url: "redis://localhost:6379"
 when "mysql" -%>
 database_url: mysql://root@localhost:3306/<%= database_name %>_development
 <% when "pg" -%>
-database_url: postgres://<%= `whoami`.strip("\n") -%>:@localhost:5432/<%= database_name %>_development
```
crimson-knight, line 22: This is the `whoami` I mentioned in another that I added to make getting started easier. I'd prefer to leave this in, unless we absolutely need it for docker to work. And even then I'd like to open another issue so we can keep the convenience of the `whoami` working for those on native installs and not docker.

crimson-knight (dismissed):

crimson-knight:
> 

Woops! :) I thought I had specified only the binary and ensure it's compiling into the `bin` directory

src/amber/cli/templates/app/config/environments/development.yml.ecr
drujensen: Understood. Whoami using the logged in user of your local system.  This is not necessarily the user setup for postgres.  I log into a separate box for development.  Let's see if we can come up with another solution.

crimson-knight: Let's just bring this in as-is for now and make the `whoami`/default db user stuff another issue

Dockerfile
```diff
@@ -2,9 +2,9 @@ FROM crystallang/crystal:latest
 
 # Install Dependencies
 ARG DEBIAN_FRONTEND=noninteractive
-RUN apt-get update -qq && apt-get install -y --no-install-recommends libpq-dev libsqlite3-dev libmysqlclient-dev libreadline-dev git curl vim netcat
+RUN apt-get update -qq && apt-get install -y libpq-dev libsqlite3-dev libmysqlclient-dev libreadline-dev curl vim
 
-# Install Node
+# Install nodejs
```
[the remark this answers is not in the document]
drujensen: something still requires node.  I haven't been able to track it down but you get this when running specs:
```
: 1: node: not found
.--: 1: node: not found
.--: 1: node: not found
.--: 1: node: not found
.--: 1: node: not found
.--: 1: node: not found
.--: 1: node: not found
.--: 1: node: not found
.--: 1: node: not found
.--: 1: node: not found
```
let me see if I can find it.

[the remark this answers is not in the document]
drujensen: ok, this took me down a rabbit hole of a bunch of stuff i didn't know existed.  what a web we weave.  I pulled a thread and its still coming out....

spec/amber/pipes/pipeline_spec.cr
```diff
@@ -16,15 +16,18 @@ module Amber
 
       describe "with given server config" do
         pipeline = Pipeline.new
+        before_each do
+          pipeline = Pipeline.new
```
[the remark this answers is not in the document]
drujensen: so, this is a bit strange and let me know if there is a better way to do this.  We need the `pipeline` to be in scope for the tests but we want to re-initialize it before_each test or they won't run independently.  I could have left the pipeline uninitialized but I think that is a bad practice.   If there is a better way of doing this, let me know.

spec/support/helpers/cli_helper.cr
drujensen: yeah, it breaks the docker but also my local setup.  :-)

crimson-knight (approved):

PR #1306, opened 2022-11-15
Remove garnet spec

drujensen:
### Description of the Change

This removes the Garnet Spec dependency that relied on selenium driver.  This adds a RequestHelper set of macros that provide a harness to test controller tests.  The macros create the http request methods to simplify specs.

This also enables the granite test suite which is crucial for a full set of test coverage for Amber.

### Alternate Designs

Fix the garnet_spec and selenium driver.

### Benefits

Although system or integration tests are valuable, the unit level tests shouldn't be dependent on this library.  This separates the 2 concerns.  You can still add integration tests using garnet (once its fixed) but they should be a separate set of tests and not tightly coupling the unit tests to these set of dependencies.

### Possible Drawbacks

Moving the RequestHelper inside of a module instead of in its own shard may limit reuse of this functionality in other frameworks.

### DIFF
--- a/src/amber/cli/templates/auth/spec/controllers/spec_helper.cr.ecr
+++ b/src/amber/cli/templates/auth/spec/controllers/spec_helper.cr.ecr
@@ -1,1 +1,2 @@
 require "../spec_helper"
+require "../request_helper"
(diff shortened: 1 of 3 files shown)


robacarp (approved):
There is some drawback to community re-usability but I think I prefer a "wait until there's an obvious use case" for that kind of thing. The big crystal frameworks have a huge maintenance burden on them in releasing shards because of how much relentless tagging and pushing is required, and amber is no exception. Anything that lightens the maintenance burden is a win.

Issue #1227, opened 2020-08-06
render "index.slang" generates macro errors with LAYOUT  = false


jeffreydvp:
### Description

With the default setup of a brand new Amber and an example scaffold, it creates the application.cr to look like this

### Steps to Reproduce

```crystal
require "jasper_helpers"

class ApplicationController < Amber::Controller::Base
  include JasperHelpers
  LAYOUT = false
end
```

If you comment out the `LAYOUT = false` and create the `application.slang` file, everything works as normal

**Expected behavior:** It should render just fine

**Actual behavior:** All the pages being rendered generate some strange macro errors.

**Reproduces how often:** 100% of the time

### Versions

Amber CLI (amberframework.org) - v0.35.0
Crystal 0.35.1 [5999ae29b] (2020-06-19)
LLVM: 8.0.0
Default target: x86_64-unknown-linux-gnu

eliasjpr:
@jeffreydvp thank you for submitting this issue. If you have a fix please feel free to open a pull request. It looks like the source of the issue is in the generators.

Let me know if you would like to submit a fix.

jeffreydvp:
Well I just commented it out and created the application.slim I’m not sure why that didn’t exist.

andrewc910:
@jeffreydvp What's the error message? Is it:
```sh
Which expanded to:

 > 1 | Kilt.render("#{"src/views"}/"layouts/#{false}"")
                                    ^
Error: expecting token ')', not 'layouts'
2020-08-15T16:51:46.322333Z   INFO - watch.run: Compile time errors detected, exiting...
12:51:46 watch.run (Info) Compile time errors detected, exiting...
```

Reproduced By: Creating a new project, setting `LAYOUT` to `false` in `application_controller.cr` & building.

@eliasjpr I just tested. If my error is correct, It's failing in `render_template` at `Kilt.render` because of line 27 `render_template("layouts/#{{{layout.class_name == "StringLiteral" ? layout : LAYOUT}}}", {{path}})`. It's converting the bool `false` to `"false"` and then trying to find that layout. This might be something we want to fix because the only non-hacky solution is setting `layout: false` in every controller that invokes `render`.

crimson-knight:
Closing this because the fix as been merged. Thank you @drujensen!

PR #1307, opened 2022-11-18
1227 fix layout equals false

drujensen:
### Description of the Change

This PR fixes issue #1227 .  This adds a check to see if LAYOUT is false or "false" and disables the layout for all of the render calls in the controller.

### Alternate Designs

Passing `layout=false` to all render calls in the controller

### Benefits

Reduces the code needed to disable the layout.

### Possible Drawbacks

It's not clear that LAYOUT can be set to false.  Its a bit awkward to have a (String | Boolean) union type for this field.


crimson-knight (approved):
LGTM

PR #1308, opened 2023-01-03
Bump ameba version to ~> 1.3.0

Sija:
### Description of the Change

Bumps ameba dependency.

### Benefits

It allows newly created apps to be runnable out-of-the-box (ameba < 1.3 doesn't work with recent Crystal versions).

### Possible Drawbacks

None


crimson-knight (approved):
Nice catch and thanks for doing this!

PR #1319, opened 2023-05-06
Shard updates, mysql uuid type, and api route bug fix

kalinon:
### Description of the Change

1. update shard versions
2. add UUID as a field type for mysql
3. fixed printing of `api` routes

### DIFF
--- a/src/amber/cli/commands/routes.cr
+++ b/src/amber/cli/commands/routes.cr
@@ -75,6 +75,7 @@ module Amber::CLI
         if route_match = resource_string.to_s.match(RESOURCE_ROUTE_REGEX)
           filter = route_match[4]?
           filter_actions = route_match[5]?.to_s.gsub(/\:|\s/, "").split(",")
+
           ACTION_MAPPING.each do |verb, v|
             v.each do |action|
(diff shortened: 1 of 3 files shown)


src/amber/cli/commands/routes.cr
```diff
@@ -75,20 +75,21 @@ module Amber::CLI
...
               when "except"
                 next if filter_actions.includes?(action)
-              else
-                build_route(
-                  verb: verb, controller: route_match[3]?, action: action,
-                  pipeline: current_pipe, scope: current_scope,
-                  uri_pattern: build_uri_pattern(route_match[2]?, action, current_scope)
-                )
               end
+
```
crimson-knight, line 87: I'm trying to wrap my head around what's happening here and why it was in the `else` clause to begin with.

I'm going to roll with this for now, as long as CI passes.

crimson-knight (approved):

src/amber/cli/commands/routes.cr
kalinon: think it was a mistake. i dont see how it would have ever worked.

PR #1324, opened 2023-06-24
Fixed the STD Lib monkey patch

crimson-knight:
While trying to create a client to send `POST` requests with type `multipart/form-data`, my requests were being caught and messed up somehow.

This is infuriating to track down because the monkey patch was not at all obvious so I went ahead and removed this.

Overall the behavior is kind of weird. Changing form submissions from `POST` to other request types feels a bit dangerous.

There are a few tests that are failing for `io` reasons now. I'm running this branch in prod on my own project to see what kind of side effects it might have (yolo!)

Anyway, this was a huge blocker. Anyone using or trying to create an HTTP client of any kind would be affected.

### DIFF
--- a/spec/amber/router/request_spec.cr
+++ b/spec/amber/router/request_spec.cr
@@ -74,42 +74,5 @@ module Amber::Router
         end
       end
     end
-
-    describe "#method" do
-      %w(PUT PATCH DELETE).each do |method|
-        it "overrides form POST method to PUT, PATCH, DELETE" do
-          headers[HTTP::Request::OVERRIDE_HEADER] = method
-          request = HTTP::Request.new("POST", "/?test=test", headers)
-          request.method.should eq method
-        end
-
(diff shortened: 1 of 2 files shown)


crimson-knight:
The monkey patch in the STD lib is to handle HTTP Tunneling from HTML forms that are using the regular form submission behavior but are being used for `PATCH` requests (ie the 'edit' view).

But because `HTTP::Request` from the STD lib is used for both incoming requests and outgoing requests, anyone who uses an `HTTP::Client` to try and send certain types of requests would encounter interference in how the requests are formed (which is how I found this).

Standard HTML forms only submit with `GET` and `POST`. So this means `DELETE` and `PATCH` requests will need some form of Javascript to be properly formed and sent.

I'm still working through a way to solve this. Obviously there are existing JS libraries out there and writing your own `fetch` is easy, which is how I'm working around the removal of this monkey patch for now.

PR #1338, opened 2023-08-08
Handle nested JSON and JSON arrays in params properly

a-alhusaini:
Before:
```cr
class SomeController
  def index
    params.validation { # ... }
    
    params["x"] # => string "[1,2,3]"
end
```

After
```cr
class SomeController
  def index
    params.validation { # ... }
    
    params["x"] # => Array(JSON::Any)
end
```

### DIFF
--- a/spec/amber/validations/params_spec.cr
+++ b/spec/amber/validations/params_spec.cr
@@ -3,6 +3,19 @@ require "../../spec_helper"
 module Amber::Validators
   describe Params do
     describe "#validation" do
+      context "JSON param" do
+        it "parses json array as [] of JSON::Any" do
+          params = json_params_builder({ x: [ 1, 2, 3 ] }.to_json)
+          validator = Validators::Params.new(params)
+
+          validator.validation do
+            required(:x)
+          end
+
+          validator.validate!["x"].should be_a Array(JSON::Any)
+        end
+      end
+
       context "required params" do
         context "when missing" do
           it "is not valid and has 2 errors" do
--- a/spec/amber/validations/params_spec.cr
+++ b/spec/amber/validations/params_spec.cr
@@ -5,14 +5,16 @@ module Amber::Validators
     describe "#validation" do
       context "JSON param" do
         it "parses json array as [] of JSON::Any" do
-          params = json_params_builder({ x: [ 1, 2, 3 ] }.to_json)
+          params = json_params_builder({ x: [ 1, 2, 3 ], y: {z: [ 1, 2 ] } }.to_json)
           validator = Validators::Params.new(params)
 
           validator.validation do
             required(:x)
+            required("y")
           end
 
           validator.validate!["x"].should be_a Array(JSON::Any)
+          validator.validate!["y"].should be_a JSON::Any
         end
       end
 
--- a/spec/support/helpers/validations_helper.cr
+++ b/spec/support/helpers/validations_helper.cr
@@ -1,4 +1,9 @@
 module ValidationsHelper
+  def json_params_builder(body = "")
+    request = HTTP::Request.new("POST", "/", HTTP::Headers{"Content-Type" => "application/json"}, body)
+    params = Amber::Router::Params.new(request)
+  end 
+
   def params_builder(body = "")
     request = HTTP::Request.new("GET", "")
     params = Amber::Router::Params.new(request)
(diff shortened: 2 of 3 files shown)


src/amber/validators/params.cr
```diff
@@ -6,7 +6,7 @@ module Amber::Validators
   class BaseRule
     getter predicate : (String -> Bool)
     getter field : String
-    getter value : String?
+    getter value : String | Array(JSON::Any) | Nil
```
crimson-knight, line 9: Let's add in `JSON::Any` to this as well

```suggestion
    getter value : String | Array(JSON::Any) | JSON::Any | Nil
```

src/amber/validators/params.cr
```diff
@@ -82,10 +87,11 @@ module Amber::Validators
   class Params
     getter raw_params : Amber::Router::Params
     getter rules = [] of BaseRule
-    getter params = {} of String => String?
+    getter params = {} of String => String | Array(JSON::Any) | Nil
```
crimson-knight, line 90:
```suggestion
    getter params = {} of String => String | Array(JSON::Any) | JSON::Any | Nil
```

src/amber/validators/params.cr
```diff
@@ -31,7 +31,12 @@ module Amber::Validators
     end
 
     private def call_predicate(params : Amber::Router::Params)
-      @value = params[@field]
+      begin
```
crimson-knight, line 34: Instead of a `begin`/`rescue` here we can use a `case` statement, pass in the request object or some of the information about the request (specifically the `content-type` value if it's present) so we can determine if we need to parse for JSON or not.

a-alhusaini: That is handled earlier up the stack. The parameters passed into the validator are all the body params and the route and any other form of param concatenated together. I like the idea of using a `case statment`.

a-alhusaini: Actually, we can't use a case statement because JSON.parse raises and there is no other way to handle it.

src/amber/validators/params.cr
```diff
@@ -6,7 +6,7 @@ module Amber::Validators
   class BaseRule
     getter predicate : (String -> Bool)
     getter field : String
-    getter value : String?
+    getter value : String | Array(JSON::Any) | JSON::Any | Nil
```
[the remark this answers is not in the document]
crimson-knight: That was my thought as well, so I don't think we need the extra `Array(JSON::Any)`

spec/support/helpers/validations_helper.cr
```diff
@@ -1,4 +1,9 @@
 module ValidationsHelper
+  def json_params_builder(body = "")
+    request = HTTP::Request.new("POST", "/", HTTP::Headers{"Content-Type" => "application/json"}, body)
```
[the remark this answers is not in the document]
crimson-knight: Well this is just a test, but we should be testing `POST` `PATCH` AND `PUT`

spec/amber/validations/params_spec.cr
```diff
@@ -3,6 +3,21 @@ require "../../spec_helper"
...
+        it "parses json array as [] of JSON::Any" do
+          params = json_params_builder({ x: [ 1, 2, 3 ], y: {z: [ 1, 2 ] } }.to_json)
+          validator = Validators::Params.new(params)
+
+          validator.validation do
+            required(:x)
+            required("y")
+          end
+
+          validator.validate!["x"].should be_a Array(JSON::Any)
```
hugopl, line 16: Maybe returning a `JSON::PullParser` would be better here, so people could use  `JSON::Serializable` to decode these parameters. 🤔

a-alhusaini: When was that introduced in the language? I don't recall seeing this at the time I wrote this PR.

crimson-knight: @a-alhusaini it's been part of the std library for a few years

https://crystal-lang.org/api/1.10.1/JSON/PullParser.html

crimson-knight:
Closing during a backlog sweep — this sat far too long, and that's on us, not you. Where this landed: the approach works (rebased onto current master it passes the full suite, 488/0), but parsing every param with a blanket `JSON.parse` changes the type of plain form values like `"123"` or `"true"` from `String` to `JSON::Any` for every existing app, which is a breaking change for the validator's public getters — that's the piece that kept this from merging as-is, along with gating the parsing on `Content-Type: application/json` (per the 2023 review threads). Nested-param handling is still wanted; if you'd like to pick it back up with content-type-gated parsing we'd be glad to review promptly (the project now compiles on Crystal 1.21 and reviews are being turned around again). Thank you for the contribution and the patience.

PR #1341, opened 2023-10-15
Update Linux Installation instructions

HCLarsen:
Updates and improves the installation instructions for Debian based Linux.


crimson-knight (approved):
Thank you!

PR #1343, opened 2023-10-22
Remove unneeded monkey patches for Number class.

hugopl:
### Description of the Change

Just removes re-implementations of:

- Number#positive?
- Number#negative?
- Number#zero?


crimson-knight (approved):
We should probably just remove all of the monkey patches, or more them into the validation objects/process

PR #1353, opened 2023-10-27
fixes auth template when model entity name is different of User

aissarmurad:
### Description of the Change

This PR fixes a issue with auth generator template when We generate an auth model  using a different name from documentation.

### Benefits

This change allows generate an auth model different of User.

### Possible Drawbacks

None


crimson-knight (approved):
Thank you!

PR #1358, opened 2023-11-14
Feature Adding AssetPipeline & ImportMaps

crimson-knight:
### Description of the Change

This is a big step forward in modern JS for Amber!

This change adds the new `import maps` feature and the `asset_pipeline` shard for new apps by default. This also has some minimal updates to get the existing client_reload feature working and get the 

### Benefits

- No build step during deployments, but allows JS locally and from CDNs (npm, etc.)
- This fingerprints asset files so when files change, the links/references to the files are automatically updated to more easily break any browser caching that is configured.

### Possible Drawbacks

Using module-based JS does require a different approach than what most backend devs are probably comfortable with. If you're accustomed to the standard of just putting plain JS into a file and it executes on it own, this is going to be a bit different.


src/amber/cli/templates/app/config/initializers/import_map.cr.ecr
```diff
@@ -0,0 +1,15 @@
+require "asset_pipeline"
+
+FRONT_LOADER = AssetPipeline::FrontLoader.new(js_source_path: Path["src/javascript"], js_output_path: Path["public"]) do |import_maps|
```
drujensen, line 3: wdyt of `assets/javascript` instead?

src/amber/cli/templates/app/config/initializers/import_map.cr.ecr
```diff
@@ -0,0 +1,15 @@
...
+  if Amber.settings.auto_reload
+    import_map.add_import("client_reload", "/client_reload.js")
+  end
+
+  import_map.add_import("@popperjs/core", "https://cdn.jsdelivr.net/npm/@popperjs/core@2.11.6/dist/umd/popper.min.js")
+  import_map.add_import("bootstrap", "https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/js/bootstrap.esm.min.js")
+  import_map.add_import("amber", "/amber.js")
+
+  import_maps << import_map
+end
```
drujensen, line 15: add a NL

src/amber/cli/templates/app/config/initializers/import_map.cr.ecr
```diff
@@ -0,0 +1,15 @@
...
+
+  if Amber.settings.auto_reload
+    import_map.add_import("client_reload", "/client_reload.js")
+  end
+
+  import_map.add_import("@popperjs/core", "https://cdn.jsdelivr.net/npm/@popperjs/core@2.11.6/dist/umd/popper.min.js")
+  import_map.add_import("bootstrap", "https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/js/bootstrap.esm.min.js")
+  import_map.add_import("amber", "/amber.js")
+
+  import_maps << import_map
```
drujensen, line 14: This is nice!

drujensen (commented):
lgtm. 💯   one suggestion is to use `assets` directory instead of `src` for familiarity by Rails developers.

drujensen (approved):

src/amber/cli/templates/app/config/initializers/import_map.cr.ecr
crimson-knight: I kicked that around. If the deployment doesn't require the entire code base be cloned, then it makes total sense.
But, wouldn't most dockerized/containerized deploys mean the entire code base was cloned into the image and then deployed?

drujensen: I would just copy the `assets` directory to the docker image instead of `src/javascript`.  I think it will actually make creating the image easier.

crimson-knight:
Closing as superseded (own PR, backlog sweep). The durable artifact here shipped independently: `amberframework/asset_pipeline` is alive and current (v0.36.0, 2026-07-10) and this PR's pin (`~> 0.34.0`) is already stale. The remaining surface is entirely the v1 embedded app template, and v1 is in maintenance mode — new-app scaffolding investment now belongs in the amber_cli v2 templates, where drujensen's `assets/` naming and hugopl's `import_map_tag` helper suggestions can be honored properly. Tracking the v2 port there.

PR #1359, opened 2023-11-24
Remove use of YAML.mapping from amber/cli/plugins/settings.cr.

hugopl:
### Description of the Change

Remove use of `YAML.mapping` on amber/cli/plugins/settings.cr

My original objective was to fully remove `YAML.mapping` from Amber, but without time to complete the whole removal I though that submit the initial work would be a good start.


crimson-knight (approved):
LGTM thank you

PR #1375, opened 2026-02-21
Fast Multisocket/Multichannel Redis PubSub Websocket Adapter

jonsilverman50-star:
### Description of the Change

This is a better Redis adapter for Amber Websockets. It supposed unlimited sockets and channels and uses the Redis PubSub backend properly. In the current Redis adapter you can only subscribe to one channel. This needs merge. Also included are updates for cleaning up dead sockets and rescuing sends on dead sockets. That was needed for stability.

Please see my site https://gbalda.org for an example in production.

### Alternate Designs

None.

### Benefits

Speed increases for sites that use multiple sockets.

### Possible Drawbacks

Only works with a process count of 1. (production.yml process_count)
Tests that I wrote long ago need fix. I'll try to fix them.


crimson-knight (changes requested):
Thanks for this — the multi-channel subscription rewrite addresses a real limitation: the existing adapter subscribes to a single channel topic, so anyone running more than one channel has been silently broken. The dead-socket guards, the `OpenSSL::SSL::Error` rescue in `beat`, and the JS try/catch guards are all concrete stability improvements, and the latter two were good enough that V2 has adopted them (with credit to this PR): the SSL rescue is on the V2 branch and the JS client guards shipped in amber_cli v2.0.1.

For this PR to merge into v1, a few things need to land first:

1. **Upstream crystal-redis, not a personal fork.** `shard.yml` currently pins `jonsilverman50-star/crystal-redis` with no version constraint. Please point at upstream `stefanwille/crystal-redis` with a pinned version.
2. **`CHANNEL_TOPIC_PATHS` as a mutable global constant** will misbehave under concurrent connections — move it to instance state on the adapter.
3. **Duplicated subscribe loop** — please consolidate to one path.
4. **sqlite3 removal from shard.yml** is unrelated; if intentional, it belongs in its own PR with a rationale.
5. **Tests** — the PR notes they need fixes; please get the adapter spec passing before re-requesting review.

For context: V2 (now public: https://github.com/amberframework/amber/pull/1383) removes the bundled Redis dependency entirely in favor of a pluggable `PubSubAdapter` interface, so this design doesn't carry forward there. We're keeping v1.x minimally maintained while the website still points at it, so a working version of this fix would still land — but V2 is where the energy is, and your stability instincts would be very welcome there. Apologies for the slow review.

crimson-knight:
Adding one concrete finding to the June 11 review while re-validating the backlog on Crystal 1.21: with this branch merged onto current master, `spec/amber/websockets/channel_spec.cr:6` ("#on_message should call handle_message") fails **deterministically** — in the full run the assertion gets "handle joined …" instead of the test message, and in isolation `handle_message` never fires at all. Master passes this spec. It looks connected to the `ClientSockets.client_sockets.delete(client_socket)` call (the hash is keyed by id, so deleting by socket object is a no-op) and/or the mutable `CHANNEL_TOPIC_PATHS` re-keying. Also note `spec/amber/websockets/adapters/redis_adapter.cr` doesn't end in `_spec.cr`, so `crystal spec` never runs it. The June 11 asks (pinning the redis fork, the shared subscribe loop, sqlite3 removal) still stand — happy to re-review as soon as those land.

PR #1376, opened 2026-05-19
Fix unsafe shell invocation in amber exec

tcoatswo:
## Summary
Replace shell-interpolated process execution in amber exec with argument-array Process.run calls.

## Changes
- replace shell-based editor launch with direct process execution
- replace shell-based file copy with direct process execution
- add regression coverage for shell metacharacters in --editor and copied filenames

## Verification
I could not run the Crystal spec suite in this environment because the host does not have crystal or shards installed. This PR includes regression specs for the affected code paths for maintainers/CI to exercise.

### DIFF
--- a/spec/amber/cli/commands/exec_spec.cr
+++ b/spec/amber/cli/commands/exec_spec.cr
@@ -44,10 +44,30 @@ module Amber::CLI
         File.read(logs.last?.to_s).should eq "[:a, :b]\n"
       end
 
-      it "opens editor and executes .cr file on close" do
-        MainCommand.run(["exec", "-e", "echo 'puts 1000' > "])
-        logs = `ls tmp/*_console_result.log`.strip.split(/\s/).sort
-        File.read(logs.last?.to_s).should eq "1000\n"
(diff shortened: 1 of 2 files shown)


crimson-knight:
Thank you for tracking down both injection vectors and writing regression specs to cover them. Replacing the shell-interpolated `system()` calls with argument-array `Process` invocations is exactly the right fix — shell metacharacters in editor names or filenames are a real attack surface.

Both PRs are valid and complementary (#1376 → master, #1377 → stable backport). Merging both.

You'll also want to know your fix outlived v1: the standalone CLI that replaces the bundled one had independently fixed the file-copy path with `FileUtils.cp`, but the editor-launch injection was still there. It's now hardened in [amber_cli#22](https://github.com/amberframework/amber_cli/pull/22) — using `Process.parse_arguments` so multi-word editors like `code -w` still work — with credit to this PR, and shipped in v2.0.1.

Maintainer attention has been on V2 (now public: https://github.com/amberframework/amber/pull/1383), which is why these sat for a few weeks — sorry about that, and grateful for the security work.

PR #1378, opened 2026-06-10
Fix HTTP Request URI parsing for Crystal 1.20+

renich:
### Description of the Change

Under Crystal 1.20+, absolute URIs passed in `HTTP::Request` resources (such as when testing routes or calling external APIs) are parsed with strict RFC 7230 request-target rules. This leads to `nil` host and port inside the parsed `URI` object.

To resolve this compatibility issue:
* We introduce a private helper method `parsed_uri` in `HTTP::Request` to safely fallback to parsing the resource directly with `URI.parse` if it starts with `http://` or `https://`.
* We simplified the verbose block in `src/amber/controller/helpers/responders.cr` using Crystal's idiomatic short block notation.
* We removed deprecated/obsolete specs for the method tunneling monkey patch which was deleted upstream but left orphaned tests behind.

### Alternate Designs

We considered using a custom parser wrapper, but relying on `URI.parse` for absolute URIs matches Crystal's expected stdlib URI behavior without adding overhead or complexity.

### Benefits

* The Amber framework is fully compatible with Crystal 1.20+.
* The entire spec suite compiles and passes successfully (464 examples, 0 failures, 0 errors, 0 pending).

### Possible Drawbacks

None identified. Backward compatibility for public properties was carefully preserved.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ package-lock.json
 node_modules
 /myapp
 cli.dwarf
+.agents/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,61 @@
+# Agent Guidelines & Project Cooperation Protocol
+
+This document outlines the protocols, coding conventions, testing rules, and commit message standards for AI agents cooperating on the Amber Framework.
+
+---
+
+## 1. Project Journaling Protocol (PJP)
+All agent sessions must maintain a concurrency-safe, LLM-efficient project memory using the `ajourn` tool.
+* **Startup**: Run `ajourn startup` as the very first step of any session.
+* **Persistence**: Log "Knowledge Deltas" using `ajourn log -m "..." -t "..."`.
+* **Standard Tags/Vocabulary**:
+  * `DEC` (Decision): Architectural, design, or procedural decisions.
+  * `RAT` (Rationale): Explanations for specific code patterns, fixes, or designs.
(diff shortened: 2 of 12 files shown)


crimson-knight:
Thanks for tracking down the Crystal 1.20+ URI parsing regression and putting together a fix. The nil host/port crash on absolute-URI resources is a real build-breaker for anyone running v1 on a current Crystal release, and the orphaned spec cleanup is solid housekeeping on top of it.

I'm going to merge this into v1. One note for transparency: V2 had an inline absolute-URI guard covering `port`, but your `parsed_uri` helper is more complete — it covers `url` as well — so I've ported your pattern to the V2 branch with credit to you in the commit.

On `AGENTS.md`: I'm genuinely glad to see it — we're heading the same direction. The question I want to align on is intent: is this file meant for *developing amber itself*, or meant to *ship with the shard* so a consumer's coding agent knows how to use the library? Those are different artifacts, and we don't have a solid convention yet for separating them — we're building package-manager tooling specifically to distribute consumer-facing agent resources (AGENTS.md, skills, docs) with a library on install. Which did you intend here? I'll merge the URI fix either way; depending on your answer we'll find the right home for the file rather than just dropping it.

If you're interested in where Amber is heading, V2 is now public: https://github.com/amberframework/amber/pull/1383. The standalone CLI is at https://github.com/amberframework/amber_cli. Contributions to either are very welcome.

PR #1380, opened 2026-06-10
Correctly parse and match HTTP Accept header parameters

renich:
### Description of the Change

Fixes a bug where semicolon-delimited options/parameters in individual Accept header types (such as quality values `q=...` or charsets like `charset=utf-8`) cause parsing failures. 

Previously, the Accept header parser split the entire header value on `;` first, taking only the first segment. This resulted in:
1. Ignoring all media types listed after the first semicolon.
2. Inability to match media types if the client requested them with parameters (like `application/json; charset=utf-8` from Dart/Flutter clients), returning the fallback type instead (usually HTML) even when JSON was available.

To address this:
* We split the Accept header on commas first to isolate individual media types.
* We then split each media type on semicolons to remove quality/charset parameters and extract the raw MIME type (e.g. `application/json`).
* Added a spec case to verify that complex Accept headers with parameters match correctly.

### Alternate Designs

None. This follows standard RFC 7231 content negotiation rules.

### Benefits

* Semicolons in media type list options no longer drop subsequent media types.
* Clients sending `application/json; charset=utf-8` are correctly matched to the controller's `json` block rather than receiving the default format (e.g., HTML).

### Possible Drawbacks

None.

---

### DIFF
--- a/spec/amber/controller/respond_with_spec.cr
+++ b/spec/amber/controller/respond_with_spec.cr
@@ -152,6 +152,16 @@ module Amber::Controller
           context.response.headers["Content-Type"].should eq "application/json; charset=utf-8"
           context.response.status_code.should eq 403
         end
+
+        it "responds with json for complex Accept header with parameters" do
+          context.response.status_code = 200
+          expected_result = %({"type":"json","name":"Amberator"})
+          context.request.headers["Accept"] = "application/json; charset=utf-8, text/javascript, */*; q=0.0"
+          context.request.path = "/response/1"
+          ResponsesController.new(context).index.should eq expected_result
+          context.response.headers["Content-Type"].should eq "application/json; charset=utf-8"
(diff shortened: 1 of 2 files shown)


crimson-knight:
Thanks for this fix. You correctly identified a real parsing bug — splitting on `;` before `,` meant any media type after the first quality parameter was silently dropped, and clients sending `application/json; charset=utf-8` would never match. The regression spec documents the expected behavior clearly, and the change follows RFC 7231.

Merging this to v1. For the record: the identical bug existed on the V2 branch in two places (`responders.cr` and `support/mime_types.cr`); both are now fixed there with credit to you in the commit. This kind of careful RFC-compliance work is exactly what matters in the new codebase too — V2 is public at https://github.com/amberframework/amber/pull/1383 and you'd be very welcome there.

Issue #1361, opened 2024-04-17
Should Generate Controller Specs with `amber generate api`


HCLarsen:
When generating an API only resource, it generates the files for the model and the controller, but in the `spec` folder, it only generates specs for the model, not the controller.

PR #1381, opened 2026-06-10
Generate controller specs for api and scaffold resources

renich:
### Description of the Change

When generating an `api` or a `scaffold` resource via the CLI, it generates the model, migration, views, and controller files, but in the `spec/` folder, it only generated specs for the model. No spec files were generated for the controller (except an empty `spec_helper.cr`).

This Pull Request resolves this gap (reported in **Issue #1361** for api resources) by adding controller spec templates for both the `api` and `scaffold` controller generators:
* Created `{{name}}_controller_spec.cr.ecr` under `src/amber/cli/templates/api/controller/spec/controllers/`.
* Created `{{name}}_controller_spec.cr.ecr` under `src/amber/cli/templates/scaffold/controller/spec/controllers/`.
* Restored and enabled the generator controller spec existence assertions in `spec/amber/cli/commands/generator_spec.cr` that were previously commented out.
* Added a new test `generates api correctly` in `spec/amber/cli/commands/generator_spec.cr` to verify that `amber generate api` renders the model, migration, controller, and corresponding specs, and updates routes.

### Alternate Designs

None. Restores standard skeleton spec generation for controllers.

### Benefits

* Generated API and Scaffold controllers are now covered by skeleton tests out of the box.
* Resolves **Issue #1361**.

### Possible Drawbacks

None.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ package-lock.json
 node_modules
 /myapp
 cli.dwarf
+.agents/
--- a/spec/amber/cli/commands/generator_spec.cr
+++ b/spec/amber/cli/commands/generator_spec.cr
@@ -76,9 +76,8 @@ module Amber::CLI
 
         File.exists?("./spec/models/#{snake_case}_spec.cr").should be_true
         File.exists?("./src/models/#{snake_case}.cr").should be_true
-        
-        # Recently removed the controller specs, once controller testing is more mature, we'll add them back
-        #File.exists?("./spec/controllers/#{snake_case}_controller_spec.cr").should be_true
+
+        File.exists?("./spec/controllers/#{snake_case}_controller_spec.cr").should be_true
         File.exists?("./src/controllers/#{snake_case}_controller.cr").should be_true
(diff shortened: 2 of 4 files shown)


crimson-knight:
Thanks for putting this together, @renich — controller spec skeletons for `api` and `scaffold` generators fill a real gap in the v1 CLI, and un-commenting the disabled scaffold assertions is good test hygiene.

I'm going to close this rather than merge it, and I want to be direct about why: the bundled CLI is removed in V2, and its replacement, [amberframework/amber_cli](https://github.com/amberframework/amber_cli), already ships controller spec templates for both generators — and they go further than skeletons: they use `RequestHelper` to perform real HTTP calls with status-code and body assertions across index/show/create/update/destroy (plus redirect `Location` checks on the scaffold side). The problem you identified is already solved at higher fidelity on the path getting active work.

One small note: the `.gitignore` addition of `.agents/` looks like a local tooling artifact and shouldn't land upstream regardless.

If you're open to it, amber_cli is a great place to direct this energy. The V2 framework itself is public now too: https://github.com/amberframework/amber/pull/1383. Thanks again for the time you put into this.

PR #1385, opened 2026-06-15
security: escape exception messages in default HTML error response

renich:
Escapes raw HTML exception messages in the default error controller to prevent reflected XSS.

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -1,0 +1,4 @@
+## 2026-06-11 - [XSS in Default Error Handler]
+**Vulnerability:** Reflected Cross-Site Scripting (XSS) in default HTML error response.
+**Learning:** The default error controller generated raw HTML incorporating the unescaped Exception message. In Crystal, `Exception#message` can be `nil` (`String?`), so safely interpolating it into HTML requires `HTML.escape(@ex.message || "")` and explicitly requiring the `"html"` module.
+**Prevention:** Always escape user-controllable or dynamic string data, including exception messages, before interpolating it into an HTML template or string.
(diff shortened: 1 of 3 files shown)


crimson-knight:
This is a legitimate fix and **we'll accept it** — thank you. ✅

I confirmed the vulnerability is real: `src/amber/controller/error.cr` interpolates `@ex.message` directly into the HTML 500 page, and `Amber::Exceptions::RouteNotFound` embeds the raw `request.path` into that message, so a request to `/<script>...</script>` reflects unescaped into the response. It fires in non-development environments serving `Accept: text/html` (development renders the `exception_page` shard instead), so it's a real production reflected-XSS. `HTML.escape(@ex.message || "")` is the correct, complete fix for that response, and the `nil` guard is right since `Exception#message : String?`.

One thing needed before merge:

- **Add a spec** asserting the body is escaped. The existing `spec/amber/pipes/error_spec.cr` only checks status codes, so nothing guards against regression. Please add a case that drives an exception whose message contains `<script>` (e.g. via `RouteNotFound` with a `<script>` path) and asserts the response body contains the escaped entities (`&lt;script&gt;`) and **not** the raw tag.

No need to remove `.jules/sentinel.md` — keeping the agent notes is fine with us. Once the spec is in, this is good to merge.

renich:

Following your suggestion, I have added a regression spec in `spec/amber/pipes/error_spec.cr` that sets the `Accept: text/html` header and drives an exception whose message contains HTML characters (e.g. `<script>alert('xss')</script>`), verifying that the response status is 500 and the returned HTML body contains the escaped entities (`&lt;script&gt;alert(&#39;xss&#39;)&lt;/script&gt;`) instead of raw unescaped tags.

All specs compile and pass.

PR #1387, opened 2026-06-15
perf: optimize socket route checking to avoid array allocation

renich:
Replaces array mapping with a direct .any? check to avoid allocating a new array on every incoming HTTP request.


crimson-knight (approved):
Reviewed and verified — approving. ✅

`socket_route_defined?` runs on every request (via `Request#valid_route?`), and the old `@socket_routes.map(&.[:path]).includes?(request.path)` allocated a temporary `Array(String)` on each call before scanning it. Switching to `@socket_routes.any? { |route| route[:path] == request.path }` removes that allocation and short-circuits.

Confirmed behavior is identical (differential check over exact match, trailing slash, case variant, empty string, missing path, prefix, and empty collection — zero divergence) and the full router spec suite passes (153 examples, 0 failures) on Crystal 1.20.0.

Clean, correct, pure upside. Merging.

PR #1388, opened 2026-06-15
security: prevent command injection in encrypt CLI command

renich:
Quotes file arguments passed to system calls in encrypt.cr when invoking the user's editor.

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -1,0 +1,4 @@
+## 2024-06-13 - Command Injection in CLI Utilities
+**Vulnerability:** Command injection in `src/amber/cli/commands/encrypt.cr` via string interpolation in the `system(...)` command (`system("#{options.editor} #{unencrypted_file}")`). If `env` or `editor` contains shell special characters, arbitrary commands could be executed.
+**Learning:** Passing a single string to `system(...)` causes it to be evaluated by a shell, allowing shell meta-characters to alter execution. User-controlled variables interpolated into this string create command injection vulnerabilities.
+**Prevention:** Avoid string interpolation when calling system commands. Use the array form of `system(executable, args_array)` or `Process.run(executable, args_array)` to bypass the shell entirely and pass arguments safely.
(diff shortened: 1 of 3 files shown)


crimson-knight:
Thanks for catching the shell-injection pattern here — the underlying issue is real, but the fix as written is **incomplete**, so requesting changes.

What's real: on `master`, `src/amber/cli/commands/encrypt.cr` does `system("#{options.editor} #{unencrypted_file}")`. A single-string `system(...)` runs via `/bin/sh -c`, so interpolated metacharacters are interpreted. (Severity is low in practice — this is a local developer CLI where the input and the operator are the same person — but worth fixing cleanly.)

Why the current fix doesn't close it: wrapping only the filename in `Process.quote(...)` leaves `options.editor` (and `$EDITOR`) raw and still passes a single string to `system`, so a shell is still spawned. Verified: with `editor = "vim; touch /tmp/PWNED"`, the injection still fires — only the filename arg was hardened.

Requested change — match the pattern we already settled on in `exec.cr` (PR #1376), which bypasses the shell entirely with the args-array form:

```crystal
Process.run(options.editor, [unencrypted_file], output: Process::Redirect::Inherit, error: Process::Redirect::Inherit) unless options.noedit?
```

(Note: `Process.quote`-ing the editor instead would break legitimate multi-word editors like `code --wait`; the array form is the right call. Your own `.jules/sentinel.md` actually prescribes the array form — the code just didn't implement it.)

Please also **add an adversarial spec** mirroring `spec/amber/cli/commands/exec_spec.cr` ("does not execute shell metacharacters in the editor option" / "...in filenames"). Once it's the array form + a spec, happy to merge.

renich:

Following your suggestions:
1. I replaced the shell-spawning `system(...)` call with `Process.run(options.editor, [unencrypted_file], ...)` using the safe array-argument form.
2. I added a test case in `spec/amber/cli/commands/encrypt_spec.cr` verifying that command injection in the editor option is prevented.
3. I also updated the rescue block in `src/amber/cli/commands/encrypt.cr` to propagate the exception during spec runs (when `AMBER_ENV` or `CRYSTAL_CLI_ENV` is set to `"test"`) instead of aborting the process via `exit!`, allowing the tests to run and verify the behavior in-process.

All encrypt specs compile and pass.

